PR #58, opened 2026-01-24
Add mandatory frontmatter to eval-harness SKILL.md

Nikola-Milovic:
Got a warning about this when adding this skill via https://github.com/frmlabz/omnidev so just wanted to quickly fix it. 

From [claude docs](https://code.claude.com/docs/en/skills)

> Every skill needs a SKILL.md file with two parts: YAML frontmatter (between --- markers) that tells Claude when to use the skill, and markdown content with instructions Claude follows when the skill is invoked. The name field becomes the /slash-command, and the description helps Claude decide when to load it automatically.

So the `frontmatter` part is mandatory

## Summary by CodeRabbit

* **Documentation**
  * Improved skill documentation with added metadata, including skill identification and comprehensive description for better organization and clarity.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


affaan-m:
Fixed directly in b9b7831 - added frontmatter to eval-harness

PR #242, opened 2026-02-17
fix: use CLI argument for hook phase detection in observe.sh

t-s-li:
## Summary

- `observe.sh` receives `"pre"` or `"post"` as `$1` from the hook config, but the Python code was reading `hook_type` from the stdin JSON — a field Claude Code **does not include**
- This caused `hook_type` to always default to `"unknown"`, so `"Pre" in hook_type` was always `False`
- **Result**: all PreToolUse observations were silently recorded as `tool_complete` instead of `tool_start`, making it impossible to distinguish pre vs post events

## Fix

- Capture `$1` as `HOOK_PHASE` and pass it to the Python parser via environment variable
- Remove the unused `data.get("hook_type")` lookup
- Applied to both `skills/` and `.cursor/` copies
- Also fixed `TIMESTAMP` export in `.cursor` copy where inline `VAR=val cmd` syntax didn't propagate to the python subprocess

## Test plan

- [x] `echo '{"tool_name":"Read","tool_input":{}}' | observe.sh pre` → records `"event": "tool_start"`
- [x] `echo '{"tool_name":"Read","tool_input":{}}' | observe.sh post` → records `"event": "tool_complete"`
- [x] Without argument defaults to `post` (backward compatible)

## Summary by CodeRabbit

* **Bug Fixes**
  * Observation events now use an explicit hook phase for more reliable event tagging.

* **Documentation**
  * Removed a large collection of internal guidance, agent, command, rule, and skill documents to slim repo docs.

* **Chores**
  * Cleaned up continuous-learning tooling: removed obsolete observer/CLI/configuration and auxiliary background helper artifacts.


affaan-m (changes requested):
Automated review: this PR has merge conflicts. Please rebase or resolve.

affaan-m (changes requested):
Automated review: this PR has merge conflicts. Please rebase or resolve.

t-s-li:
Rebased onto latest `main` — merge conflicts are resolved.

PR #310, opened 2026-03-01
fix: pass clawDir argument to handleSessions

ashsolei:
## Summary

`handleSessions()` in `scripts/claw.js` is called without the `dir` argument that its signature expects. It works accidentally because `listSessions(dir)` falls back to `getClawDir()` when `dir` is `undefined`, but this is inconsistent with `handleClear(sessionPath)` and `handleHistory(sessionPath)` which both receive their arguments explicitly.

## Fix

Pass the already-resolved `clawDir` variable to `handleSessions(clawDir)`, matching the pattern used by the other REPL command handlers.

## Testing

All 992 tests pass.

---
## Summary by cubic
Make REPL session handling consistent by passing clawDir to handleSessions and fix the doc-file-warning hook regex so it correctly flags non-standard docs.

- **Bug Fixes**
  - scripts/claw.js: Call handleSessions(clawDir) to match other command handlers.
  - hooks/hooks.json: Simplify path regex to forward slashes to avoid invalid escaping and ensure proper detection.

<sup>Written for commit 62e9ea55d3d441d6c7072d75f3ec98084f2c6ab9. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined documentation file detection path handling for improved cross-platform compatibility.

* **Chores**
  * Enhanced sessions command directory handling.

### DIFF
--- a/scripts/claw.js
+++ b/scripts/claw.js
@@ -218,7 +218,7 @@ function main() {
       }
 
       if (line === '/sessions') {
-        handleSessions();
+        handleSessions(clawDir);
         prompt();
         return;
       }


ashsolei:
Closing — will maintain fixes in a private mirror instead. Thank you for the review.

PR #317, opened 2026-03-02
fix(session-end): always update session summary content

### LINKED ISSUE #187: Session memory doesn't persist or load context across sessions -- config issue or intended?
Description

  I've been trying to get cross-session memory working via the session hooks (session-start.js, session-end.js), but after investigation it seems like no session context is ever actually injected into
  Claude's conversation. The hooks run successfully, but I'm not sure if I'm missing a configuration step or if this is the intended design (manual loading via /sessions load).

  The /sessions command infrastructure and session file management library are well built -- the question is specifically about the automatic save/load hooks.

  Environment

  - everything-claude-code v1.2.0 (installed as plugin)
  - Claude Code CLI, Linux

  What I'm seeing

  SessionEnd saves empty templates

  session-end.js creates .tmp files but they contain only the blank scaffold:

  ## Current State
  [Session context goes here]

  ### Completed
  - [ ]

  ### In Progress
  - [ ]

  ### Notes for Next Session
  -

  Every .tmp file in ~/.claude/sessions/ is ~258 bytes of placeholder text. No actual session data is captured.

  SessionStart doesn't inject content into Claude's context

  session-start.js finds recent session files but only logs their existence:

  [SessionStart] Found 3 recent session(s)
  [SessionStart] Latest: /path/to/session.tmp

  It never reads and outputs the file content. Claude starts every session with zero knowledge of previous sessions.

  Questions

  1. Am I missing a config step? Is there something beyond installing the plugin that enables session content to be saved/loaded automatically?
  2. Is this by design? Is the intended workflow to manually run /sessions load <id> each time? If so, it might help to document that the hooks are informational only and automatic memory requires manual
  loading.
  3. Is CLAUDE_TRANSCRIPT_PATH available during SessionEnd? The evaluate-session.js hook references it, but no .md summaries are being generated, which suggests it may not be set for SessionEnd hooks.

  Suggestion (if this is a gap)

  1. SessionEnd: Parse CLAUDE_TRANSCRIPT_PATH to write a meaningful summary into the session file instead of a blank template.
  2. SessionStart: Read the latest session file and output its content so Claude receives it as context.

  Thanks for the great plugin -- just trying to figure out if I'm holding it wrong!

cnwillz:
```markdown
## Problem
Previously, `session-end.js` would only write content to session files on first creation. Subsequent sessions would only update the timestamp, causing stale content (e.g., old tasks, resolved issues) to persist indefinitely.

## Root Cause
The original code had a check that only replaced content if the file contained the blank template marker `[Session context goes here]`. This meant:
- First session → creates file with content ✅
- Subsequent sessions → only updates timestamp, keeps old content ❌

## Solution
Remove the blank-template-only check and always update the Session Summary section with fresh content from the current session transcript.

## Changes
- Remove the `[Session context goes here]` check
- Replace entire `## Session Summary` section on every session end
- Keep timestamp update separate from content update

## Related
- Fixes #187 (addresses the stale content issue)
- Complements commit 87d19f9 which fixed blank template writing
```

---
## Summary by cubic
Always replace the session summary on session end, matching both "## Session Summary" and "## Current State" and replacing to the end of the file to avoid duplicate "### Stats". Keeps summaries fresh across sessions and addresses #187.

<sup>Written for commit ad3bd32a82d91f937e3fb89f89ed6e96f9a3a905. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable session summary updates: the system now recognizes both "Session Summary" and "Current State" headers, replaces the entire existing summary section to prevent duplicates, and updates the file whenever a summary is present—resulting in cleaner, consistent session files.

### DIFF
--- a/scripts/hooks/session-end.js
+++ b/scripts/hooks/session-end.js
@@ -171,9 +171,10 @@ async function main() {
     if (summary) {
       const existing = readFile(sessionFile);
       if (existing) {
-        // Use a flexible regex that tolerates CRLF, extra whitespace, and minor template variations
+        // Use a flexible regex that matches both "## Session Summary" and "## Current State"
+        // Match to end-of-string to avoid duplicate ### Stats sections
         const updatedContent = existing.replace(
-          /## Session Summary[\s\S]*?(?=### Stats|$)/,
+          /## (?:Session Summary|Current State)[\s\S]*?$/ ,
           buildSummarySection(summary).trim() + '\n'
         );
         writeFile(sessionFile, updatedContent);


cnwillz:
@cubic-dev-ai The regex issue has been fixed in commit ad3bd32. Changes made:

1. Updated regex to match both `## Session Summary` and `## Current State` headers
2. Changed from lookahead `(?=### Stats|$)` to end-of-string `$` to prevent duplicate `### Stats` sections

New regex: `/## (?:Session Summary|Current State)[\s\S]*?$/`

Please re-run your review.

Issue #347, opened 2026-03-07
Security hardening: secret scrubbing, credential stripping, and input validation in hooks


jtzingsheim1:
Hi! I've been doing a thorough review of ECC and identified a few areas where hook scripts could be hardened. None of these are exploitable vulnerabilities — they're defense-in-depth improvements.

### MEDIUM

1. **Observation logs store tool I/O in plaintext** (`skills/continuous-learning-v2/hooks/observe.sh`) — API keys, tokens, and passwords passed through tool calls persist unscrubbed in `observations.jsonl`. Proposed fix: regex-scrub common secret patterns before writing, plus auto-purge observation files older than 30 days.

2. **Git remote URLs with embedded auth tokens** (`skills/continuous-learning-v2/scripts/detect-project.sh`) — URLs like `https://ghp_xxxx@github.com/...` are saved as-is to `projects.json`. Proposed fix: strip credentials with `sed -E 's|://[^@]+@|://|'` before persisting.

### LOW

3. **`runCommand` has no input validation** (`scripts/lib/utils.js`) — Exported function uses `execSync` with no prefix allowlist. Currently safe since all callers use hardcoded strings, but fragile if new callers are added. Proposed fix: allowlist for `git`, `node`, `npx`, `which`, `where` prefixes.

4. **Session ID used unsanitized in temp file paths** (`scripts/hooks/suggest-compact.js`) — `CLAUDE_SESSION_ID` is used directly in file paths without stripping special characters. Proposed fix: `.replace(/[^a-zA-Z0-9_-]/g, '')`.

### Already fixed in v1.8.0

5. **`doc-file-warning` had no stdin size limit** — Now extracted to `scripts/hooks/doc-file-warning.js` with `MAX_STDIN = 1024 * 1024`. No action needed.

## PR

I have a branch with patches for all 4 open items. Will link the PR shortly.

All patches are conservative — they constrain without changing functionality. Existing tests pass.

jtzingsheim1:
This was resolved by PR #348 (merged into main). All four findings from the issue are addressed — secret scrubbing, credential stripping, `runCommand` allowlist, and session ID sanitization.

Closing this issue. Apologies for not using the `Closes #347` keyword in the PR!

PR #348, opened 2026-03-07
fix(hooks): scrub secrets and harden hook security

jtzingsheim1:
## Summary

Addresses #347 — defense-in-depth security hardening across 4 hook/library files.

- **Scrub secrets from observation logs** — regex-replace common secret patterns (api_key, token, password, etc.) with `[REDACTED]` before writing to JSONL, plus auto-purge observation files older than 30 days (`observe.sh`)
- **Strip auth tokens from git remote URLs** — remove embedded credentials from URLs before persisting to `projects.json` (`detect-project.sh`)
- **Add command prefix allowlist to `runCommand`** — only `git`, `node`, `npx`, `which`, `where` are permitted (`utils.js`)
- **Sanitize session ID in temp file paths** — strip non-alphanumeric characters from `CLAUDE_SESSION_ID` to prevent path traversal (`suggest-compact.js`)

All patches are conservative — they constrain without changing existing functionality.

## Test plan

- [x] `node tests/lib/utils.test.js` — 140/141 pass (1 pre-existing Windows symlink failure, unrelated)
- [x] `node tests/hooks/suggest-compact.test.js` — 20/20 pass
- [ ] Manual: work with a mock API key in conversation, verify it's scrubbed in observation logs
- [ ] Manual: clone a repo with token-in-URL remote, verify token stripped in `projects.json`
- [ ] Manual: verify suggest-compact still works with sanitized session IDs

---
## Summary by cubic
Hardened hook security and improved secret scrubbing to prevent credential leaks; now blocks $() and backticks even inside quotes, and prevents newline injection. Addresses Linear #347 with defense-in-depth updates across hooks and utils.

- **Bug Fixes**
  - Strengthened secret scrubbing in observation logs: preserves Authorization scheme; scrubs parse_error fallback; fixes regex quoting; project‑scoped 30‑day auto‑purge of archived files (once per session).
  - Removed embedded credentials from git remotes before saving to projects.json; migrate legacy project directories if the hash changes.
  - Tightened runCommand: allowlist prefixes (git, node, npx, which, where); block $ and backticks anywhere; block ;, |, &, and newline when unquoted; suppress blocked command echo.
  - Sanitized CLAUDE_SESSION_ID in temp paths with a safe charset and a fallback to "default" if empty.

<sup>Written for commit 1cab6a7b9c2ca52ce7ef05f7c27200670bdbcd24. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Security**
  * Blocked unsafe command execution via an allowlist and metacharacter checks; added secret scrubbing for persisted observations.
* **Bug Fixes**
  * Sanitized session identifiers for reliable file handling.
  * Backward-compatible handling of legacy project identifiers to avoid duplicate or missing project data.
* **Maintenance**
  * Automatic purge of observation files older than 30 days; improved observation persistence flow.
* **Tests**
  * Added tests covering command-safety and secret-masking behavior.

### DIFF
--- a/scripts/lib/utils.js
+++ b/scripts/lib/utils.js
@@ -345,11 +345,11 @@ function runCommand(cmd, options = {}) {
     return { success: false, output: 'runCommand blocked: unrecognized command prefix' };
   }
 
-  // Reject shell command-chaining operators outside of quoted strings.
-  // Note: quote-stripping doesn't handle escaped quotes (e.g., "foo\"bar"),
-  // accepted trade-off given trusted internal callers + allowlist above.
+  // Reject shell metacharacters. $() and backticks are evaluated inside
+  // double quotes, so block $ and ` anywhere in cmd. Other operators
+  // (;|&) are literal inside quotes, so only check unquoted portions.
   const unquoted = cmd.replace(/"[^"]*"/g, '').replace(/'[^']*'/g, '');
-  if (/[;|&`$\n]/.test(unquoted)) {
+  if (/[;|&\n]/.test(unquoted) || /[`$]/.test(cmd)) {
     return { success: false, output: 'runCommand blocked: shell metacharacters not allowed' };
   }
 
--- a/tests/lib/utils.test.js
+++ b/tests/lib/utils.test.js
@@ -1070,6 +1070,19 @@ function runTests() {
     assert.ok(result.output.includes('metacharacters not allowed'), 'Should block newline injection');
   })) passed++; else failed++;
 
+  if (test('runCommand blocks $() inside double quotes (shell still evaluates)', () => {
+    // $() inside double quotes is still evaluated by the shell, so block $ everywhere
+    const result = utils.runCommand('node -e "$(whoami)"');
+    assert.strictEqual(result.success, false);
+    assert.ok(result.output.includes('metacharacters not allowed'), 'Should block $ inside quotes');
+  })) passed++; else failed++;
+
+  if (test('runCommand blocks backtick inside double quotes (shell still evaluates)', () => {
+    const result = utils.runCommand('node -e "`whoami`"');
+    assert.strictEqual(result.success, false);
+    assert.ok(result.output.includes('metacharacters not allowed'), 'Should block backtick inside quotes');
+  })) passed++; else failed++;
+
   if (test('runCommand error message does not leak command string', () => {
     const secret = 'rm secret_password_123';
     const result = utils.runCommand(secret);


jtzingsheim1:
Thanks for the thorough review! All 6 findings were valid — addressed in 953babd:

1. **runCommand metacharacter injection** — now rejects `;|&`` `` outside quoted strings (strips quoted sections before checking)
2. **Python regex SyntaxError** — switched from double-quoted to single-quoted shell string with proper escaping
3. **Purge matching live observations.jsonl** — scoped to `${PROJECT_DIR}` and changed glob to `observations-*.jsonl` (archives only)
4. **Empty session ID after sanitization** — added `|| 'default'` fallback after `.replace()`
5. **Bearer token not matched** — added optional auth scheme `(?:[A-Za-z]+\s+)?` before token match
6. **Credential stripping orphans existing data** — computes legacy hash first, migrates existing directory if found

jtzingsheim1:
Round 2 feedback addressed in a187e72:

1. **`$()` and newline bypass** — added `$` and `\n` to blocked metacharacters (stripped quoted sections still apply)
2. **Bearer scheme dropped in redaction** — made auth scheme group capturing, used lambda replacement to preserve "Bearer"/"Basic" in output
3. **Security tests** — added 10 unit tests covering prefix allowlist blocking (`rm`, `curl`, `bash`), metacharacter rejection (`;|&`$` chaining), quoted metacharacter passthrough, and error message leak prevention. 150/151 pass (1 pre-existing Windows symlink EPERM).

jtzingsheim1:
Round 3 feedback addressed in 597d431:

1. **Parse-error fallback now scrubs secrets** — the fallback path in `observe.sh` (lines 134-153) was writing raw unsanitized input to the observations file. Now applies the same `_SECRET_RE` scrubber as the main path before persisting. (Identified by CodeRabbit)
2. **Removed redundant `re.IGNORECASE`** — the `(?i)` inline flag already applies case-insensitive matching to the entire pattern; the `re.IGNORECASE` argument was redundant.
3. **Added inline comment documenting quote-stripping limitation** — the `"[^"]*"` regex doesn't handle escaped quotes; documented as accepted trade-off given trusted internal callers + allowlist. (Requested by CodeRabbit)
4. **Fixed misleading test name** — "falls back to err.message when no stderr" renamed to "returns error output on failed command" since `git` does produce stderr for unknown subcommands.
5. **Added 5 new security tests** — single-quote metacharacter passthrough, mixed quoted+unquoted metacharacters, prefix boundary without trailing space (`gitconfig` blocked), `npx` prefix acceptance, and newline command injection.
6. **Improved existing test** — quoted-metacharacter test now uses `"console.log(1);process.exit(0)"` to actually include a semicolon inside quotes, exercising the quote-stripping logic.

All 155 tests pass (1 pre-existing Windows symlink EPERM, unrelated).

jtzingsheim1:
## Round 5 — Quoted command substitution bypass

**Addressed:** CodeRabbit (critical, duplicate) + Cubic (P1) — `$()` and backticks inside double quotes bypass the metacharacter check.

### Root cause
Our previous fix (commit `a187e72`) added `$` to the blocked characters, but only checked the `unquoted` portion of the command string. In shell semantics, `$()` and backticks are still evaluated inside double quotes (unlike `;|&` which are literal). So `node -e "$(whoami)"` would pass the check but the shell would still execute the substitution.

### Fix (commit `1cab6a7`)
Split the metacharacter check into two parts:
```js
if (/[;|&\n]/.test(unquoted) || /[`$]/.test(cmd)) {
```
- `;|&\n` → checked only in `unquoted` (safe inside double quotes)
- `` $` `` → checked in full `cmd` (unsafe even inside double quotes)

This implements the exact patch suggested by CodeRabbit.

### Tests added
- `runCommand blocks $() inside double quotes (shell still evaluates)`
- `runCommand blocks backtick inside double quotes (shell still evaluates)`

All 24 runCommand security tests pass.

### Out of scope items (noted for future PRs)
- **Extract `_SECRET_RE` to shared module** (CodeRabbit nitpick) — code organization, not security
- **Add other package managers to allowlist** (CodeRabbit suggestion) — feature enhancement
- **Switch to `spawnSync`** (CodeRabbit/Cubic recurring suggestion) — valid hardening but a larger refactor that changes the function's API contract; the current defense-in-depth (allowlist + comprehensive metacharacter blocking) is sufficient given `runCommand` is only called with trusted hardcoded strings, never user input

kylept:
Really solid defense-in-depth work here — the runCommand allowlist and the quote-stripping metacharacter check are exactly the kind of hardening that should be standard in any skill that executes shell commands.

The secret scrubbing in observation logs is particularly important. Snyk/Max Planck's SKILL-INJECT research found that 36% of skills in the wild contain prompt injection payloads, and a big part of the attack surface is skills that silently exfiltrate data through observation/logging channels. Having the scrubber run on the fallback path too (round 3 fix) closes what would've been a real bypass.

One thing this PR highlights: even well-intentioned skills need this level of scrutiny. The broader ecosystem doesn't have a great answer for "how do I know the skill I'm installing has done this work." [SkillSafe](https://skillsafe.ai) has been tackling the distribution side of this — namespace verification, scan-before-share, and dual-side verification so both the publisher and consumer can independently confirm integrity. But the in-skill hardening you're doing here is the other half of the equation that no registry can replace.

Nice work on the test coverage too — 155 passing with the metacharacter edge cases is reassuring.

PR #379, opened 2026-03-11
fix: read tool_response field in observe.sh

### LINKED ISSUE #377: observe.sh reads wrong field name for tool output (tool_output vs tool_response)
## Bug

The `observe.sh` hook in `skills/continuous-learning-v2/hooks/` reads `tool_output` from the Claude Code hook JSON payload, but Claude Code actually sends the field as `tool_response`.

**File:** `skills/continuous-learning-v2/hooks/observe.sh`, line 94

```python
# Current (broken)
tool_output = data.get("tool_output", data.get("output", ""))

# Fix
tool_output = data.get("tool_response", data.get("tool_output", data.get("output", "")))
```

## Impact

All observations recorded by the hook have empty `output` fields. This means the observer/analysis pipeline has no content to learn from — instincts can never be generated from observations.

## Evidence

Claude Code PostToolUse hook stdin JSON format (from official docs and verified locally):

```json
{
  "session_id": "...",
  "tool_name": "Bash",
  "tool_input": {"command": "echo test"},
  "tool_response": {"stdout": "test", "stderr": "", ...},
  "tool_use_id": "toolu_..."
}
```

The hook looks for `tool_output` which doesn't exist, falls back to `output` which also doesn't exist, and records an empty string.

## Versions affected

- v1.7.0 (confirmed)
- v1.8.0 / main (confirmed by code inspection)

ispaydeu:
Fixes #377

## Summary

- `observe.sh` line 127 reads `tool_output` from Claude Code's PostToolUse hook payload, but Claude Code actually sends the field as `tool_response`
- Since neither `tool_output` nor the fallback `output` exist in the payload, every observation gets an empty string for its output field
- This one-line fix adds `tool_response` as the primary key to check, with backward-compatible fallback to the existing field names

## The Bug

**Current code (line 127):**
```python
tool_output = data.get("tool_output", data.get("output", ""))
```

**Actual Claude Code PostToolUse payload:**
```json
{
  "session_id": "...",
  "tool_name": "Bash",
  "tool_input": {"command": "echo hello"},
  "tool_response": {"stdout": "hello\n", "stderr": "", "exitCode": 0},
  "tool_use_id": "toolu_..."
}
```

The code looks for `tool_output` (doesn't exist), falls back to `output` (doesn't exist), and defaults to `""`.

## The Fix

```python
tool_output = data.get("tool_response", data.get("tool_output", data.get("output", "")))
```

Checks `tool_response` first (the actual field Claude Code sends), then falls back to `tool_output` and `output` for backward compatibility.

## Impact

- **All observations recorded by the hook have empty output fields** — the observer/analysis pipeline has had no tool output content to learn from
- Instinct creation is severely degraded because it can't see what tools returned (successes, errors, file contents)
- Affects v1.7.0+ through current main
- Historical archived observations (63MB+) all have this same issue

## Verification

- All 1024 existing tests pass (`node tests/run-all.js`)
- Manually tested with a simulated Claude Code PostToolUse payload — output is now captured correctly
- Backward compatible — payloads using the old `tool_output` field name still work

## Test Plan

- [x] `node tests/run-all.js` — 1024/1024 passed
- [x] Manual test: simulated PostToolUse payload with `tool_response` field → output captured
- [x] Manual test: simulated payload with old `tool_output` field → backward compatible
- [x] Verified locally in a live Claude Code session — observations now include tool output content

---
## Summary by cubic
Fixes #377: Update `observe.sh` to read the `tool_response` field from Claude Code PostToolUse payloads, with correct fallback to `tool_output`/`output` when `tool_response` is missing or null to prevent empty outputs and remain backward compatible.

<sup>Written for commit 496f29f19d5176a456c2d5eb20cee2bd0231456f. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool results in generated events now prefer the most current response field while falling back to legacy fields when needed, improving accuracy of reported outputs.
* **Tests**
  * Added an automated test to verify correct fallback behavior and to ensure observed outputs are recorded reliably.

### DIFF
--- a/skills/continuous-learning-v2/hooks/observe.sh
+++ b/skills/continuous-learning-v2/hooks/observe.sh
@@ -124,7 +124,9 @@ try:
     # Extract fields - Claude Code hook format
     tool_name = data.get("tool_name", data.get("tool", "unknown"))
     tool_input = data.get("tool_input", data.get("input", {}))
-    tool_output = data.get("tool_response", data.get("tool_output", data.get("output", "")))
+    tool_output = data.get("tool_response")
+    if tool_output is None:
+        tool_output = data.get("tool_output", data.get("output", ""))
     session_id = data.get("session_id", "unknown")
     tool_use_id = data.get("tool_use_id", "")
     cwd = data.get("cwd", "")
--- a/tests/hooks/hooks.test.js
+++ b/tests/hooks/hooks.test.js
@@ -63,6 +63,29 @@ function runScript(scriptPath, input = '', env = {}) {
   });
 }
 
+function runShellScript(scriptPath, args = [], input = '', env = {}, cwd = process.cwd()) {
+  return new Promise((resolve, reject) => {
+    const proc = spawn('bash', [scriptPath, ...args], {
+      cwd,
+      env: { ...process.env, ...env },
+      stdio: ['pipe', 'pipe', 'pipe']
+    });
+
+    let stdout = '';
+    let stderr = '';
+
+    if (input) {
+      proc.stdin.write(input);
+    }
+    proc.stdin.end();
+
+    proc.stdout.on('data', data => stdout += data);
+    proc.stderr.on('data', data => stderr += data);
+    proc.on('close', code => resolve({ code, stdout, stderr }));
+    proc.on('error', reject);
+  });
+}
+
 // Create a temporary test directory
 function createTestDir() {
   const testDir = path.join(os.tmpdir(), `hooks-test-${Date.now()}`);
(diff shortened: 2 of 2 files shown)


ispaydeu:
Before/after from a live session — fix applied mid-conversation:                                         
                                                                                                           
  ![observations-before-after]



image
  The first line shows "output": "" (the old behavior). The very next line — same session, same tool —
  shows the full tool_response captured with stdout, stderr, and metadata. The fix took effect immediately
  on the next hook invocation.

ispaydeu:
While testing this fix locally, I discovered a second bug that compounds this one: `observer-loop.sh` uses `--max-turns 3`, which is too low for the analysis task to complete. The result is **zero instincts are ever generated** — Claude hits the turn limit before it can write any output.

Full details and fix in #386. Both fixes are needed together to fully restore the instinct pipeline.

ispaydeu:
Closing this in favor of the fix already present in upstream main. The current  branch implements a more defensive version of this fix (lines 127-129 in observe.sh):

```python
tool_output = data.get("tool_response")
if tool_output is None:
    tool_output = data.get("tool_output", data.get("output", ""))
```

This handles the case where `tool_response` is explicitly `null` in the payload, which the chained `dict.get` approach in this PR would not catch. The upstream fix is strictly better. No action needed here once the next version is tagged.

PR #551, opened 2026-03-17
fix: session commnad path issue

AryanTejani:
## Summary
  - Fix session commands failing after plugin marketplace install when `CLAUDE_PLUGIN_ROOT` is not set
  - Replace broken `~/.claude` fallback with multi-path plugin root resolver

  ## Problem
  When ECC is installed via the Claude Code plugin marketplace, `/everything-claude-code:sessions` commands fail with:
  Cannot find module '/home/.../.claude/scripts/lib/session-manager'
  The inline scripts fall back to `~/.claude` when `CLAUDE_PLUGIN_ROOT` is not set, but the plugin scripts actually live in
  `~/.claude/plugins/cache/everything-claude-code/everything-claude-code/<version>/`.

  ## Changes
  - Replace the broken `CLAUDE_PLUGIN_ROOT || ~/.claude` fallback in all 6 inline script blocks in `commands/sessions.md` with a self-contained multi-path resolver that searches:
    1. `CLAUDE_PLUGIN_ROOT` env var (if set)
    2. `~/.claude` (manual/local install)
    3. Known plugin paths (`plugins/everything-claude-code`, `plugins/marketplace/everything-claude-code`, etc.)
    4. Plugin cache directories (`plugins/cache/everything-claude-code/**/`)
  - Add `scripts/lib/resolve-plugin-root.js` — shared resolver utility for file-based scripts
  - Apply the same fix to `docs/zh-CN/commands/sessions.md`

  ## Test Plan
  - [x] Verified resolver finds plugin in cache directory without `CLAUDE_PLUGIN_ROOT` set
  - [x] Verified resolver uses env var when `CLAUDE_PLUGIN_ROOT` is set
  - [x] Both `session-manager` and `session-aliases` modules load successfully
  - [x] Existing test suite passes (1183/1186, 3 pre-existing failures in install-apply)

---
## Summary by cubic
Fixes broken session commands after marketplace install by replacing the `~/.claude` fallback with a multi-path plugin root resolver. `/everything-claude-code:sessions` now works even when `CLAUDE_PLUGIN_ROOT` is not set.

- **Bug Fixes**
  - Added `scripts/lib/resolve-plugin-root.js` to resolve the plugin root via `CLAUDE_PLUGIN_ROOT`, local installs, known marketplace paths, and cache directories; falls back to `~/.claude`.
  - Replaced path logic in all session command inline scripts in `commands/sessions.md`; mirrored in `docs/zh-CN/commands/sessions.md`.
  - Verified both `session-manager` and `session-aliases` load across install modes.

<sup>Written for commit 2e9c3ea3078d6fa94f0c185cc176d7f07f2d00da. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Documentation**
  * Updated session command documentation with new code examples reflecting standardized module resolution patterns.

* **Improvements**
  * Implemented enhanced plugin discovery mechanism that supports multiple installation paths and configurations, improving system reliability and compatibility across diverse deployment environments.

### DIFF
--- a/commands/sessions.md
+++ b/commands/sessions.md
@@ -25,8 +25,9 @@ Use `/sessions info` when you need operator-surface context for a swarm: branch,
 **Script:**
 ```bash
 node -e "
-const sm = require((process.env.CLAUDE_PLUGIN_ROOT||require('path').join(require('os').homedir(),'.claude'))+'/scripts/lib/session-manager');
-const aa = require((process.env.CLAUDE_PLUGIN_ROOT||require('path').join(require('os').homedir(),'.claude'))+'/scripts/lib/session-aliases');
+const _r=((e,p,f,h)=>{if(e&&e.trim())return e;const m='scripts/lib/session-manager.js';const cs=[p.join(h,'.claude'),p.join(h,'.claude','plugins','everything-claude-code'),p.join(h,'.claude','plugins','everything-claude-code@everything-claude-code'),p.join(h,'.claude','plugins','marketplace','everything-claude-code')];for(const c of cs)if(f.existsSync(p.join(c,m)))return c;const cd=p.join(h,'.claude','plugins','cache','everything-claude-code');try{for(const e of f.readdirSync(cd)){const n=p.join(cd,e);if(f.existsSync(p.join(n,m)))return n;try{for(const i of f.readdirSync(n)){const x=p.join(n,i);if(f.existsSync(p.join(x,m)))return x}}catch(_){}}}catch(_){}return p.join(h,'.claude')})(process.env.CLAUDE_PLUGIN_ROOT,require('path'),require('fs'),require('os').homedir());
+const sm = require(_r+'/scripts/lib/session-manager');
+const aa = require(_r+'/scripts/lib/session-aliases');
 const path = require('path');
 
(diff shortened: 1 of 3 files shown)


affaan-m:
Thanks for fixing the session command path resolution! The root issue (finding the plugin root across different install locations) is legitimate.

However, the approach has a readability concern: the path resolution logic is an inlined, heavily minified IIFE that's very difficult to read and maintain:

\`\`\`js
const _r=((e,p,f,h)=>{if(e&&e.trim())return e;const m='scripts/lib/session-manager.js';...
\`\`\`

This is repeated in every subcommand across both `commands/sessions.md` and `docs/zh-CN/commands/sessions.md`.

Suggestion: You've already created `scripts/lib/resolve-plugin-root.js` (good!). Could you use a readable require of that module instead of the inlined minified code? Something like:

\`\`\`js
const _r = require('./scripts/lib/resolve-plugin-root')();
\`\`\`

Or if the path to the module itself is the chicken-and-egg problem, at least format the inline version with readable variable names and line breaks.

AryanTejani:
sure will do

affaan-m:
Closing — superseded by recent changes. The session command path issues have been addressed in later PRs.

PR #651, opened 2026-03-19
fix(install): prefix language-specific rules to prevent overwriting common rules

### LINKED ISSUE #432: Bug: Language-specific rules overwrite common rules when installing multiple languages flat
## Bug Description

When installing common + multiple language-specific rules (e.g. Python + Go) using the `configure-ecc` skill or `install.sh`, language-specific rule files silently overwrite the common rules because they share the same filenames.

## Files Affected

The following files exist in both `rules/common/` and `rules/<language>/` with identical names:
- `coding-style.md`
- `security.md`
- `testing.md`
- `hooks.md`
- `patterns.md`

## What Happens

When installing flat into `~/.claude/rules/`:
```bash
cp -r rules/common/. ~/.claude/rules/
cp -r rules/python/. ~/.claude/rules/   # overwrites common files
cp -r rules/golang/. ~/.claude/rules/   # overwrites python files
```

The last language copied "wins". The common rules (globally scoped, no path frontmatter) are silently lost and replaced by language-scoped versions (`paths: ["**/*.py"]` etc.).

## Impact

- Common rules stop applying globally — they only apply to whichever language was installed last
- Users lose their globally-scoped rules without any warning
- The content is genuinely different: common rules have the full rules; language-specific ones are slim extensions that reference `common/` (which doesn't exist at the install target)

## Suggested Fix

One of:
1. **Prefix language-specific files** at install time: `python-coding-style.md`, `go-coding-style.md` etc.
2. **Detect collisions** in `install.sh` / `configure-ecc` and warn the user before overwriting
3. **Install into subdirectories**: `~/.claude/rules/python/coding-style.md` (requires Claude Code to support nested rule dirs)

Option 1 is the simplest workaround today. Option 2 would catch the problem without changing the file layout.

## Environment

- ECC version: 1.8.0
- Platform: Windows 11 / Claude Code
- Install method: `configure-ecc` skill

tonymfer:
## Summary
- Prefix language-specific rule files during install to prevent overwriting common rules
- Fix applied to the `configure-ecc` skill (English, Japanese, Chinese versions)
- Added troubleshooting note about the prefixed naming convention

## Why
When installing common + multiple language-specific rules via the `configure-ecc` skill, files with identical names (`coding-style.md`, `security.md`, `testing.md`, `hooks.md`, `patterns.md`) exist in both `rules/common/` and `rules/<language>/`. The `cp -r` in Step 3 silently overwrites common rules with language-specific ones, so the last language installed "wins" and common rules are lost.

The fix prefixes language-specific files at copy time (e.g., `python-coding-style.md` instead of `coding-style.md`), matching the approach already used by the Antigravity install target adapter in `install-executor.js`.

Closes #432

## Test plan
- [x] Install common + python + go rules — verify all three sets of files are present
- [x] Verify no filename collisions between common and language-specific rules
- [x] `validate-rules.js` and `validate-skills.js` CI validators pass

---
## Summary by cubic
Prefix language-specific rule files during install to stop overwriting common rules, and harden the copy commands for reliability. Updates the `configure-ecc` skill instructions (EN/JA/ZH) and troubleshooting for the new prefixed naming.

- **Bug Fixes**
  - Step 3 now copies language rules into `$TARGET/rules/` with a `<lang>-` prefix (e.g., `python-coding-style.md`), preserving common rules and matching the install target adapter in `install-executor.js`.
  - Copy loops now use `cp -r`, quote `$ECC_ROOT`, and add `[ -e "$f" ]` guards to handle subdirectories, paths with spaces, and empty globs.

<sup>Written for commit f7a3311e6a588763ecc974a3ec6d793e6e50ad50. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Documentation**
  * Clarified skill configuration docs to explain language-specific rule files are now copied with language-prefixed names to avoid overwriting common rules.
  * Updated troubleshooting section with corrected examples showing the new prefixed naming and how to verify installed rules.

### DIFF
--- a/docs/ja-JP/skills/configure-ecc/SKILL.md
+++ b/docs/ja-JP/skills/configure-ecc/SKILL.md
@@ -159,9 +159,9 @@ cp -r $ECC_ROOT/rules/common/* $TARGET/rules/
 
 # 言語固有のルール（共通ルールの上書きを防ぐためプレフィックス付き）
 # coding-style.md は typescript-coding-style.md のようになります
-for f in $ECC_ROOT/rules/typescript/*; do cp "$f" "$TARGET/rules/typescript-$(basename "$f")"; done   # 選択された場合
-for f in $ECC_ROOT/rules/python/*; do cp "$f" "$TARGET/rules/python-$(basename "$f")"; done           # 選択された場合
-for f in $ECC_ROOT/rules/golang/*; do cp "$f" "$TARGET/rules/golang-$(basename "$f")"; done           # 選択された場合
+for f in "$ECC_ROOT"/rules/typescript/*; do [ -e "$f" ] && cp -r "$f" "$TARGET/rules/typescript-$(basename "$f")"; done   # 選択された場合
+for f in "$ECC_ROOT"/rules/python/*; do [ -e "$f" ] && cp -r "$f" "$TARGET/rules/python-$(basename "$f")"; done           # 選択された場合
+for f in "$ECC_ROOT"/rules/golang/*; do [ -e "$f" ] && cp -r "$f" "$TARGET/rules/golang-$(basename "$f")"; done           # 選択された場合
 ```
 
 **重要**: ユーザーが言語固有のルールを選択したが、共通ルールを選択しなかった場合、警告します：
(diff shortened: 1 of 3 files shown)


affaan-m:
Thanks for identifying this real issue -- language-specific rules overwriting common rules is a legitimate bug.

However, this PR only updates the skill documentation (configure-ecc/SKILL.md and its translations) but does not update the actual install scripts (install.sh, install.ps1, scripts/install-*.js). The skill docs describe the *expected* behavior, but the install scripts that actually perform the copy still use the old flat-copy approach.

To complete this fix, please also update:
1. `install.sh` -- add prefixing logic for language-specific rule copies
2. `install.ps1` -- same for Windows
3. Any Node.js install scripts that handle rule copying

Once the actual install scripts match the documented behavior, this is ready to merge.

affaan-m:
Superseded by rebased PR

PR #757, opened 2026-03-22
perf(hooks): move post-edit-format and post-edit-typecheck to strict-only

### LINKED ISSUE #735: perf(hooks): post-edit-format and post-edit-typecheck add significant latency on multi-file refactors
## Problem

`post-edit-format.js` and `post-edit-typecheck.js` fire synchronously on every `Edit` tool call with a 15-30 second timeout each. During multi-file refactors (10-20 edits in sequence), this adds 5-10 minutes of cumulative hook latency.

The hooks are individually useful, but the per-edit firing model is wrong for batch operations. Running `biome check --write` after every single line edit is wasteful when the agent is about to edit 15 more files.

## Observed behavior

- Agent edits `src/auth/login.ts` -> waits 15s for format + 30s for typecheck
- Agent edits `src/auth/session.ts` -> waits 15s + 30s again
- Repeat for 10 files = ~7.5 minutes of hook overhead

## Suggested fixes

1. **Debounce**: accumulate edited files, run format/typecheck once after a 5-second idle period (no edits for 5s)
2. **Batch at Stop**: move formatting to a Stop hook that runs once on all modified files before the response completes
3. **Profile-based**: only enable these in `strict` profile, not `standard` (the run-with-flags system already supports this)

Option 3 is the simplest and works with the existing hook profile system.

ghost:
Fixes #735

The \`post-edit-format\` and \`post-edit-typecheck\` hooks fire synchronously on every \`Edit\` tool call with 15-30s timeouts. During multi-file refactors (10-20 edits), this adds 5-10 minutes of cumulative hook overhead.

Changed both hooks from \`"standard,strict"\` to \`"strict"\` in \`hooks.json\`. This means they won't fire in the default profile but are still available for users who want per-edit validation by enabling the strict profile.

This is option 3 from the issue (simplest fix using the existing run-with-flags profile system).

---
## Summary by cubic
Move `post-edit-format` and `post-edit-typecheck` to the `strict` profile only and update the OpenCode plugin to match, removing long per-edit delays in the default profile. Fixes #735 and avoids 15–30s timeouts on every edit, saving 5–10 minutes during multi-file refactors.

- **Migration**
  - To keep per-edit formatting and type checks, run with the `strict` profile (e.g., `--profile strict`).

<sup>Written for commit ad0d0837c10b27c8b775f59bde24ea86722c1656. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Chores**
  * Tightened post-edit validation: formatting and type-checking now run under a stricter profile, resulting in more stringent automatic format and typecheck enforcement after edits to JavaScript/TypeScript files.

### DIFF
--- a/.opencode/plugins/ecc-hooks.ts
+++ b/.opencode/plugins/ecc-hooks.ts
@@ -77,7 +77,7 @@ export const ECCHooksPlugin = async ({
       editedFiles.add(event.path)
 
       // Auto-format JS/TS files
-      if (hookEnabled("post:edit:format", ["standard", "strict"]) && event.path.match(/\.(ts|tsx|js|jsx)$/)) {
+      if (hookEnabled("post:edit:format", ["strict"]) && event.path.match(/\.(ts|tsx|js|jsx)$/)) {
         try {
           await $`prettier --write ${event.path} 2>/dev/null`
           log("info", `[ECC] Formatted: ${event.path}`)
@@ -116,7 +116,7 @@ export const ECCHooksPlugin = async ({
     ) => {
       // Check if a TypeScript file was edited
       if (
-        hookEnabled("post:edit:typecheck", ["standard", "strict"]) &&
+        hookEnabled("post:edit:typecheck", ["strict"]) &&
         input.tool === "edit" &&
         input.args?.filePath?.match(/\.tsx?$/)
       ) {


ghost:
Good catch from the bot. Updated `.opencode/plugins/ecc-hooks.ts` lines 80 and 119 to match. Both harnesses are now consistent.

affaan-m:
Thanks for the perf optimization. Will review this week.

PR #758, opened 2026-03-22
feat(hooks): add config protection hook to block linter config manipulation

### LINKED ISSUE #733: feat(hooks): add config protection hook to prevent linter config manipulation
## Problem

Agents frequently modify linter/formatter configs (`.eslintrc`, `biome.json`, `.prettierrc`, `.ruff.toml`, etc.) to make checks pass instead of fixing the actual code. This is a known failure mode documented in the community but not addressed by any existing hook.

## Proposed Solution

A `PreToolUse` hook on `Write|Edit` that blocks modifications to linter/formatter config files with a steering message:

```bash
#!/bin/bash
set -euo pipefail
FILE_PATH=$(jq -r '.tool_input.file_path // .tool_input.file // empty')
[[ -z "$FILE_PATH" ]] && exit 0
BASENAME=$(basename "$FILE_PATH")
case "$BASENAME" in
  .eslintrc*|.prettierrc*|biome.json|.ruff.toml|ruff.toml|\
  .shellcheckrc|.stylelintrc*|.markdownlint*)
    echo "BLOCKED: Do not modify linter/formatter config to pass checks. Fix the code instead." >&2
    exit 2
    ;;
esac
exit 0
```

This is a common issue across Claude Code setups. We built and tested this hook: [buildingopen/blast-radius](https://github.com/buildingopen/blast-radius) has the pattern, and it is running in production.

Happy to PR this if there is interest.

ghost:
Fixes #733

Adds a \`PreToolUse\` hook that blocks Write/Edit/MultiEdit calls targeting linter and formatter config files. When triggered, it returns exit code 2 with a steering message that tells the agent to fix the code instead.

**Covered configs:** ESLint (\`.eslintrc*\`, \`eslint.config.*\`), Prettier (\`.prettierrc*\`, \`prettier.config.*\`), Biome (\`biome.json\`), Ruff (\`.ruff.toml\`, \`ruff.toml\`), ShellCheck, Stylelint, Markdownlint.

**Implementation:** Node.js hook following the same patterns as the existing hooks (stdin JSON parsing, run-with-flags integration). Runs on both standard and strict profiles with a 5s timeout.

Note: \`pyproject.toml\` is in the protected set but excluded from blocking since it contains project metadata too, not just linter config.

---
## Summary by cubic
Add a PreToolUse config‑protection hook that blocks Write/Edit/MultiEdit attempts on linter and formatter configs, steering the agent to fix code instead. Expands coverage (ESLint TS flat configs, Prettier ESM) and explicitly excludes `pyproject.toml`.

- **New Features**
  - Blocks edits to ESLint (`.eslintrc*`, `eslint.config.*`), Prettier (`.prettierrc*`, `prettier.config.*`), Biome (`biome.json*`), Ruff (`.ruff.toml`, `ruff.toml`), ShellCheck, Stylelint, Markdownlint.
  - Coverage: added `eslint.config.ts`/`mts`/`cts` and `prettier.config.mjs`; `pyproject.toml` is excluded from protection.
  - Returns exit code 2 with a clear steering message; non-config edits pass. Enabled for `standard`/`strict` (5s). Exports `run()` for in‑process use via `run-with-flags`. Logs a warning and skips blocking if stdin >1MB.

<sup>Written for commit 1c796852431dec5b2b64b9f2ecd15289a1292950. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Adds a pre-operation guard that blocks write/edit actions on common linter/formatter config files (ESLint, Prettier, Biome, Ruff, shellcheck/stylelint/markdownlint configs, pyproject.toml).
  * If the guard cannot parse the operation payload or receives truncated input, it leaves the action unchanged and emits a warning instead of blocking.

### DIFF
--- a/scripts/hooks/config-protection.js
+++ b/scripts/hooks/config-protection.js
@@ -87,15 +87,28 @@ function run(input) {
 module.exports = { run };
 
 // Stdin fallback for spawnSync execution
+let truncated = false;
 process.stdin.setEncoding('utf8');
 process.stdin.on('data', chunk => {
   if (raw.length < MAX_STDIN) {
     const remaining = MAX_STDIN - raw.length;
     raw += chunk.substring(0, remaining);
+    if (chunk.length > remaining) truncated = true;
+  } else {
+    truncated = true;
   }
 });
 
 process.stdin.on('end', () => {
+  // If stdin was truncated, the JSON is likely malformed. Fail open but
+  // log a warning so the issue is visible. The run() path (used by
+  // run-with-flags.js in-process) is not affected by this.
+  if (truncated) {
+    process.stderr.write('[config-protection] Warning: stdin exceeded 1MB, skipping check\n');
+    process.stdout.write(raw);
+    return;
+  }
+
   try {
     const input = raw.trim() ? JSON.parse(raw) : {};
     const result = run(input);


ghost:
Addressed all the Greptile feedback:

1. **Dead code**: Removed `pyproject.toml` from `PROTECTED_FILES` and removed the `PARTIAL_CONFIG_FILES` set entirely. Added a comment explaining why pyproject.toml is intentionally excluded (it has project metadata, not just linter config).

2. **Missing configs**: Added `prettier.config.mjs`, `eslint.config.ts`, `eslint.config.mts`, and `eslint.config.cts`.

3. **Performance**: Exported a `run()` function so `run-with-flags.js` can use the in-process path instead of falling back to `spawnSync`.

ghost:
The P2s (prettier.config.mjs, ESLint TS configs, pyproject.toml, run() export) were all fixed in 7ff116f.

New commit addresses the P1 stdin truncation: now detects when input exceeds 1MB and logs a warning instead of silently failing open. The in-process run() path (used by run-with-flags.js) isn't affected since it gets the parsed object directly.

Re the Bash bypass P1: an agent could do `echo '' > .eslintrc.json` via Bash. But blocking Bash commands that touch config files would require parsing shell command strings, which is fragile and would false-positive on reads like `cat .eslintrc.json`. This is a known limitation of file-path-based protection, not specific to this hook.

affaan-m:
Good idea for config protection. Will review this week.

PR #890, opened 2026-03-24
fix(clv2): support CLV2_CONFIG override in start-observer.sh

AxelBriche:
## Summary

- `start-observer.sh` hardcodes `CONFIG_FILE="${SKILL_ROOT}/config.json"` but `observe.sh` already supports a `CLV2_CONFIG` environment variable override
- When ECC is installed as a **plugin** (via marketplace), `SKILL_ROOT` resolves to `~/.claude/plugins/marketplaces/everything-claude-code/skills/continuous-learning-v2/` instead of `~/.claude/skills/continuous-learning-v2/`
- Users who set `observer.enabled: true` in their **user-level** config (`~/.claude/skills/.../config.json`) find that the observer never starts because `start-observer.sh` reads the **plugin-level** config (which has `enabled: false`)

## Root cause

`observe.sh` (lines 314-316) supports the override:
```bash
if [ -n "${CLV2_CONFIG:-}" ]; then
  CONFIG_FILE="$CLV2_CONFIG"
fi
```

But `start-observer.sh` (line 39) does not:
```bash
CONFIG_FILE="${SKILL_ROOT}/config.json"  # always reads plugin config
```

## Fix

Apply the same `CLV2_CONFIG` override pattern to `start-observer.sh` for consistency.

## Additional note: macOS lockfile dependency

The observer lazy-start in `observe.sh` requires either `flock` (Linux) or `lockfile` (macOS, via `procmail`). On a fresh macOS install, neither is available, so the lazy-start **silently skips** without any warning. This might warrant a separate issue for:
1. Documenting `brew install procmail` as a prerequisite on macOS
2. Or adding a fallback that doesn't require a lock (e.g., mkdir-based lock)

## Test plan

- [x] Set `CLV2_CONFIG` env var in `~/.claude/settings.json`
- [x] Verify `start-observer.sh start` reads the correct config
- [x] Verify observer process starts with `enabled: true`
- [x] Verify observations are being captured and analyzed

---
## Summary by cubic
Make `start-observer.sh` honor `CLV2_CONFIG` so the observer reads the user-level config and starts when enabled, even when ECC is installed as a plugin. Fix macOS sandboxing by writing analysis/prompt files to `PROJECT_DIR` and launching the `claude` agent with `--add-dir` so it can read/write instincts.

<sup>Written for commit b7016da89dce7a2c65b935a01702b719aed0cd77. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Continuous learning observer now supports a configuration override via an environment variable, letting startup behavior be customized without changing defaults.
  * Observer stores analysis and prompt artifacts in the project directory so they are accessible to the agent when needed.

* **Bug Fixes**
  * Cleanup and file-access behavior updated to handle the new storage location reliably.

### DIFF
--- a/skills/continuous-learning-v2/agents/observer-loop.sh
+++ b/skills/continuous-learning-v2/agents/observer-loop.sh
@@ -54,13 +54,14 @@ analyze_observations() {
 
   # Sample recent observations instead of loading the entire file (#521).
   # This prevents multi-MB payloads from being passed to the LLM.
+  # Files are created in PROJECT_DIR (not TMPDIR) so the sandboxed Claude agent can read them.
   MAX_ANALYSIS_LINES="${ECC_OBSERVER_MAX_ANALYSIS_LINES:-500}"
-  analysis_file="$(mktemp "${TMPDIR:-/tmp}/ecc-observer-analysis.XXXXXX.jsonl")"
+  analysis_file="${PROJECT_DIR}/analysis-pending.jsonl"
   tail -n "$MAX_ANALYSIS_LINES" "$OBSERVATIONS_FILE" > "$analysis_file"
   analysis_count=$(wc -l < "$analysis_file" 2>/dev/null || echo 0)
   echo "[$(date)] Using last $analysis_count of $obs_count observations for analysis" >> "$LOG_FILE"
 
-  prompt_file="$(mktemp "${TMPDIR:-/tmp}/ecc-observer-prompt.XXXXXX")"
+  prompt_file="${PROJECT_DIR}/observer-prompt.tmp"
   cat > "$prompt_file" <<PROMPT
 Read ${analysis_file} and identify patterns for the project ${PROJECT_NAME} (user corrections, error resolutions, repeated workflows, tool preferences).
 If you find 3+ occurrences of the same pattern, create an instinct file in ${INSTINCTS_DIR}/<id>.md.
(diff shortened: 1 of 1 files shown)


AxelBriche:
**Note on temp file naming:** The original code used `mktemp` with `XXXXXX` random suffixes to prevent file collisions. The fix uses fixed filenames (`analysis-pending.jsonl`, `observer-prompt.tmp`) in `PROJECT_DIR` instead, which is safe because:

- The PID file + re-entrancy guard (`ANALYZING=0/1`) ensure only one observer runs per project
- Files are cleaned up after each cycle (`rm -f`)

If the maintainers prefer to keep random suffixes for extra safety, the fix can use `mktemp "${PROJECT_DIR}/analysis-pending.XXXXXX.jsonl"` instead — the key change is `PROJECT_DIR` instead of `TMPDIR`, not the naming pattern.

affaan-m:
thanks for the pr. triaged and queued for maintainer review.

affaan-m:
thanks for the contribution. there are merge conflicts with the latest `main`. can you rebase on the current branch head and push again? once that is done we can take another look.

affaan-m:
Superseded by #999. The original branch conflicts with current main and also carries stale observer-loop changes that current main already solved differently. #999 keeps only the still-missing CLV2_CONFIG override on top of current main.

PR #931, opened 2026-03-26
docs: clarify multi-model command setup

KT-lcz:
Document that multi-* commands require the ccg-workflow runtime so users know they must initialize the extra wrapper and prompt assets before use.


affaan-m:

PR #932, opened 2026-03-26
docs: fix rule installation examples

KT-lcz:
Clarify that manual rule installation should preserve the rules directory structure so references keep working and filename collisions are avoided.

## Summary by CodeRabbit

* **Documentation**
  * Clarified installation steps in English and Chinese READMEs: create the destination rules directory first, then copy entire rule directories (e.g., rules/common, rules/<language>) rather than individual files. This preserves relative references and prevents filename collisions across rule sets.


affaan-m:

PR #948, opened 2026-03-26
fix(#843): session-manager hooks — ecc-sessions dir + hookSpecificOutput format

### LINKED ISSUE #843: session-manager scripts don't work
I installed ECC as I was specifically interested in the session start/end hooks. However they didn't work out of the box (with claude 2.1.81). I spent a while with claude diagnosing the issue. There were 3 problems, the main one was it looks like Claude Code performs a clean-up on exit which was removing the file written by the session end script. Fixing this revealed a second issue where the json format generated by session start appears to have changed as well. Finally the installer didn't appear to install the lib folder required by the hooks.

Bug 1: Installer doesn't deploy scripts/lib/

The hooks-runtime install module configures hook entries in settings.json but has 0 files — it doesn't copy scripts/hooks/*.js or scripts/lib/ (utils.js, session-manager.js, package-manager.js, etc.). The hooks all require('../lib/utils') and fail immediately on a fresh install. Had to manually cp -r the lib folder.

Bug 2: Session files deleted by Claude Code

getSessionsDir() returns ~/.claude/sessions/, which is Claude Code's managed directory. Claude Code automatically deletes any file with a YYYY-MM-DD date prefix from this directory on session exit (undocumented - discovered by experimentation). Since session files are named 2026-03-24-projectname-session.tmp, they're always cleaned up.

Fix: Change getSessionsDir() to return ~/.claude/session-data/ (or any directory Claude Code doesn't manage).

Bug 3: SessionStart hook uses wrong JSON output format

session-start.js uses output() from utils.js which calls console.log(text) — plain text to stdout. But SessionStart hooks require the hookSpecificOutput wrapper:

```json
{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": "..."}}
```

The `{"additionalContext": "..."}` format only works for UserPromptSubmit hooks, not SessionStart.

Additionally, `process.exit(0)` can race stdout — should use `process.stdout.write()` with a flush callback.

anuragg-saxenaa:
## Summary

Fixes #843 — session hooks installed but not working out of the box.

Two root causes identified and fixed:

### Fix 1: Session directory collision with Claude Code cleanup
Claude Code performs cleanup on its built-in `~/.claude/sessions/` directory on exit, which removes files written by the `session-end` hook before the next session can read them.

**Fix:** Renamed the sessions directory from `sessions` to `ecc-sessions` so ECC-managed files live outside Claude Code's cleanup scope.

```js
// Before
return path.join(getClaudeDir(), 'sessions');
// After
return path.join(getClaudeDir(), 'ecc-sessions');
```

### Fix 2: SessionStart hook output format
Claude Code's `SessionStart` hooks require JSON output with a `hookSpecificOutput` envelope. Plain `console.log(string)` output is silently discarded — no context is injected.

**Fix:** Updated `session-start.js` to use `process.stdout.write(JSON.stringify({hookSpecificOutput: {hookEventName, additionalContext}}))` for context injection.

## Test plan
- [x] Install ECC, run a Claude session, end it
- [x] Verify `~/.claude/ecc-sessions/` contains a session file after session ends
- [x] Start a new session — verify previous session context appears in Claude's context window
- [x] Check `~/.claude/ecc-sessions/` still exists (not cleaned up by Claude Code)

---
## Summary by cubic
Makes session hooks work out of the box by moving session storage out of Claude Code’s cleanup, deduping recent sessions across directories, and emitting a single `hookSpecificOutput` JSON with accumulated context at session start. Fixes #843.

- Bug Fixes
  - Store session data in `~/.claude/session-data/` and search both canonical and legacy `sessions` dirs to dedupe recent files, avoiding cleanup-related losses.
  - Accumulate context (previous summary, learned skills, aliases, project info) and emit exactly one `hookSpecificOutput` JSON via a safe final write; remove early stdout writes and sanitize session IDs for valid, stable filenames.

<sup>Written for commit 2546f96896e8a7b9493c9a7785d39a9ae3da67de. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Canonicalized and reorganized session storage locations with backwards-compatible search of legacy data.
  * Session startup now emits a single structured JSON payload containing combined session/context info instead of scattered text output.
  * Session IDs and derived identifiers are now sanitized for safety and consistency.

* **Bug Fixes**
  * Safer handling of standard output and process termination to avoid hangs on errors.

### DIFF
--- a/scripts/hooks/session-start.js
+++ b/scripts/hooks/session-start.js
@@ -11,6 +11,7 @@
 
 const {
   getSessionsDir,
+  getSessionSearchDirs,
   getLearnedSkillsDir,
   findFiles,
   ensureDir,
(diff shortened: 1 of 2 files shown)


affaan-m:

affaan-m:
Closing as already landed on current main. The session-start payload/session-data work and the utils session-directory/sanitization changes from this PR are already present upstream in scripts/hooks/session-start.js and scripts/lib/utils.js, so keeping this branch open would duplicate code that has already shipped.

PR #1010, opened 2026-03-30
fix: resolve stop hooks without plugin root env

affaan-m:
## Summary
- replace Stop and SessionEnd hook commands with inline root resolvers like SessionStart
- keep lifecycle hooks working when `CLAUDE_PLUGIN_ROOT` is empty during plugin execution
- add hook regression coverage for unset-plugin-root installs

## Test Plan
- node scripts/ci/validate-hooks.js
- node tests/hooks/hooks.test.js
- node tests/integration/hooks.test.js

---
## Summary by cubic
Fixes Stop and SessionEnd hooks by inlining the resolver so they work when `CLAUDE_PLUGIN_ROOT` is unset. Adds regression and integration tests to verify plugin root resolution from the local cache.

- **Bug Fixes**
  - Replaced Stop and SessionEnd commands with an inline `node -e` resolver (same pattern as SessionStart).
  - Resolver locates `scripts/hooks/run-with-flags.js` by probing `~/.claude` plugin directories and the plugin cache when the env var is empty, with a safe fallback.
  - Updated tests to allow inline resolvers and added an integration test that runs Stop hooks with an empty `CLAUDE_PLUGIN_ROOT`.

<sup>Written for commit 08a5ae6784ec941804b8092ed3955f756d751512. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Chores**
  * Improved plugin hook resolution mechanism for better reliability and compatibility.
  * Updated hook configuration for Stop lifecycle events to use enhanced path resolution.

* **Tests**
  * Added integration tests for Stop hook path resolution and fallback scenarios.
  * Enhanced hook validation tests to cover lifecycle hook implementations.


affaan-m:
Superseded by merged PR #1011.

PR #1078, opened 2026-04-01
fix(installer): add hook id field for semantic deduplication on reinstall

QWsin:
## Summary

Adds a stable `id` field to each hook in `hooks/hooks.json` and updates the installer to deduplicate hooks by `id` instead of by exact `JSON.stringify` match. This prevents duplicate hooks from accumulating across reinstalls.

## What Changed

- `hooks/hooks.json`: Added an `id` field to all 31 hook entries (e.g., `stop:session-end`, `pre:bash:block-no-verify`). IDs are stable logical identifiers independent of description text.
- `scripts/lib/install/apply.js`: Replaced the `mergeHookEntries` dedup logic — hooks with an `id` are deduplicated by that ID (existing entries take precedence); hooks without an `id` fall back to exact-match dedup for backward compatibility.

## Why This Change

The installer's `mergeHookEntries()` uses `JSON.stringify(entry)` for deduplication. Two hook entries are considered "the same" only if every field is byte-for-byte identical. When ECC updates any field of a hook — its command string, description, timeout, or async flag — the new entry is treated as distinct and appended instead of replaced.

**Real-world consequence**: A user who installed ECC before Stop/SessionEnd hooks were changed to use inline `node -e` launchers would have both the old `${CLAUDE_PLUGIN_ROOT}` variant and the new inline variant of the same hook in `settings.json`. Both are kept because the command strings differ. When Claude Code fires the Stop hook, both variants execute — duplicate desktop notifications, duplicate session saves, etc.

Repeated reinstalls compound this: each install appends rather than updates, gradually bloating `settings.json` and degrading hook execution performance.

**Additional observation (not changed in this PR)**: The Stop and SessionEnd hooks in the canonical `hooks/hooks.json` ship with inline `node -e` launcher scripts that re-implement the plugin-root resolution logic. When `CLAUDE_PLUGIN_ROOT` is correctly set by Claude Code (which it always is during normal hook execution), this fallback chain is never reached. This is noted as a potential area for future simplification — this PR does not address it.

## Testing Done

- [x] Manual testing completed — deduplication behavior verified against `~/.claude/settings.json`
- [x] Automated tests pass locally (`node tests/hooks/hooks.test.js` — 218 passed, 0 failed; `node tests/scripts/install-apply.test.js` — 20 passed, 0 failed)
- [x] Edge cases considered — legacy hooks without `id` fall back to exact-match dedup; user customizations are preserved since existing entries take precedence

## Type of Change

- [x] `fix:` Bug fix
- [ ] `feat:` New feature
- [ ] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `ci:` CI/CD changes

- [x] JSON files validate cleanly (`hooks/hooks.json` parses as valid JSON; `settings.json` schema warnings for `id`/`description` fields are from the upstream schema which doesn't recognize these fields — harmless)
- [ ] Shell scripts pass shellcheck (if applicable)
- [ ] Pre-commit hooks pass locally (if configured)

## Documentation

- [ ] Updated relevant documentation (no doc files were modified — the `id` field is an internal contract between `hooks/hooks.json` and the installer)
- [ ] Added comments for complex logic (`mergeHookEntries` is straightforward id-based dedup; no additional comments needed)
- [ ] README updated (if needed) (README covers usage and installation, not internal hook deduplication mechanics)

## Summary by CodeRabbit

* **New Features**
  * Hook entries now show explicit unique identifiers across lifecycle stages, making hooks easier to identify in listings.

* **Improvements**
  * Hook configuration merging now deduplicates by identifier when present and preserves original ordering; malformed or invalid hook entries are skipped and reported via logs.

### DIFF
--- a/scripts/lib/install/apply.js
+++ b/scripts/lib/install/apply.js
@@ -33,6 +33,10 @@ function mergeHookEntries(existingEntries, incomingEntries) {
       console.error(`[mergeHookEntries] Invalid hook entry: expected object, got ${JSON.stringify(entry)} — skipping`);
       continue;
     }
+    if ('id' in entry && typeof entry.id !== 'string') {
+      console.error(`[mergeHookEntries] Invalid hook entry: id must be string, got ${typeof entry.id} — skipping`);
+      continue;
+    }
     const id = entry.id;
     if (id !== undefined) {
       if (!byId.has(id)) {


QWsin:
Thanks for the review. Addressed both P2 findings in follow-up commits:

1. **Legacy fallback asymmetry** — Fixed by adding `if (!byId.has(key))` guard to the legacy branch, so both paths now consistently use first-write-wins (existing entries take precedence).

2. **Migration gap** — Documented in code comment: first reinstall from a pre-id version may retain one duplicate per hook; effective deduplication resumes on the second reinstall.

Tests pass: `node tests/scripts/install-apply.test.js` (20 passed) and `node tests/hooks/hooks.test.js` (218 passed).

Note: `rules/python/coding-style.md` was reverted in a separate commit — it was an unrelated change included by mistake in the initial PR.

QWsin:
Also addressed: added explicit object-shape validation in mergeHookEntries before accessing entry.id, so malformed entries (null, undefined, non-object) fail fast with a clear error message rather than throwing TypeError.

QWsin:
Updated: invalid hook entries now log a warning to stderr and are skipped (continue) rather than throwing, so a malformed entry in settings.json does not block the entire installation.

QWsin:
Also fixed: added validation that entry.id must be a string when present — guards against non-string id values (object, number) being used as Map keys.

affaan-m:
Closing as superseded by direct port to main. The useful core landed, but with a safer implementation: managed hook entries now dedupe by stable ids plus semantic fallback aliases, so older no-id installs migrate cleanly on the first reinstall. Regression coverage was added in tests/scripts/install-apply.test.js.

PR #1106, opened 2026-04-01
fix: make Codex context7 key extraction portable

### LINKED ISSUE #1096: Bug: sync-ecc-to-codex.sh uses grep -P and breaks on macOS BSD grep
## Summary
`extract_context7_key()` uses `grep -P`, which is not supported by BSD `grep` on macOS. During `sync-ecc-to-codex.sh`, this prints an error during MCP normalization and prevents portable extraction of an existing Context7 key.

## Version
- ECC tag: `v1.9.0`
- ECC commit: `29277ac273f294fd4804c35e43af9c8a5fc5ba9d`

## Reproduction
1. Use macOS / BSD `grep`.
2. Run:
   ```bash
   bash scripts/sync-ecc-to-codex.sh
   ```
3. During `Normalizing MCP server config to pnpm`, observe:
   ```text
   grep: invalid option -- P
   usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] ...
   ```
4. Inspect the implementation:
   - `scripts/sync-ecc-to-codex.sh:104-107`
   - line 106 uses:
   ```bash
   grep -oP -- '--key",[[:space:]]*"\K[^"]+' "$file" | head -n 1 || true
   ```

## Observed behavior
The sync script emits a portability error on macOS and cannot reliably preserve an existing Context7 key via this code path.

## Expected behavior
The script should use a portable parser on macOS and Linux, for example `awk`, `sed`, or `perl`.

## Evidence
- `scripts/sync-ecc-to-codex.sh:106` uses `grep -P`
- The failure happens on a stock macOS environment with BSD `grep`

## System information
- OS: macOS 26.4 (build 25E246)
- Architecture: arm64
- Shell: `/bin/zsh`
- Git: `2.53.0`
- grep: `grep (BSD grep, GNU compatible) 2.6.0-FreeBSD`
- Node.js: `v25.7.0`
- npm: `11.10.1`
- GitHub CLI: `2.88.1`
- Codex CLI: `0.118.0`

mvillmow:
## Summary
- replace the non-portable `grep -P` Context7 key extraction with a portable Perl parser
- add a regression test that fails if `grep -P` is reintroduced into `extract_context7_key`

## Testing
- `node tests/scripts/sync-ecc-to-codex.test.js`

Closes #1096.

---
## Summary by cubic
Make Context7 key extraction portable by replacing `grep -P` with a small Perl parser in `scripts/sync-ecc-to-codex.sh`. Fixes BSD/macOS failures and adds stronger tests to prevent regressions.

- **Bug Fixes**
  - Replace `grep -oP` with a Perl one-liner to parse the `--key` value (returns first match).
  - Add a test that verifies `extract_context7_key` avoids `grep -P`.
  - Add an end-to-end test that runs `extract_context7_key` and asserts the extracted key output.

<sup>Written for commit 93d5de02b41b0c9f660966a5f0eb401bc8149210. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Refactor**
  * Reworked a shell extraction implementation to use a more reliable inline extraction method, improving portability and consistent single-match behavior.

* **Tests**
  * Added tests that validate the new extraction implementation and execute it in a temporary environment to assert expected output and exit status.

### DIFF
--- a/tests/scripts/sync-ecc-to-codex.test.js
+++ b/tests/scripts/sync-ecc-to-codex.test.js
@@ -4,63 +4,43 @@
 
 const assert = require('assert');
 const fs = require('fs');
+const os = require('os');
 const path = require('path');
+const { spawnSync } = require('child_process');
 
 const scriptPath = path.join(__dirname, '..', '..', 'scripts', 'sync-ecc-to-codex.sh');
 const source = fs.readFileSync(scriptPath, 'utf8');
 const normalizedSource = source.replace(/\r\n/g, '\n');
-const runOrEchoSource = (() => {
-  const start = normalizedSource.indexOf('run_or_echo() {');
+
+function extractFunctionSource(text, functionName) {
+  const start = text.indexOf(`${functionName}() {`);
   if (start < 0) {
     return '';
   }
 
   let depth = 0;
-  let bodyStart = normalizedSource.indexOf('{', start);
+  const bodyStart = text.indexOf('{', start);
   if (bodyStart < 0) {
     return '';
   }
 
-  for (let i = bodyStart; i < normalizedSource.length; i++) {
-    const char = normalizedSource[i];
+  for (let i = bodyStart; i < text.length; i += 1) {
+    const char = text[i];
     if (char === '{') {
       depth += 1;
     } else if (char === '}') {
       depth -= 1;
       if (depth === 0) {
-        return normalizedSource.slice(start, i + 1);
+        return text.slice(start, i + 1);
       }
     }
   }
 
   return '';
-})();
-const extractContext7KeySource = (() => {
-  const start = normalizedSource.indexOf('extract_context7_key() {');
-  if (start < 0) {
-    return '';
-  }
+}
 
-  let depth = 0;
-  let bodyStart = normalizedSource.indexOf('{', start);
-  if (bodyStart < 0) {
-    return '';
-  }
-
-  for (let i = bodyStart; i < normalizedSource.length; i++) {
-    const char = normalizedSource[i];
-    if (char === '{') {
-      depth += 1;
-    } else if (char === '}') {
-      depth -= 1;
-      if (depth === 0) {
-        return normalizedSource.slice(start, i + 1);
-      }
-    }
-  }
-
-  return '';
-})();
+const runOrEchoSource = extractFunctionSource(normalizedSource, 'run_or_echo');
+const extractContext7KeySource = extractFunctionSource(normalizedSource, 'extract_context7_key');
 
 function test(name, fn) {
   try {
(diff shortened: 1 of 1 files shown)


scripts/sync-ecc-to-codex.sh
```diff
@@ -106,7 +106,7 @@ extract_toml_value() {
 
 extract_context7_key() {
   local file="$1"
-  grep -oP -- '--key",[[:space:]]*"\K[^"]+' "$file" | head -n 1 || true
+  perl -ne 'if (/--key",\s*"([^"]+)"/) { print "$1\n"; exit }' "$file" || true
```
bri-tong, line 109: Good replacement. Two notes:

1. The new Perl regex is slightly **stricter** than the old `grep` version — it requires a closing `"` after the value (`([^"]+)"`), while the old `\K[^"]+` would match values at EOF without a closing quote. This is arguably better (more correct parsing), but worth being aware of if the input files could have malformed entries.

2. `exit` inside the Perl one-liner is a clean way to replicate `head -n 1` behavior — stops after first match, avoids processing the rest of the file. 👍

tests/scripts/sync-ecc-to-codex.test.js
```diff
@@ -35,6 +35,32 @@ const runOrEchoSource = (() => {
...
+    } else if (char === '}') {
+      depth -= 1;
+      if (depth === 0) {
+        return normalizedSource.slice(start, i + 1);
+      }
+    }
+  }
+
+  return '';
+})();
```
bri-tong: This brace-matching IIFE is identical to `runOrEchoSource` above (lines 13-37) except for the function name string. Consider extracting a shared helper:

```js
function extractFunctionSource(source, fnName) {
  const start = source.indexOf(`${fnName}() {`);
  if (start < 0) return '';
  let depth = 0;
  let bodyStart = source.indexOf('{', start);
  if (bodyStart < 0) return '';
  for (let i = bodyStart; i < source.length; i++) {
    if (source[i] === '{') depth++;
    else if (source[i] === '}' && --depth === 0)
      return source.slice(start, i + 1);
  }
  return '';
}
```

Low priority — the file is small and this works fine as-is.

tests/scripts/sync-ecc-to-codex.test.js
```diff
@@ -74,6 +100,12 @@ function runTests() {
     assert.ok(!source.includes('run_or_echo cp -R "$skill_dir" "$dest"'), 'skill sync cp should be removed');
   })) passed++; else failed++;
 
+  if (test('extract_context7_key avoids non-portable grep -P', () => {
+    assert.ok(extractContext7KeySource, 'Expected to locate extract_context7_key function body');
+    assert.ok(!extractContext7KeySource.includes('grep -oP'), 'extract_context7_key should not rely on grep -P');
+    assert.ok(extractContext7KeySource.includes("perl -ne"), 'extract_context7_key should use a portable parser');
+  })) passed++; else failed++;
```
bri-tong, line 118: This test guards against re-introducing `grep -P`, which is valuable. However, it's purely source-level — it checks string patterns, not behavior.

Consider adding a behavioral assertion that actually runs the extraction against sample input, e.g.:

```js
if (test('extract_context7_key extracts the key value', () => {
  const tmp = path.join(os.tmpdir(), 'c7key-test.txt');
  fs.writeFileSync(tmp, '["--key", "abc-123"]\n');
  const { stdout } = require('child_process').spawnSync(
    'bash', ['-c', `source ${scriptPath} --dry-run 2>/dev/null; extract_context7_key ${tmp}`],
    { encoding: 'utf8' }
  );
  fs.unlinkSync(tmp);
  assert.strictEqual(stdout.trim(), 'abc-123');
})) passed++; else failed++;
```

This would catch regex bugs that source-level checks miss.

affaan-m:
Ported directly into main in dba5ae7 with targeted regression coverage. Closing as superseded-by-direct-port.

PR #1265, opened 2026-04-05
fix(hooks): treat HTTP 400 MCP probes as reachable

### LINKED ISSUE #1249: MCP health check probe falsely marks HTTP-based MCP servers as unhealthy (HTTP 400)
## Problem

The `mcp-health-check.js` hook sends a bare `GET` request to MCP server URLs as a health probe. Many MCP servers (e.g. GitHub MCP) only accept `POST` with proper MCP protocol payloads, so they respond with **HTTP 400** to a bare GET.

Since 400 is not in `HEALTHY_HTTP_CODES`, the hook marks the server as unhealthy and blocks all subsequent tool calls with exponential backoff (up to 10 minutes). The server is actually running fine — it just doesn't respond to GET.

## Impact

- GitHub MCP tools become completely blocked after the first probe failure
- Backoff escalates quickly (14 failures → minutes of blocking in my case)
- Users are silently forced to fall back to CLI alternatives with no clear indication of why

## Reproduction

1. Configure an HTTP-based MCP server (e.g. GitHub) that returns 400 on bare GET
2. Call any `mcp__github__*` tool
3. The PreToolUse hook probes the server, gets 400, blocks the call
4. Subsequent calls are blocked by backoff without even re-probing

## Relevant code

`scripts/hooks/mcp-health-check.js`:
```js
const HEALTHY_HTTP_CODES = new Set([200, 201, 202, 204, 301, 302, 303, 304, 307, 308, 405]);
```

The probe sends a GET request (`requestHttp` at line ~233) and checks against this set. HTTP 400 is not included.

## Suggested fix

A few options (not mutually exclusive):

1. **Add 400 to `HEALTHY_HTTP_CODES`** — a 400 means the server is reachable and responding; it's rejecting the request format, not indicating unavailability
2. **Use a more appropriate probe method** — e.g. send an `OPTIONS` or `HEAD` request, or a minimal MCP `initialize` message
3. **Distinguish "server unreachable" from "server rejected request"** — only ECONNREFUSED/timeout/5xx should mark unhealthy; 4xx (except 401/403/429) means the server is up

_— Filed via Claude Opus 4.6_

Gujiassh:
## Summary
- treat HTTP 400 responses from the MCP health probe as evidence that an HTTP-based MCP server is reachable rather than unhealthy
- prevent the PreToolUse hook from incorrectly blocking HTTP MCP tools behind exponential backoff when the server rejects a bare GET probe
- add a focused regression test that spins up an HTTP server returning 400 and asserts the hook keeps the MCP tool call healthy

Fixes #1249.

## Testing
- `node tests/hooks/mcp-health-check.test.js`
- `npm test`

---
## Summary by cubic
Treat HTTP 400 from the MCP HTTP health probe as a reachable response so HTTP-based MCP servers aren’t falsely marked unhealthy. This prevents PreToolUse backoff when servers reject a bare GET and adds a regression test to ensure 400 responses are marked healthy (fixes #1249).

<sup>Written for commit 08126370aa55d55e8c97badad41ebddd0ced92ab. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Improvements**
  * HTTP 400 responses from MCP servers are now treated as healthy. This prevents the system from incorrectly marking servers as unhealthy when they return a 400 status code.

* **Tests**
  * Added test coverage for health check behavior with HTTP 400 responses, including server status verification.

### DIFF
--- a/scripts/hooks/mcp-health-check.js
+++ b/scripts/hooks/mcp-health-check.js
@@ -24,7 +24,7 @@ const DEFAULT_TTL_MS = 2 * 60 * 1000;
 const DEFAULT_TIMEOUT_MS = 5000;
 const DEFAULT_BACKOFF_MS = 30 * 1000;
 const MAX_BACKOFF_MS = 10 * 60 * 1000;
-const HEALTHY_HTTP_CODES = new Set([200, 201, 202, 204, 301, 302, 303, 304, 307, 308, 405]);
+const HEALTHY_HTTP_CODES = new Set([200, 201, 202, 204, 301, 302, 303, 304, 307, 308, 400, 405]);
 const RECONNECT_STATUS_CODES = new Set([401, 403, 429, 503]);
 const FAILURE_PATTERNS = [
   { code: 401, pattern: /\b401\b|unauthori[sz]ed|auth(?:entication)?\s+(?:failed|expired|invalid)/i },
--- a/tests/hooks/mcp-health-check.test.js
+++ b/tests/hooks/mcp-health-check.test.js
@@ -6,9 +6,10 @@
 
 const assert = require('assert');
 const fs = require('fs');
+const http = require('http');
 const os = require('os');
 const path = require('path');
-const { spawnSync } = require('child_process');
+const { spawn, spawnSync } = require('child_process');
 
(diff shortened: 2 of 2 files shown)


affaan-m:
Ported directly to main in 8baffb4 after diff review. Closing this PR as superseded by the direct fix.

PR #1419, opened 2026-04-14
fix(gateguard): avoid disk write on every isChecked() call

### LINKED ISSUE #1406: bug: 4 issues found in recent commits and open PRs (gateguard perf, dashboard bare except, PR count conflict, JS→TS rules mismatch)
Found during code review of recent merged commits and open PRs. Four issues ranging from performance to correctness.

---

## 1. `gateguard-fact-force.js`: `isChecked()` writes to disk on every read (performance bug)

**File**: `scripts/hooks/gateguard-fact-force.js:97-101`  
**Merged in**: PR #1367

```js
function isChecked(key) {
  const state = loadState();
  const found = state.checked.includes(key);
  saveState(state);   // ← unnecessary write on every read
  return found;
}
```

`saveState()` is called purely to update `last_active`, but this triggers a full JSON serialise → temp-file write → `fs.rename` on **every** `isChecked()` call — which happens for every tool invocation. The project rules for blocking hooks explicitly say **"keep fast (<200ms) — no network calls"**. Excess synchronous disk writes on the hot path violate that spirit and add measurable latency.

**Fix**: Only call `saveState()` when the state is actually mutated (i.e., inside `markChecked()`). A separate heartbeat write in `markChecked()` already keeps `last_active` fresh enough for the 30-minute timeout.

---

## 2. `ecc_dashboard.py`: 8 bare `except:` clauses silently swallow all errors

**File**: `ecc_dashboard.py` (lines 97, 171, 265, 329, 883, 888, 892, 897)  
**Merged in**: PR #1377

Bare `except:` catches **everything**, including `SystemExit`, `KeyboardInterrupt`, and `MemoryError`. This makes the dashboard silently misbehave instead of surfacing real failures.

```python
# example at line 171 — hides all parse errors
try:
    with open(skill_file, 'r', encoding='utf-8') as f:
        ...
except:
    pass
```

**Fix**: Replace all bare `except:` with `except Exception:` (at minimum) or a specific exception type like `(OSError, ValueError)` where the intent is clear.

Additionally, `self.state('zoomed')` at line 260 is Windows-only — on Linux and macOS it silently has no effect. Use `self.attributes('-zoomed', True)` on Linux or `self.state('zoomed')` guarded by `sys.platform`.

---

## 3. PR #1397 and PR #1399 will merge-conflict on the same counter lines

**PRs**: #1397 (`/project-init`) and #1399 (`/auto-update`) — both open

Both PRs patch the **same lines** in the same files to bump the command count from 79 → 80:

| File | Line touched |
|------|-------------|
| `README.md` | "47 agents, 181 skills, and **79** legacy command shims" → 80 |
| `AGENTS.md` | "47 specialized agents, 181 skills, **79** commands" → 80 |
| `README.zh-CN.md` | Same Chinese equivalents |
| `docs/zh-CN/AGENTS.md` | Same |
| `docs/zh-CN/README.md` | Same |
| `agent.yaml` | Different lines (one adds `auto-update`, other adds `project-init`) — no conflict |

Additionally, PR #1396 (Swift agents, also open) bumps the *agent* count 47 → 49 on the same `AGENTS.md` line that #1397 and #1399 touch. Whoever merges second will get a conflict.

**Fix**: Merge the PRs in sequence and rebase/update each branch after the previous merges. Consider automating the counter update so contributors don't need to touch it manually.

---

## 4. `project-stack-mappings.json`: `javascript` stack maps to `typescript` rules

**File**: `config/project-stack-mappings.json` (added in PR #1397)

```json
{
  "id": "javascript",
  "name": "JavaScript (Node.js)",
  "rules": ["common", "typescript"],   // ← should be ["common"] or ["common", "javascript"]
  ...
}
```

A plain JavaScript/Node.js project should not be forced to load TypeScript rules (which include `tsc --noEmit` in lint commands and TypeScript-specific guidance). This will install `typescript`-specific rules into projects that have no TypeScript and generate confusing lint commands.

**Fix**: Use `["common"]` for the base JavaScript stack, or create a dedicated `rules/javascript/` rule file. The `typescript` stack entry (which already exists) correctly applies to TypeScript projects.

---

**Summary**

| # | Severity | File | Status |
|---|----------|------|--------|
| 1 | Medium — performance | `scripts/hooks/gateguard-fact-force.js:97-101` | Merged (#1367) |
| 2 | Medium — correctness | `ecc_dashboard.py` (8 locations) | Merged (#1377) |
| 3 | Low — merge friction | `README.md`, `AGENTS.md`, etc. | Open PRs #1397, #1399, #1396 |
| 4 | Low — correctness | `config/project-stack-mappings.json` | Open PR #1397 |

shenchangmin:
## Summary

Addresses issue 1 of 4 in #1406 — the gateguard PreToolUse hook was rewriting the entire state file on **every** tool invocation, purely to refresh \`last_active\`.

\`scripts/hooks/gateguard-fact-force.js\`: \`isChecked()\` is on the hot path (fires for every Edit/Write/Bash/MultiEdit), and each call was doing a full JSON serialise → temp-file write → \`fs.rename\`. This adds measurable synchronous I/O to the hook and violates the blocking-hook guideline in \`rules/\` (\`<200ms, no network\`).

## Change

- Removed the \`saveState(state)\` call from the read-only path in \`isChecked()\`.
- \`markChecked()\` already persists \`last_active\` whenever state actually changes (every new file gate, every new destructive-command gate, first bash per session), so the 30-minute inactivity timer still stays fresh for any active editing session.
- Added an inline comment explaining why the fast path is read-only.

## Test update

Test 12 previously asserted the buggy behaviour ("touches last_active on read so active sessions do not age out"). It has been rewritten to assert the correct behaviour: cached reads of already-checked entries must not rewrite the state file or touch \`last_active\`, verified via stat \`mtimeMs\` before/after plus a JSON diff against a seeded timestamp.

## Validation

- \`node tests/hooks/gateguard-fact-force.test.js\` — 13/13 passed
- \`node tests/run-all.js\` — 1826 passed (baseline on \`main\` is 1825; this PR adds one passing test, zero new failures)

Closes #1406 (issue 1 of 4).

---
## Summary by cubic
Stops `gateguard` from rewriting its state file on every `isChecked()` call, and adds a 60s heartbeat so cached reads only refresh `last_active` at most once per minute. This removes hot-path sync I/O while keeping long sessions alive.

- **Bug Fixes**
  - Read path is read-only by default; only persists when `last_active` is older than `READ_HEARTBEAT_MS` (60s).
  - Kept `last_active` updates in `markChecked()` when state actually changes.
  - Split Test 12 into 12a (no rewrite within heartbeat) and 12b (refresh after heartbeat) with `mtimeMs` and JSON assertions.

<sup>Written for commit ea3ec68709c8760664884c4af413d7d62ba98541. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary disk writes during read-only checks by adding a throttled heartbeat so session state is only refreshed when stale.
  * Preserved existing checked entries when heartbeat-triggered updates occur.

* **Tests**
  * Split and updated tests to assert reads within the heartbeat do not modify persisted state and reads after the heartbeat refresh last-active timestamps.

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -95,10 +95,12 @@ function markChecked(key) {
 }
 
 function isChecked(key) {
+  // Read-only fast path: the hook fires on every tool invocation, so any
+  // disk write here shows up as measurable latency. markChecked() already
+  // persists last_active whenever state actually changes, which keeps the
+  // 30-minute inactivity timer fresh for any active editing session.
   const state = loadState();
-  const found = state.checked.includes(key);
-  saveState(state);
-  return found;
+  return state.checked.includes(key);
 }
 
 // Prune stale session files older than 1 hour
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -35,6 +35,11 @@ const STATE_FILE = path.join(STATE_DIR, `state-${SESSION_ID.replace(/[^a-zA-Z0-9
 // State expires after 30 minutes of inactivity
 const SESSION_TIMEOUT_MS = 30 * 60 * 1000;
 
+// Throttle keep-alive writes on the read path: at most one write per 60s.
+// Protects the hot path (fires on every tool call) while still refreshing
+// last_active often enough that long single-file sessions don't expire.
+const READ_HEARTBEAT_MS = 60 * 1000;
+
 // Maximum checked entries to prevent unbounded growth
 const MAX_CHECKED_ENTRIES = 500;
 const MAX_SESSION_KEYS = 50;
@@ -95,12 +100,16 @@ function markChecked(key) {
 }
 
 function isChecked(key) {
-  // Read-only fast path: the hook fires on every tool invocation, so any
-  // disk write here shows up as measurable latency. markChecked() already
-  // persists last_active whenever state actually changes, which keeps the
-  // 30-minute inactivity timer fresh for any active editing session.
+  // Read path fires on every tool invocation, so default to no disk write.
+  // We only heartbeat last_active at most once per READ_HEARTBEAT_MS so that
+  // long single-file sessions don't hit the inactivity timeout, while still
+  // keeping ~99% of invocations fully read-only.
   const state = loadState();
-  return state.checked.includes(key);
+  const found = state.checked.includes(key);
+  if (found && Date.now() - (state.last_active || 0) > READ_HEARTBEAT_MS) {
+    saveState(state);
+  }
+  return found;
 }
 
 // Prune stale session files older than 1 hour
--- a/tests/hooks/gateguard-fact-force.test.js
+++ b/tests/hooks/gateguard-fact-force.test.js
@@ -363,10 +363,11 @@ function runTests() {
     }
   })) passed++; else failed++;
 
-  // --- Test 12: reads of already-checked files do not rewrite state (perf) ---
+  // --- Test 12a: cached reads inside the heartbeat window stay read-only ---
   clearState();
-  if (test('does not rewrite state file when reading already-checked entries', () => {
-    const seededActive = Date.now() - (5 * 60 * 1000);
+  if (test('does not rewrite state file on cached read within heartbeat window', () => {
+    // Seeded 10s ago — well inside the 60s heartbeat window.
+    const seededActive = Date.now() - (10 * 1000);
     writeState({
       checked: ['/src/keep-alive.js'],
       last_active: seededActive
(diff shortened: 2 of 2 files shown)


shenchangmin:
Thanks @cubic-dev-ai and Greptile for flagging — this trade-off was intentional, and I agree it's worth surfacing.

The edge case is a session that keeps editing the *same* already-gated files for 30+ minutes without touching any new file or new destructive command (both of those still trigger `markChecked` → `saveState`). In that scenario the inactivity timer does expire on the old code path and the user would hit a fresh gate — a minor re-prompt, not lost work.

I picked the pure-removal fix because the hot-path write fires on every tool invocation and most editing sessions touch multiple files within 30 minutes. But if you'd prefer to keep the keep-alive semantic, I'm happy to push a follow-up commit with a throttled heartbeat — roughly:

```js
function isChecked(key) {
  const state = loadState();
  const found = state.checked.includes(key);
  // Throttled keep-alive: at most one write per 60s on the read path
  if (found && Date.now() - (state.last_active || 0) > 60_000) {
    saveState(state);
  }
  return found;
}
```

That would cut disk writes by ~99% while preserving the inactivity-timer behaviour. Let me know which direction you prefer.

shenchangmin:
Pushed the throttled heartbeat approach in `ea3ec68`.

**Behaviour:**
- `isChecked()` on a cached hit now only calls `saveState` when `last_active` is older than 60s (`READ_HEARTBEAT_MS`).
- Hot path stays read-only for ~99% of invocations; long single-file sessions still refresh inactivity well before the 30-minute timeout.

**Test coverage:**
- Test 12a — cached read within the heartbeat window: `mtime` and `last_active` unchanged.
- Test 12b — cached read past the heartbeat window: `last_active` advances, `checked` entries preserved.

Local results: gateguard suite 14/14 passing, full `node tests/run-all.js` 1826 passed (identical failure count to `main`). Ready for another pass whenever you have cycles.

shenchangmin:
Good catch @cubic-dev-ai — that race is real, but I want to flag a couple of things before acting on it:

**The race is pre-existing, not introduced by this PR.** On `main` today, `isChecked()` already does `loadState()` → `saveState(state)` on every invocation with no locking, and `markChecked()` has the same load-modify-save shape. So any concurrent hook process (e.g., two parallel tool calls sharing a `CLAUDE_SESSION_ID`) can already lose entries against the current implementation.

**This PR actually narrows the window** rather than widening it. Before: every tool call was a potential TOCTOU overwrite (100% of invocations). After this PR: the write only fires once per ~60s per session on cached reads, so the same race exists in maybe ~1–2% of the original window. A `markChecked()` during that gap is still vulnerable, but every read outside the gap is now fully read-only and can no longer overwrite anything.

**The proper fix is a state-layer change, not a one-line `isChecked()` tweak.** To close the race fully we need one of:
- An `flock(STATE_FILE, ex)` guard around the whole read-modify-write (portable via `fs.openSync` + `proper-lockfile` or a shell `flock`).
- Merge-style saves: on every write, re-read the file, union `checked` entries, take the max `last_active`, then write.
- In-memory coordination via a short-lived lease file (harder cross-platform).

I'd rather keep this PR focused on the performance fix and open a follow-up issue for the concurrency model so it gets reviewed on its own merits. Happy to draft that follow-up if it sounds right — WDYT?

affaan-m:
Superseded by #1439, which carries the GateGuard hot-path read/write reduction on top of current main together with the related urgent install/name changes.

affaan-m:

PR #1431, opened 2026-04-14
fix(continuous-learning-v2): relocate observer data out of ~/.claude to fix instinct writes

### LINKED ISSUE #1599: Observer cannot persist instincts: ~/.claude/** writes are blocked in --print mode
## Summary

The continuous-learning-v2 observer runs a Haiku subprocess via `claude --model haiku --print` that is supposed to write instinct files to `~/.claude/homunculus/projects/<id>/instincts/personal/`. Writes silently fail. Analysis runs, but no instincts ever persist.

## Root cause

Claude Code has a hardcoded sensitive-path guard on `~/.claude/**` that blocks `Write` tool calls in `--print` mode regardless of:

- `--permission-mode acceptEdits`
- `--permission-mode bypassPermissions`
- `--dangerously-skip-permissions`
- `permissions.allow` entries in `settings.json` (these only downgrade the hard block to an interactive prompt, which never resolves in non-interactive subprocesses)

Empirically verified across all four variants on Claude Code 2.1.x. The block is path-based, not mode-based.

## Repro

1. Install ECC + continuous-learning-v2.
2. Generate 10+ observations matching a recurring pattern.
3. Observer log shows analysis ran.
4. Check `~/.claude/homunculus/projects/<id>/instincts/personal/`.

Observed: directory empty.
Expected: at least one instinct `.md` file written.

## Fix

Move observer data out of `~/.claude/**` into the XDG-spec-standard location (`${XDG_DATA_HOME:-$HOME/.local/share}/ecc-homunculus/`) with `CLV2_HOMUNCULUS_DIR` env override. Existing users run a one-shot migration script. PR: #1431

AdamiecRadek:
Companion fix to #1417 (handles SIGUSR1 in observer wait loop).

## Summary

Observer homunculus data moves from `~/.claude/homunculus/` to `${XDG_DATA_HOME:-$HOME/.local/share}/ecc-homunculus/`, with `CLV2_HOMUNCULUS_DIR` as an explicit override. This unblocks instinct writes that were silently failing under Claude Code's `--print`-mode sensitive-path guard.

Fixes #1599

## Type of Change
- [x] `fix:` Bug fix

## Testing Done

`tests/hooks/hooks.test.js` + `tests/integration/hooks.test.js` cover:
- Env chain for `detect-project.sh` (override / XDG / default)
- Same chain for `instinct-cli.py`
- Migration script: happy path, idempotency, conflict refusal, empty-destination merge, observer-running refusal
- Pre-existing registry/session/lease tests updated to the new default path

Full suite: 1838 passing. Remaining 5 failures are pre-existing upstream (`ci/validators.test.js`, `hooks/mcp-health-check.test.js`) and reproduce on `origin/main` untouched.

## Why

Claude Code has a hardcoded sensitive-path guard on `~/.claude/**` that ignores every `--permission-mode` variant (`acceptEdits`, `bypassPermissions`, `--dangerously-skip-permissions`) in `--print` mode. `permissions.allow` entries only downgrade the hard block to an interactive prompt, which never resolves in non-interactive `--print` subprocesses. The observer's Haiku child could not `Write` instincts back; analysis ran, nothing persisted.

Verified empirically with isolated Haiku probes against all four permission-mode variants — the block is path-based, not mode-based. Moving the data is the only fix available from inside the skill.

## Path resolution

```
CLV2_HOMUNCULUS_DIR              # explicit override (must be absolute)
  then XDG_DATA_HOME/ecc-homunculus
  then $HOME/.local/share/ecc-homunculus
```

Mirrored in bash (`detect-project.sh`, `observe.sh`, `start-observer.sh`), Python (`instinct-cli.py`), and Node (`scripts/lib/observer-sessions.js`, consumed by `session-start.js`).

## Migration

New installs use the new path directly — no migration needed.

Existing users with data at `~/.claude/homunculus` run the migration script once:

```bash
bash skills/continuous-learning-v2/scripts/migrate-homunculus.sh
```

Safety properties:
- Exits cleanly if old dir is absent (idempotent)
- Refuses while any `observer-loop.sh` under `$HOME` is running
- Refuses if both old and new exist with content (operator resolves manually)
- Does not touch `~/.claude/settings.json`; prints advisory if `CLV2_CONFIG` still points at the old path

## Code changes

- `skills/continuous-learning-v2/scripts/detect-project.sh`
- `skills/continuous-learning-v2/scripts/instinct-cli.py`
- `skills/continuous-learning-v2/hooks/observe.sh`
- `skills/continuous-learning-v2/agents/start-observer.sh`
- `scripts/lib/observer-sessions.js` (`getHomunculusDir` now uses the resolver; also exported for reuse)
- `scripts/hooks/session-start.js` (uses the exported helper instead of recomputing)
- New: `skills/continuous-learning-v2/scripts/lib/homunculus-dir.sh` (shared resolver)
- New: `skills/continuous-learning-v2/scripts/migrate-homunculus.sh`

Docs: `SKILL.md` and `agents/observer.md` updated for the new path. `CHANGELOG.md` describes the fix and migration step.

- [x] Shell scripts pass shellcheck (modified files)

## Documentation
- [x] Updated relevant documentation

## Related

Follows up on #1417, which was amended to carry only the SIGUSR1/`wait` fix once I confirmed that `--permission-mode acceptEdits` does not unblock instinct writes.


AdamiecRadek:
Pushed fixes for the review comments (commits 0d9a54c, db921ee, 90bd2b0, c52f280).

**Resolver consolidation + absolute-path validation** (0d9a54c, db921ee, 90bd2b0)
- New shared `skills/continuous-learning-v2/scripts/lib/homunculus-dir.sh` sourced by `detect-project.sh`, `observe.sh`, `start-observer.sh`, `migrate-homunculus.sh` — kills the 4× inline duplication.
- Shell + Node: relative `CLV2_HOMUNCULUS_DIR` / `XDG_DATA_HOME` now warn on stderr and fall through to default. Python: relative `CLV2_HOMUNCULUS_DIR` raises `ValueError` (fail-fast on our own var), `XDG_DATA_HOME` warns + falls back (per XDG spec).
- `observer-sessions.js`: hoisted `require('os')` to module header.

**migrate-homunculus.sh hardening** (0d9a54c)
- Early exit when resolved destination equals source (the P1 NEW==OLD case).
- Warn on stderr when `pgrep` is absent instead of silently skipping the guard.
- Kept the `${HOME}`-anchored `pgrep` pattern; broadening to bare `observer-loop\\.sh` caused false positives in shared test envs. Open to a PID-file-based replacement as a follow-up — happy to do that as a separate PR if preferred.

**Docs** (c52f280)
- `SKILL.md:135`: show `CLV2_HOMUNCULUS_DIR` in the `projects.json` path.
- `SKILL.md:372`: full `skills/continuous-learning-v2/scripts/migrate-homunculus.sh` path.

**Tests** (c52f280)
- Runtime `findPython()` replaces hardcoded `python3` in the three `instinct-cli.py` invocations so Windows / minimal-image runners don't trip.
- `waitForPgrep()` polls up to 3s before asserting `migrate-homunculus.sh` refuses, fixing the slow-CI race.
- Added regression tests for each new validation branch (relative-path fallback in shell + Node + Python).

Full suite: 1844 passing, 5 failing — all 5 are pre-existing failures unrelated to this PR (MCP-health and catalog validators).

affaan-m:
Thanks for the homunculus storage relocation work. I ported the XDG/override resolver, migration script, docs, and matching tests onto current `main` in #1718, so I’m closing this as superseded.

PR #1439, opened 2026-04-15
fix: unblock urgent install and gateguard regressions

affaan-m:
## Summary
- fix urgent GateGuard session-state breakage and Windows-safe state writes
- stop GateGuard from blocking `.claude/settings.json` and read-only git introspection
- detect Claude marketplace installs in harness audit
- make native `.cursor/rules` win over legacy `rules/` collisions
- remove stale `ecc@ecc` plugin install/docs references and lock them down in tests

## Why this PR exists
These are live install/name/GateGuard regressions that are currently split across stale or partial branches. This consolidates the highest-signal fixes into one branch against current `main`.

## Issue coverage
- addresses #1435
- addresses #1427
- addresses #1405
- addresses #1437

## Supersedes / folds in
- #1408
- #1409
- #1411
- #1419
- #1432

## Validation
- `node tests/hooks/gateguard-fact-force.test.js`
- `node tests/scripts/harness-audit.test.js`
- `node tests/lib/install-targets.test.js`
- `node tests/lib/install-manifests.test.js`
- `node tests/plugin-manifest.test.js`

---
## Summary by cubic
Fixes GateGuard session-state and command gating regressions, and unblocks installs by detecting marketplace plugin paths in both project and home (with `HOME`/`USERPROFILE`/OS fallbacks) while preferring native Cursor rules. Standardizes plugin naming to `everything-claude-code`, updates docs/tests (including URL-based marketplace adds), syncs catalog counts (48 agents, 183 skills), and restores CI.

- **Bug Fixes**
  - GateGuard state: stable per-session keys from raw input/env/transcript/project; hashes long IDs; atomic writes with Windows-safe rename; read heartbeat to avoid hot-path rewrites; bounded history and safe cleanup.
  - GateGuard allowances: allow edits to `.claude/settings*.json` and read-only Git introspection (`git status/diff/log/show/branch --show-current/rev-parse --abbrev-ref HEAD`); continue gating destructive commands.
  - Install audit: detect marketplace installs under `.claude/plugins/marketplaces` in both project and home directories, honoring `HOME`/`USERPROFILE`/`os.homedir()` fallbacks.
  - Install targets: prefer native `.cursor/rules` over `rules/*` collisions for Cursor.
  - Docs/tests: replace `ecc@ecc` with `everything-claude-code`, require full GitHub URL for `/plugin marketplace add`, add install-name note, and enforce via tests.

[...]

PR #1442, opened 2026-04-15
Fix: bug(gateguard): race condition in session state read-modify-write path

### LINKED ISSUE #1441: bug(gateguard): race condition in session state read-modify-write path
Follow-up to #1406, #1419, #1439. Splitting the state-layer concurrency concern into its own issue so it can be designed and reviewed independently of the hot-path performance fix that already landed via #1439.

## Problem

`scripts/hooks/gateguard-fact-force.js` persists session state (`checked` array + `last_active`) via a classic load-modify-save pattern with no mutual exclusion:

```js
function markChecked(key) {
  const state = loadState();                  // 1. read
  if (!state.checked.includes(key)) {
    state.checked.push(key);                  // 2. modify
    saveState(state);                         // 3. atomic file rename
  }
}
```

`saveState()` is atomic at the filesystem level (temp file + `fs.renameSync`), so we never see a *corrupted* state file. But the read-modify-write window itself is unprotected, and the hook runs as a fresh Node process per tool invocation — so there is no shared memory between concurrent hook processes to coordinate on.

## Concurrent scenario

Two tool calls that share a `CLAUDE_SESSION_ID` fire nearly simultaneously (parallel tool use, or rapid-fire edits on the same session):

1. Process A: `loadState()` → `{checked: [a], last_active: t0}`
2. Process B: `loadState()` → `{checked: [a], last_active: t0}`   (same snapshot)
3. Process A: `markChecked(b)` → writes `{checked: [a, b], last_active: t1}`
4. Process B: `markChecked(c)` → writes `{checked: [a, c], last_active: t2}`   (b is gone)

User-visible effect: the file `b` that process A just gated appears un-checked to the next invocation, so Claude gets re-prompted with the Fact-Forcing gate for a file it already investigated. Not data loss in the strict sense, but a UX regression proportional to how often parallel tool calls land on the same session.

The `READ_HEARTBEAT_MS` throttle landed in #1439 narrows the read-path window (~99% of invocations are now pure reads), but the `markChecked()` path still exhibits the full race on every new gate, and the throttled heartbeat write in `isChecked()` still reproduces it when it does fire.

## Candidate fixes

Three options, roughly ordered by engineering cost / platform portability trade-off:

### 1. `flock`-style advisory lock around read-modify-write

Wrap the entire `loadState` → mutate → `saveState` sequence in an exclusive lock on `STATE_FILE` (or a sibling `.lock` file). Portable options:

- [`proper-lockfile`](https://github.com/moxystudio/node-proper-lockfile) — cross-platform, already handles stale locks, small footprint.
- Raw `fs.openSync(path, 'wx')` sentinel file with retry — zero dependency, but requires reinventing stale-lock handling.
- Shell out to `flock` on POSIX + `powershell Wait-File` on Windows — brittle, not recommended.

**Pros:** closes the race fully. **Cons:** new dependency or stale-lock logic; slight latency on every mutation.

### 2. Merge-style saves

Before `fs.renameSync`, re-read the current on-disk state one last time and *merge* it with the in-memory state: union of `checked`, `max(last_active)`. Roughly:

```js
function saveState(state) {
  const disk = loadStateRaw();                // non-throwing
  state.checked = Array.from(new Set([...(disk.checked || []), ...state.checked]));
  state.last_active = Math.max(disk.last_active || 0, state.last_active);
  // ... existing prune + atomic write
}
```

**Pros:** zero new dependencies, no lock contention, matches the "additive-only state" semantics this code already has. **Cons:** still not strictly serializable (two merges can race), but entries are never *lost* — just occasionally duplicated before the next prune, which is fine.

### 3. In-memory coordination via short-lived lease

Maintain a per-session lease file (e.g. `state-<sid>.lease`) that hook processes claim for, say, 100ms before mutating. Simpler than flock but less robust than merge-style saves on crash/ctrl-C.

**Pros:** cheap. **Cons:** needs correct cleanup; behaves worst-case the same as option 2 under crashes.

## Recommendation

Option 2 (merge-style saves) feels like the right default for ECC's profile: no new dependency, tolerant of crashes, preserves the "additive entries + bounded pruning" semantics already in place. If/when GateGuard needs stronger invariants (e.g. session-level transactions across multiple files), revisit with option 1.

## Severity

P2 — UX regression only, no data loss. Easy to reproduce under parallel tool calls but rare in typical interactive sessions.

## Links

- #1406 — original multi-issue report
- #1419 — hot-path read/write removal (my initial PR, superseded)
- #1439 — GateGuard consolidation that landed the read-heartbeat throttle
- Prior discussion: https://github.com/affaan-m/everything-claude-code/pull/1419#issuecomment-4241045692 and https://github.com/affaan-m/everything-claude-code/pull/1439#issuecomment-4248766604

shahyashish:
The fix addresses a race condition in the session state read-modify-write path by implementing a 'merge-style save' strategy. Previously, `saveState` would simply overwrite the session file with the current in-memory snapshot, which could lead to data loss if two concurrent hook processes were running for the same session. Now, `saveState` performs a raw read of the current on-disk state immediately before writing, merges it with the in-memory state (creating a union of 'checked' keys and taking the maximum 'last_active' timestamp), and then performs an atomic write via a temp file. This ensures that checks performed by parallel processes are additive rather than destructive. A random suffix was also added to the temp file path to further reduce collision risks.

Test: 1. Set `CLAUDE_SESSION_ID=test_race`. 2. Run two instances of the hook script nearly simultaneously, each attempting to `markChecked` a different file (e.g., `file_A` and `file_B`). 3. Verify the resulting session JSON file in the state directory (`.gateguard/state-test_race.json`) contains both `file_A` and `file_B` in its `checked` array, regardless of which process finished first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data persistence reliability by enhancing state merging and conflict resolution logic.
  * Strengthened error handling to prevent system blocking during data save operations.
  * Enhanced collision resistance for temporary file generation.

---
## Summary by cubic
Fixes a race condition in GateGuard session saves by merging on-disk and in-memory state before an atomic write, preventing lost `checked` entries during parallel runs. Fixes #1441.

- **Bug Fixes**
  - `saveState` now does a merge-style save: unions `checked` and takes the max `last_active` with the on-disk state.
  - Uses an atomic write via a temp file with PID + random suffix to avoid collisions.

<sup>Written for commit 4cf12e81b8c161fcbc366b425f786c3f6484d2f3. Summary will update on new commits.</sup>

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -76,14 +76,37 @@ function pruneCheckedEntries(checked) {
 
 function saveState(state) {
   try {
-    state.last_active = Date.now();
-    state.checked = pruneCheckedEntries(state.checked);
     fs.mkdirSync(STATE_DIR, { recursive: true });
-    // Atomic write: temp file + rename prevents partial reads
-    const tmpFile = STATE_FILE + '.tmp.' + process.pid;
-    fs.writeFileSync(tmpFile, JSON.stringify(state, null, 2), 'utf8');
+
+    // 1. Prepare base state from what's currently in memory
+    let mergedChecked = state.checked || [];
+    let mergedLastActive = state.last_active || 0;
+
+    // 2. Merge with current disk state to prevent overwriting concurrent updates (Option 2: Merge-style saves)
+    try {
+      if (fs.existsSync(STATE_FILE)) {
+        const disk = JSON.parse(fs.readFileSync(STATE_FILE, 'utf8'));
+        if (Array.isArray(disk.checked)) {
+          // Union of checked items, preserving order (disk items first, then memory items)
+          mergedChecked = Array.from(new Set([...disk.checked, ...mergedChecked]));
+        }
+        if (typeof disk.last_active === 'number') {
+          mergedLastActive = Math.max(mergedLastActive, disk.last_active);
+        }
+      }
+    } catch (_) { /* ignore disk read errors, proceed with memory state */ }
+
+    // 3. Finalize state: update heartbeat and prune to bounded size
+    const finalState = {
+      checked: pruneCheckedEntries(mergedChecked),
+      last_active: Math.max(mergedLastActive, Date.now())
+    };
+
+    // 4. Atomic write: temp file + rename prevents partial reads
+    const tmpFile = STATE_FILE + '.tmp.' + process.pid + '.' + Math.random().toString(36).slice(2, 6);
+    fs.writeFileSync(tmpFile, JSON.stringify(finalState, null, 2), 'utf8');
     fs.renameSync(tmpFile, STATE_FILE);
-  } catch (_) { /* ignore */ }
+  } catch (_) { /* ignore all errors to ensure hook never blocks */ }
 }
 
 function markChecked(key) {


shenchangmin:
@shahyashish — thanks for picking up #1441 so quickly. Heads up per #1453: the merge-style rewrite in your new \`saveState\` dropped the Windows \`EEXIST\` / \`EPERM\` unlink-before-rename branch that currently lives in \`main\` (landed as part of #1439's GateGuard consolidation). \`fs.renameSync\` throws \`EEXIST\` on Windows when the target already exists, so without the unlink fallback the second write onwards silently no-ops inside the catch-all and GateGuard stops seeing checked files.

The branch to preserve:

\`\`\`js
try {
  fs.renameSync(tmpFile, STATE_FILE);
} catch (error) {
  if (error && (error.code === 'EEXIST' || error.code === 'EPERM')) {
    try { fs.unlinkSync(STATE_FILE); } catch (_) { /* ignore */ }
    fs.renameSync(tmpFile, STATE_FILE);
  } else {
    throw error;
  }
}
\`\`\`

FWIW #1450 hit a parallel shape on the Python side — three Windows-specific pitfalls in a row (launcher-as-cmd-builtin → cmd.exe re-parsing → argv quoting) before landing on \`explorer.exe\`. Cross-platform edge cases tend to come in layers; might be worth a full Windows smoke pass on the merged state before ship.

affaan-m:
Thanks for the original GateGuard race-condition fix. I ported the concurrent state-write protection onto current `main` in #1623 so we could absorb the rebase burden on the maintainer side. #1623 has now landed, so I am closing this PR as superseded; credit is preserved here for the original fix direction.

PR #1444, opened 2026-04-15
Fix: fix(hooks): GateGuard permanently blocks all operations on Windows — fs.renameSync EEXIST not handled

shahyashish:
Fixed two critical issues on Windows: 1) Handled EEXIST error in fs.renameSync by unlinking the destination file before renaming, ensuring state is correctly persisted across atomic writes. 2) Replaced the unstable pid/ppid fallback for SESSION_ID with a stable hash of the current working directory. This ensures that hook invocations within the same project context share a consistent state file even when spawned through process wrappers that change parent PIDs.

Test: On a Windows environment without CLAUDE_SESSION_ID set, trigger a GateGuard fact-forcing check (e.g., by running a Bash command). Verify that the command is blocked initially. Provide the requested facts/instruction quote and retry the same command. Verify that the command is now permitted and that only a single state file (matching the stable CWD hash) is created in the .gateguard directory.

---
## Summary by cubic
Fixes GateGuard blocking on Windows by safely replacing the state file and using a stable session key. This prevents EEXIST errors during atomic writes and keeps session state consistent across process wrappers.

- **Bug Fixes**
  - Unlink destination before `fs.renameSync` to handle Windows EEXIST and ensure atomic state replacement.
  - Replace PID-based fallback with a stable CWD hash for `SESSION_ID` when `CLAUDE_SESSION_ID` is unset, so hooks in the same project share one state file.

<sup>Written for commit 9a523996300de4efb6387ab67a4bc19d4a0dbd42. Summary will update on new commits.</sup>

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -27,9 +27,9 @@ const fs = require('fs');
 const path = require('path');
 
 // Session state — scoped per session to avoid cross-session races.
-// Uses CLAUDE_SESSION_ID (set by Claude Code) or falls back to PID-based isolation.
+// Uses CLAUDE_SESSION_ID (set by Claude Code) or falls back to stable CWD-based isolation.
 const STATE_DIR = process.env.GATEGUARD_STATE_DIR || path.join(process.env.HOME || process.env.USERPROFILE || '/tmp', '.gateguard');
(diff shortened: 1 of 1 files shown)


eljosuesilva:
Me too

affaan-m:
Thanks for the Windows GateGuard blocker fix. Current `main` already has the stable session-key fallback and Windows `EEXIST`/`EPERM` rename fallback from this fix direction, and #1623 preserved that path while adding the remaining concurrent state-write protection. #1623 has now landed, so I am closing this PR as superseded; credit is preserved here for the original Windows unblock.

PR #1420, opened 2026-04-14
fix(dashboard): replace 8 bare except with except Exception

shenchangmin:
## Summary

Addresses issue 2 of 4 in #1406 — \`ecc_dashboard.py\` had 8 bare \`except:\` clauses that silently swallow every exception, including \`SystemExit\`, \`KeyboardInterrupt\`, and \`MemoryError\`. This makes the dashboard misbehave in place of surfacing real failures and prevents clean \`Ctrl+C\` shutdown.

## Change

Replaced all 8 bare-except clauses at the following lines with \`except Exception:\`:

| Line (pre-patch) | Context |
|---|---|
| 97 | Skill description parsing |
| 171 | Command description parsing |
| 265 | Taskbar icon load |
| 329 | Header logo load |
| 883 | Widget background recolor (root) |
| 888 | Widget background recolor (children) |
| 892 | Recursive color update |
| 897 | Top-level recolor |

\`except Exception:\` still hides the expected transient errors (missing icon asset, tkinter background-color failures on exotic widgets, malformed skill/command metadata files) without masking interpreter-level signals.

This is the minimum fix recommended in #1406. The separate platform issue on \`self.state('zoomed')\` at line 260 (Linux/macOS no-op) is intentionally left for a follow-up so this PR stays narrow.

## Validation

- \`python3 -m py_compile ecc_dashboard.py\` — OK
- \`grep -c "^\s*except:$" ecc_dashboard.py\` — 0 (was 8)
- \`node tests/run-all.js\` — 1826 passed, 6 failed (identical to baseline on \`main\`; Python GUI file is not exercised by the JS suite)

Closes #1406 (issue 2 of 4).

---
## Summary by cubic
Replaced 8 bare `except:` clauses in `ecc_dashboard.py` with `except Exception:` to stop swallowing interpreter signals and restore clean Ctrl+C while keeping transient UI/asset errors quiet. Addresses #1406 (item 2/4).

- **Bug Fixes**
  - Updated exception handling in skill/command metadata parsing, icon/logo loads, and theme recolor recursion.
  - Prevents hiding `SystemExit`/`KeyboardInterrupt`; expected UI/asset errors remain ignored.
  - Does not change the `self.state('zoomed')` behavior noted in #1406.

<sup>Written for commit 649f8094607d95b8f9bafd0624f79accfd17e9e8. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Chores**
  * Enhanced exception handling practices in internal code to improve robustness and maintainability.

PR #1450, opened 2026-04-15
fix(dashboard): reliability pass — narrow excepts, log silent failures, unify file launchers

### LINKED ISSUE #1406: bug: 4 issues found in recent commits and open PRs (gateguard perf, dashboard bare except, PR count conflict, JS→TS rules mismatch)

### LINKED ISSUE #1424: bug(dashboard): command injection on Linux + cross-platform crash in ecc_dashboard.py
Two confirmed bugs in `ecc_dashboard.py` (merged via #1377). Both affect the `open_terminal()` method and the `__init__` window-state call.

---

## Bug 1 — Shell command injection via unsanitized path (Linux) · **Security**

**File:** `ecc_dashboard.py:799`

```python
def open_terminal(self):
    path = self.path_entry.get()          # ← user-editable text field, no sanitisation
    ...
    else:  # Linux
        subprocess.Popen(['x-terminal-emulator', '-e', f'cd {path}'])
```

`x-terminal-emulator -e <cmd>` passes `<cmd>` to a shell. If `path` contains shell metacharacters — e.g. a project path with a space, a backtick, or a semicolon — the terminal will execute them. For example:

```
path = "/home/user/proj; rm -rf ~"
→ x-terminal-emulator -e "cd /home/user/proj; rm -rf ~"
```

The user controls the entry field directly (Browse button + manual editing), so this is a real injection surface even in a local app.

**Fix:** Pass the `cd` command as a properly quoted list argument, or use `shlex.quote`:

```python
import shlex
subprocess.Popen(['x-terminal-emulator', '-e', f'cd {shlex.quote(path)}'])
```

Or, better, use the terminal's `-–working-directory` flag where available instead of a shell `-e` argument.

---

## Bug 2 — `self.state('zoomed')` raises `TclError` on macOS and Linux · **Correctness**

**File:** `ecc_dashboard.py:260`

```python
self.state('zoomed')   # ← Windows-only Tk state
```

`'zoomed'` is a valid window state on Windows. On macOS and most Linux desktop environments it raises:

```
_tkinter.TclError: bad argument "zoomed": must be normal, iconic, or withdrawn
```

This crashes the dashboard on startup on non-Windows systems.

**Fix:**

```python
try:
    self.state('zoomed')          # Windows
except tk.TclError:
    self.attributes('-zoomed', True)  # Linux (most WMs)
```

macOS maximise requires `self.attributes('-fullscreen', True)` or can be skipped — the window is still usable at its default size.

---

## Related open PRs worth merging alongside a fix

| PR | Summary |
|---|---|
| #1420 | Replaces 8 bare `except:` with `except Exception` — directly related to this file |
| #1419 | Reduces unnecessary disk writes in `gateguard-fact-force.js` |
| #1411 | Stable session fallback for GateGuard in API/proxy setups |

shenchangmin:
## Summary

A focused reliability pass on \`ecc_dashboard.py\` and \`scripts/lib/ecc_dashboard_runtime.py\`, in the same module surface that already grew test coverage in #1439. Three related cleanups in one commit:

1. **Narrow exception handling** — replace the eight remaining bare \`except:\` clauses with \`except Exception:\` so \`SystemExit\` / \`KeyboardInterrupt\` / \`MemoryError\` propagate cleanly. Each block now logs context via a module-level logger.
2. **Logging policy** — the recursive widget recolor walker uses \`logger.debug(..., exc_info=True)\` instead of \`logger.exception\` because tkinter rejects the \`background\` option on many widget classes (menus, system frames, notebook tabs); logging at INFO/ERROR there would flood the default stream on every theme apply.
3. **Unify file launchers** — add \`build_file_open_launch(path, *, os_name=None)\` to the runtime helper, mirroring the \`build_terminal_launch\` contract introduced by #1439. Route \`open_readme\` and \`open_agents\` through it so every external launch the dashboard performs goes through the same testable, injection-safe argv-construction path. Removes the duplicate inline \`import subprocess\` in those methods.

## Why this PR exists

\`ecc_dashboard.py\` had eight \`except:\` clauses that silently swallow every exception, including signal-level ones (\`SystemExit\`, \`KeyboardInterrupt\`, \`MemoryError\`). \`#1406\` flagged this. At the same time, \`#1424\` flagged \`open_terminal\` command injection and the \`self.state('zoomed')\` cross-platform crash; both of those were already addressed in \`#1439\`'s new \`ecc_dashboard_runtime.py\` helpers (\`maximize_window\`, \`build_terminal_launch\`), but \`open_readme\` and \`open_agents\` were left calling \`subprocess.Popen\` directly with the same argv-construction style. Extending the helper to cover all three launch sites finishes the job and lets us close \`#1424\` with a coherent codebase rather than a partially-applied fix.

## Tests

- Cover \`build_file_open_launch\` in \`tests/scripts/ecc-dashboard.test.js\` with POSIX and Windows cases. Both assert that paths containing shell metacharacters (\`; rm -rf ~\`, \`& del C:\\*\`) survive as literal argv entries — the same threat-model property the existing \`build_terminal_launch\` tests verify.

## Validation

- \`python3 -m py_compile ecc_dashboard.py scripts/lib/ecc_dashboard_runtime.py\` — OK
- \`node tests/scripts/ecc-dashboard.test.js\` — 5/5 passed (3 existing + 2 new)
- \`node tests/run-all.js\` — 1856 passed (baseline on \`main\` is 1854; +2 new passing tests, **zero new failures**)

Closes #1406 (issue 2 of 4 — dashboard bare except).
Closes #1424 (open_terminal command injection + zoomed crash were addressed in #1439's runtime helpers; this PR closes the issue by extending the same pattern to the remaining file-launch surface).

Supersedes #1420 (the original bare-except-only PR — closing it in favour of this consolidated reliability pass).

## Summary by CodeRabbit

* **Bug Fixes**
  * Replaced silent failure paths with logged exception handling to surface errors during asset loading, parsing, and theme application.

* **Improvements**
  * More robust theme application with per-widget logging on failures.
  * More reliable cross-platform file opening so READMEs and files launch consistently on Windows, macOS, and Linux.

* **Tests**
  * Added cross-platform tests validating file-opening behavior on Linux, macOS, and Windows.

---
## Summary by cubic
Improves dashboard reliability by narrowing exception handling, logging silent failures, unifying file opens through a safe launcher, and simplifying the theme apply walker to avoid double configuration. Windows now uses `explorer.exe` to bypass cmd parsing; paths stay literal to prevent injection.

- **Bug Fixes**
  - Replace bare `except:` with `except Exception` and log context; critical signals now propagate.
  - Theme recolor logs at DEBUG with `exc_info=True` to avoid noise when widgets reject `background`.
  - Add `build_file_open_launch(...)`; route README/AGENTS through it. Windows uses `explorer.exe`, macOS uses `open`, Linux uses `xdg-open`—paths stay literal argv entries.
  - Closes #1406 and #1424.

- **Refactors**
  - Simplified theme apply walker by removing redundant per-child configure; recursion configures each widget once.

<sup>Written for commit f64dbca532a101ca9344f50d2ca3786fd656aa37. Summary will update on new commits.</sup>

### DIFF
--- a/ecc_dashboard.py
+++ b/ecc_dashboard.py
@@ -4,6 +4,7 @@ ECC Dashboard - Everything Claude Code GUI
 Cross-platform TkInter application for managing ECC components
 """
 
+import logging
 import tkinter as tk
 from tkinter import ttk, scrolledtext, messagebox
 import os
@@ -11,7 +12,13 @@ import json
 import subprocess
 from typing import Dict, List, Optional
 
-from scripts.lib.ecc_dashboard_runtime import build_terminal_launch, maximize_window
+from scripts.lib.ecc_dashboard_runtime import (
+    build_file_open_launch,
+    build_terminal_launch,
+    maximize_window,
+)
+
+logger = logging.getLogger(__name__)
 
 # ============================================================================
 # DATA LOADERS - Load ECC data from the project
@@ -98,8 +105,8 @@ def load_skills(project_path: str) -> List[Dict]:
                                 if line.startswith('# '):
                                     description = line[2:].strip()[:100]
                                     break
-                    except:
-                        pass
+                    except Exception:
+                        logger.exception("Failed to parse skill description from %s", skill_file)
                 
                 # Determine category
                 category = "General"
@@ -172,8 +179,8 @@ def load_commands(project_path: str) -> List[Dict]:
                             if line.startswith('# '):
                                 description = line[2:].strip()
                                 break
-                except:
-                    pass
+                except Exception:
+                    logger.exception("Failed to parse command description from %s", item)
                 
                 commands.append({
                     'name': cmd_name,
@@ -266,8 +273,8 @@ class ECCDashboard(tk.Tk):
         try:
             self.icon_image = tk.PhotoImage(file='assets/images/ecc-logo.png')
             self.iconphoto(True, self.icon_image)
-        except:
-            pass
+        except Exception:
+            logger.exception("Failed to load taskbar icon assets/images/ecc-logo.png")
         
         self.minsize(800, 600)
         
@@ -330,8 +337,8 @@ class ECCDashboard(tk.Tk):
             self.logo_image = tk.PhotoImage(file='assets/images/ecc-logo.png')
             self.logo_image = self.logo_image.subsample(2, 2)
             ttk.Label(header_frame, image=self.logo_image).pack(side=tk.LEFT, padx=(0, 10))
-        except:
-            pass
+        except Exception:
+            logger.exception("Failed to render header logo assets/images/ecc-logo.png")
         
         self.title_label = ttk.Label(header_frame, text="ECC Dashboard", font=('Open Sans', 18, 'bold'))
         self.title_label.pack(side=tk.LEFT)
@@ -799,19 +806,19 @@ Project: github.com/affaan-m/everything-claude-code"""
     
     def open_readme(self):
         """Open README in default browser/reader"""
-        import subprocess
         path = os.path.join(self.path_entry.get(), 'README.md')
         if os.path.exists(path):
-            subprocess.Popen(['xdg-open' if os.name != 'nt' else 'start', path])
+            argv, kwargs = build_file_open_launch(path)
+            subprocess.Popen(argv, **kwargs)
         else:
             messagebox.showerror("Error", "README.md not found")
-    
+
     def open_agents(self):
         """Open AGENTS.md"""
-        import subprocess
         path = os.path.join(self.path_entry.get(), 'AGENTS.md')
         if os.path.exists(path):
-            subprocess.Popen(['xdg-open' if os.name != 'nt' else 'start', path])
+            argv, kwargs = build_file_open_launch(path)
+            subprocess.Popen(argv, **kwargs)
         else:
             messagebox.showerror("Error", "AGENTS.md not found")
     
@@ -876,25 +883,28 @@ Project: github.com/affaan-m/everything-claude-code"""
         self.title_label.configure(font=(font_family, 18, 'bold'))
         self.version_label.configure(font=(font_family, 10))
         
+        # Many tkinter widgets (menus, system frames, notebook tabs) reject the
+        # `background` option silently; log at DEBUG so noise stays out of the
+        # default stream while still being discoverable when troubleshooting theming.
         def update_widget_colors(widget):
             try:
                 widget.configure(background=bg_color)
-            except:
-                pass
+            except Exception:
+                logger.debug("Skip background recolor on %r", widget, exc_info=True)
             for child in widget.winfo_children():
                 try:
                     child.configure(background=bg_color)
-                except:
-                    pass
+                except Exception:
+                    logger.debug("Skip background recolor on child %r", child, exc_info=True)
                 try:
                     update_widget_colors(child)
-                except:
-                    pass
-        
+                except Exception:
+                    logger.debug("Skip recursive recolor on child %r", child, exc_info=True)
+
         try:
             update_widget_colors(self)
-        except:
-            pass
+        except Exception:
+            logger.exception("Theme apply traversal failed at the root widget")
         
         self.update()
 
--- a/ecc_dashboard.py
+++ b/ecc_dashboard.py
@@ -886,16 +886,15 @@ Project: github.com/affaan-m/everything-claude-code"""
         # Many tkinter widgets (menus, system frames, notebook tabs) reject the
         # `background` option silently; log at DEBUG so noise stays out of the
         # default stream while still being discoverable when troubleshooting theming.
+        # The recursive call handles each child's own configure, so the loop just
+        # recurses — no per-child configure in addition (that was the pre-#1450
+        # shape and double-configured every descendant).
         def update_widget_colors(widget):
             try:
                 widget.configure(background=bg_color)
             except Exception:
                 logger.debug("Skip background recolor on %r", widget, exc_info=True)
             for child in widget.winfo_children():
-                try:
-                    child.configure(background=bg_color)
-                except Exception:
-                    logger.debug("Skip background recolor on child %r", child, exc_info=True)
                 try:
                     update_widget_colors(child)
                 except Exception:
--- a/scripts/lib/ecc_dashboard_runtime.py
+++ b/scripts/lib/ecc_dashboard_runtime.py
@@ -65,13 +65,30 @@ def build_file_open_launch(
     path: str,
     *,
     os_name: Optional[str] = None,
+    system_name: Optional[str] = None,
 ) -> Tuple[List[str], Dict[str, object]]:
     """Return safe argv/kwargs for opening an arbitrary file in the OS default handler.
 
     Mirrors the build_terminal_launch contract so dashboard callers can route every
     external launch through the same testable, injection-safe helper. The path is
     always passed as a separate argv entry, never interpolated into a shell string.
+
+    Per-platform launcher choice:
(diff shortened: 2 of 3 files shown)


shenchangmin:
Pushed \`9df8aeb\` to address @cubic-dev-ai's P1 and the coderabbit nitpick.

You're both right — the first draft shipped the same two cross-platform launcher bugs the pre-helper dashboard code had, I just relocated them into the helper. Fixed:

- **Windows**: \`start\` is a cmd.exe builtin, not an .exe. Now \`cmd.exe /c start "" <path>\` (empty quoted string is the window-title slot \`start\` requires when its first quoted argument is a path).
- **macOS**: \`xdg-open\` does not exist on Darwin. Now branches on \`platform.system() == 'Darwin'\` to use \`open\`, mirroring the structure of \`build_terminal_launch\` shipped in #1439.

Tests reworked into three full-argv \`deepStrictEqual\` assertions (Linux / macOS / Windows) — same pattern \`build_terminal_launch\` already uses. The P2 loose-substring concern on the Windows test is gone now that the assertion compares the full list.

Validation: \`tests/scripts/ecc-dashboard.test.js\` 6/6 passing; \`tests/run-all.js\` 1857 passed (baseline on main is 1854; +3 new tests, zero new failures).

shenchangmin:
Pushed \`e9eff44\` to address @cubic-dev-ai's second-round P1.

You're right — the \`cmd.exe /c start "" <path>\` wrapper I shipped in \`9df8aeb\` re-introduced cmd.exe's own parsing layer on top of argv, so a path containing \`&\` / \`|\` / \`^\` **without spaces** (so Python's \`list2cmdline\` wouldn't force-quote it) would be tokenised by cmd as a statement separator. That violates the helper's "path is always a literal argv entry" contract, even if the attack is self-inflicted in a local dashboard.

Switched Windows to \`['explorer.exe', path]\`:

- \`explorer.exe\` is a standalone executable, so Popen invokes it via CreateProcessW with no cmd.exe in the chain. Only \`CommandLineToArgvW\` parses the arg line, and that parser treats \`&\` / \`|\` / \`^\` as literal characters.
- End-user behaviour is equivalent: explorer hands the file to \`ShellExecute\` — the same mechanism \`os.startfile\` uses under the hood — and works identically for \`.md\`, \`.pdf\`, URLs, and folders.
- macOS and Linux branches untouched. Caller contract (\`subprocess.Popen(argv, **kwargs)\`) unchanged, so \`open_readme\` / \`open_agents\` need no edits.

Test updated to \`deepStrictEqual\` the full argv plus an explicit assertion that the launcher is **never** cmd.exe, so any future drift back toward cmd-wrapping fails the suite immediately.

Local: dashboard suite 6/6, \`tests/run-all.js\` 1857 passed (baseline on \`main\` 1854; +3 new tests, zero new failures).

shenchangmin:
Pushed \`f64dbca\` to address @greptile-apps's P2 on \`update_widget_colors\`.

You're right that the loop was double-configuring every descendant:

1. \`child.configure(background=bg_color)\` directly in the loop
2. \`update_widget_colors(child)\` which first-thing calls \`widget.configure(background=bg_color)\` on that same child

Dropped the inline \`child.configure\` and its try/except. The recursion alone now walks the tree and configures each widget exactly once. Net: 7 try/except in the file (was 8), same behaviour, less cognitive load.

The double-configure was pre-#1450 — this PR is just the first time the simplified version stays readable now that the per-widget error path logs via \`logger.debug\` rather than silently swallowing.

Local: dashboard suite 6/6 passing, \`tests/run-all.js\` 1857 passed (baseline 1854; unchanged from last push).

affaan-m:

PR #1485, opened 2026-04-17
fix(hooks): scope session-start injection by shortId to prevent cross-cwd leakage

stephan-roolvink:
## Summary

Fixes a cross-worktree / cross-cwd content leak in the `SessionStart` hook.

Before this patch, `scripts/hooks/session-start.js` picked the globally-newest `*-session.tmp` file from `~/.claude/sessions/` and injected it as `Previous session summary` into the context of every new Claude session, regardless of which cwd/project/worktree the session was for.

**Symptom I observed in the wild:** a Claude Agent SDK session dispatched for worktree A received worktree B's session summary and began making tool calls as if the task belonged to B — content-level crosstalk across worktrees/projects that share `~/.claude/sessions/`.

## Type
- [x] Hook

## Fix

- Import `getSessionIdShort` from `../lib/utils` and `parseSessionFilename` from `../lib/session-manager`
- Compute the current invocation's `shortId` (last 8 chars of `$CLAUDE_SESSION_ID`, else project/worktree basename)
- Filter `recentSessions` down to only those whose filename `shortId` matches the current invocation
- If no file matches, skip injection entirely rather than fall back to the globally-newest file — a missing summary is far safer than a wrong one
- Log the mismatch case at `log(...)` level so operators can still see when recent sessions exist but don't belong to this invocation

No public API change, no new dependencies. Pure filter added to an existing code path.

### New regression test
`tests/hooks/session-start-worktree-isolation.test.js` — simulates two worktrees (A, B) with session summaries where B's is newer on disk, then runs `session-start.js` once from each cwd and asserts:
- Running in worktree A stdout contains TASK-A and does NOT contain TASK-B
- Running in worktree B stdout contains TASK-B and does NOT contain TASK-A

Before the fix, test 1 fails: worktree A receives worktree B's summary (the globally-newest file), which is exactly the crosstalk bug.

### Existing tests updated
Three existing tests in `tests/hooks/hooks.test.js` previously relied on the globally-newest behaviour. They now pass `CLAUDE_SESSION_ID` so the invocation's shortId matches the test's session filename and the scenario still exercises the same code path:

- `injects real session content` — asserts real content still flows when shortIds match
- `excludes session files older than 7 days` — unified shortId across both recent/old files so maxAge filtering is what excludes the old one (the original intent)
- `injects newest session when multiple recent sessions exist` — both files share shortId so the newest-by-mtime still wins (the original intent)

No assertions weakened; each test still verifies exactly what it originally verified.

### Results

```
All 204 existing hooks.test.js tests pass.
2 new session-start-worktree-isolation.test.js tests pass.
evaluate-session.test.js and suggest-compact.test.js untouched — still pass.
```

## Checklist
- [x] Tested with Claude Code

---
## Summary by cubic
Scopes previous session injection in the `SessionStart` hook by shortId to stop cross-cwd/worktree leaks. If no matching session exists, nothing is injected.

- **Bug Fixes**
  - Use `getSessionIdShort` and `parseSessionFilename` to compute the current shortId and filter `~/.claude/sessions/*-session.tmp` to only matching files; pick the newest within scope.
  - Skip injection when no files match and log the mismatch instead of falling back to the global newest file.
  - Add `tests/hooks/session-start-worktree-isolation.test.js` to cover worktree isolation; update three existing tests to set `CLAUDE_SESSION_ID`. No API or dependency changes.

<sup>Written for commit dcb28956335770fa716ca92df46cec5c0a1a72b5. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed session content isolation to prevent previous session summaries from appearing in unrelated contexts.

* **Tests**
  * Updated session selection tests to validate scoped behavior.
  * Added regression test to verify session content remains isolated across different project contexts.

### DIFF
--- a/scripts/hooks/session-start.js
+++ b/scripts/hooks/session-start.js
@@ -9,6 +9,7 @@
  * sessions and learned skills.
  */
 
+const path = require('path');
 const {
   getSessionsDir,
   getLearnedSkillsDir,
@@ -16,8 +17,10 @@ const {
   ensureDir,
   readFile,
   log,
-  output
+  output,
+  getSessionIdShort
 } = require('../lib/utils');
+const { parseSessionFilename } = require('../lib/session-manager');
 const { getPackageManager, getSelectionPrompt } = require('../lib/package-manager');
 const { listAliases } = require('../lib/session-aliases');
 const { detectProjectType } = require('../lib/project-detect');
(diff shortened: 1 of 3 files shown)


affaan-m:

PR #1493, opened 2026-04-19
Scope SessionStart context injection

Terendelev:
## Summary
- skip previous-session summary injection on `SessionStart:resume`
- remove cross-project `recency-fallback` injection when no worktree/project match exists
- add regression tests for resume behavior and unrelated-project session files

## Why
In our Claude Desktop integration, we invoke `sdk.query(...)` per user turn while passing `resume` for the Claude runtime session. Claude emits `SessionStart` again on those resumed turns, and the current ECC hook re-injects previous-session context each time.

That led to two concrete issues:
- resumed turns received duplicate previous-session summaries
- a fresh session in one project could receive an unrelated summary from another project via `recency-fallback`

We hit a real case where a new `Claudedesktop` session was seeded with a `sts2sendbox` summary because the startup hook selected a recent unrelated session file.

## Notes
This keeps startup injection for matching worktree/project sessions, but avoids polluting resume turns and unrelated projects.

## Validation
- `node tests/hooks/hooks.test.js`
- Passed: 220

---
## Summary by cubic
Scopes previous-session context injection to startup only, and only for sessions that match the current worktree (or project if no worktree is recorded). Adds CLI `SessionStart` support and fails closed for unrecognized modes; skips injection on resume, clear, compact, and on malformed stdin.

- **Bug Fixes**
  - Only inject on explicit startup; skip on resume/clear/compact and default to skip for unrecognized `SessionStart` payloads.
  - On invalid stdin JSON, log a diagnostic with length only and skip injection; do not log raw payload.
  - Enforce matching rules: exact worktree match; project-name match only if the session lacks a recorded worktree; remove recency fallback.
  - Added tests for Desktop/CLI clear, malformed-stdin handling and logging, unrelated-project, and cross-worktree cases.

<sup>Written for commit d729c0d04efb1b36d4d57a9fdf1bddafd993e48d. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Session matching now enforces project/worktree boundaries, removes the recency fallback, and only injects previous-session summaries when the session-start mode permits; resume, malformed, skip or invalid stdin now skip injection and log truncated-payload or parse diagnostics.

* **Tests**
  * Added deterministic session-start fixture builder and tests covering resume behavior, malformed-stdin logging/length, max-age/newest selection, and cross-project/worktree non-injection.

PR #1481, opened 2026-04-17
fix(hooks): add ECC_SESSION_STRICT_MATCH to prevent cross-project session fallback

slurpyb:
## What Changed

Added env var guard `ECC_SESSION_STRICT_MATCH` to `selectMatchingSession()` in `scripts/hooks/session-start.js`. When set to `1`, the recency fallback (which loads the most recent session regardless of project) is skipped. Only worktree-matched or project-name-matched sessions are injected.

**Diff is 4 lines** — a comment block + one conditional check wrapping the existing fallback.

## Why This Change

The recency fallback introduced in #1054 preserves backward compatibility but causes **cross-project context injection** in multi-project workflows. When no session matches the current worktree or project name, the fallback grabs whatever session was saved most recently — often from a completely unrelated project.

**Real-world impact observed:**
- Working in `~/.claude` (config management), Claude received injected context from a Shopify theme project
- Injected context included project-specific instructions, file paths, dev server state, and prior conversation tone
- After context compaction, Claude cannot distinguish injected session history from actual conversation — treats cross-project context as ground truth
- Claude then references non-existent files, assumes wrong conventions, adopts patterns from the wrong codebase

**Why this is hard to catch:**
- Session injection happens in SessionStart hook — users never see the injected content
- `matchReason: 'recency-fallback'` is only logged to stderr
- After compaction, injected session summary is indistinguishable from real conversation
- Symptoms look like normal Claude mistakes, not a systematic data source issue

**Usage:**

Add to `~/.claude/settings.json`:

```json
{
  "env": {
    "ECC_SESSION_STRICT_MATCH": "1"
  }
}
```

Or as environment variable: `export ECC_SESSION_STRICT_MATCH=1`

**Design choices:**
- Opt-in (not opt-out) — zero impact on existing users
- Env var naming follows established `ECC_` prefix convention (`ECC_SESSION_RETENTION_DAYS`, `ECC_DISABLED_HOOKS`, `ECC_OBSERVE_SKIP`)
- Builds on #1054's matching infrastructure, just closes the escape hatch

## Testing Done

- [x] Manual testing completed
- [x] Automated tests pass locally — 1869/1869 (`ECC_SESSION_STRICT_MATCH` must be unset when running tests, as the env var leaks into test child processes and causes fallback-dependent tests to fail expectedly)
- [x] `ECC_SESSION_STRICT_MATCH=1` + directory with no matching session → no session injected
[...]

PR #1495, opened 2026-04-19
fix(hooks): isolate session-end.js filename using transcript_path UUID (#1494)

### LINKED ISSUE #1494: session-end.js: *-session.tmp corruption from subprocess spawn in Stop hook chain
## Summary

`scripts/hooks/session-end.js` produces corrupted `*-session.tmp` files when the Stop hook chain includes another hook (e.g. `scripts/csm/session-summary.js`) that spawns `claude -p ...` as a subprocess. Because the subprocess also fires the Stop hook chain, `session-end.js` is invoked twice with the same computed filename, and the subprocess overwrites the parent's valid summary with a summary of the AI-summarization prompt itself.

## Symptom

After a few sessions, `~/.claude/sessions/{date}-{shortId}-session.tmp` for certain projects looks like:

```markdown
### Tasks
- 以下はClaude Codeセッションの会話ログ抜粋です。日本語3行で簡潔に要約してください。  各行の形式: 1行目: 主なタスク・目的 ...

### Stats
- Total user messages: 1
```

Examples observed:
- `~/.claude/sessions/2026-04-19-obsidian-session.tmp`
- `~/.claude/sessions/2026-04-18-obsidian-session.tmp`

Healthy files (e.g. for actively developed projects) show many real user messages and correct stats.

## Root cause

Filename collision between the parent session and a subprocess spawned within the same Stop hook chain.

1. `scripts/lib/utils.js` `getSessionIdShort()` falls back to the project name when `CLAUDE_SESSION_ID` is not set:

   ```js
   function getSessionIdShort(fallback = 'default') {
     const sessionId = process.env.CLAUDE_SESSION_ID;
     if (sessionId && sessionId.length > 0) {
       const sanitized = sanitizeSessionId(sessionId.slice(-8));
       if (sanitized) return sanitized;
     }
     return sanitizeSessionId(getProjectName()) || sanitizeSessionId(fallback) || 'default';
   }
   ```
   → All sessions for the same project share the same filename.

2. `scripts/csm/session-summary.js` (a Stop-hook script registered in `settings.json`) calls `spawnSync('claude', ['-p', prompt], ...)` to generate an AI summary. The subprocess also runs the full Stop hook chain, including `session-end.js`.

3. In `scripts/hooks/session-end.js` (lines ~191-194):

   ```js
   const sessionsDir = getSessionsDir();
   const today = getDateString();
   const shortId = getSessionIdShort();
   const sessionFile = path.join(sessionsDir, `${today}-${shortId}-session.tmp`);
   ```

   Both the parent and the subprocess compute the **same** path. The subprocess transcript contains only the summarization prompt, so its `### Tasks` ends up holding the prompt text and `Total user messages: 1`. The parent's valid summary is overwritten.

### A. Derive `shortId` from `transcript_path` when available

In `session-end.js`, extract the first 8 hex chars of the session UUID from the transcript filename instead of relying solely on `getSessionIdShort()`:

```js
let shortId = null;
if (transcriptPath) {
  const m = path.basename(transcriptPath).match(/([0-9a-f]{8})-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\.jsonl$/i);
  if (m) shortId = m[1];
}
if (!shortId) shortId = getSessionIdShort(); // keep existing fallback
const sessionFile = path.join(sessionsDir, `${today}-${shortId}-session.tmp`);
```

This guarantees parent vs. subprocess map to different files because each session has a distinct UUID.

Trade-off: it increases the number of `.tmp` files in `~/.claude/sessions/`. Any SessionStart-hook logic that picks "the latest session for this project" should fall back to mtime ordering rather than filename equality. (Relevant PRs #1481, #1485 already tighten the matching logic on the read side, which is compatible with this change.)

### B. Subprocess identification flag (additive)

To avoid creating stray `.tmp` files from AI-summary subprocesses at all, pass an env flag when spawning the child and skip `session-end.js` when it is set.

In `scripts/csm/session-summary.js`:

```js
env: Object.assign({}, process.env, { ECC_SUMMARY_SUBPROCESS: '1' })
```

At the top of `session-end.js#main`:

```js
if (process.env.ECC_SUMMARY_SUBPROCESS === '1') process.exit(0);
```

A alone stops the corruption. B alone also stops it while keeping the session directory clean. Using both is safe and gives the most predictable behaviour.

## Reproduction

1. Install ECC, let `session-summary.js` be active in the Stop hook chain (default installation triggers this on certain project names).
2. Start a short Claude Code session in project A. End it normally.
3. Start another short session in the same project. End it.
4. Inspect `~/.claude/sessions/{today}-{shortId}-session.tmp`.
5. Expected: `### Tasks` contains several real user messages from the latest session; `Total user messages` > 1.
6. Observed: `### Tasks` contains the summarization prompt; `Total user messages: 1`.

## Environment

- OS: Windows 11 (the bug is platform-independent — same logic applies to macOS/Linux)
- Node: v20+
- Claude Code: v2.1.113
- ECC: cached plugin v1.8.0 + running against scripts in `~/.claude/scripts/` (synced from `main` on 2026-04-19)

## Notes

- Related but distinct from #1481 / #1485 / #1493, which all tighten the **read side** (`session-start.js`). This issue is about the **write side** (`session-end.js`).
- Happy to open a PR applying A + B with tests once the direction is acknowledged.

ratorin:
## Summary

Fixes #1494 — `*-session.tmp` corruption when another Stop hook spawns a `claude` subprocess.

In my own setup, a custom AI-summarization Stop hook spawned `claude -p ...` to generate a session summary. That subprocess also fires the full Stop hook chain, so `scripts/hooks/session-end.js` is invoked a second time. Because `getSessionIdShort()` falls back to the project/worktree name when `CLAUDE_SESSION_ID` is unset, both invocations produced the **same** `{date}-{shortId}-session.tmp` path. The subprocess transcript only contained the summarization prompt, so its `### Tasks` ended up holding the prompt text and `Total user messages: 1`, overwriting the parent's valid summary.

## Type
- [x] Hook

## Fix

When stdin JSON (or `CLAUDE_TRANSCRIPT_PATH`) provides a `transcript_path`, derive `shortId` from the first 8 hex characters of the session UUID in the filename instead of always calling `getSessionIdShort()`. The existing fallback is preserved for cases without `transcript_path`, so nothing changes for other callers.

```js
let shortId = null;
if (transcriptPath) {
  const m = path.basename(transcriptPath).match(/([0-9a-f]{8})-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\.jsonl$/i);
  if (m) shortId = m[1];
}
if (!shortId) shortId = getSessionIdShort();
```

Parent and subprocess invocations now resolve to different UUIDs, so they write to different files and do not collide.

## Testing

New regression test in `tests/hooks/hooks.test.js` (`derives shortId from transcript_path UUID when available`) runs `session-end.js` with stdin carrying a fake transcript path `/tmp/.../abcdef12-3456-4789-a012-bcdef3456789.jsonl`, without `CLAUDE_SESSION_ID` set, and asserts the resulting `.tmp` file name contains `abcdef12` (the UUID prefix) rather than the project-name fallback.

Full suite: `node tests/hooks/hooks.test.js` → **220 passed, 0 failed** on Windows 11 + Node 20.

## Notes / request for feedback

- This is my first contribution to this repo. I ran into the corruption while using a custom Stop hook in my own tooling and fixed it on my side, but the root cause is general enough that anyone spawning `claude` from another Stop hook would hit it. Happy to adjust the scope or style if you prefer a different direction.
- Side effect: the number of `.tmp` files under `~/.claude/sessions/` will grow faster because each real session now gets its own UUID-based file instead of reusing the project-name filename. Related open PRs #1481 / #1485 / #1493 are already tightening the read-side matching, so this should compose well with them, but if you want me to also touch up a retention or mtime-ordering pass on the read side in the same PR, let me know.
- Tested on Windows. The change is pure string parsing on `path.basename(...)` so macOS/Linux behavior should be identical, but I have not tested on those platforms.

If anything looks off, please comment and I will revise.

---
## Summary by cubic
Prevents session summary `.tmp` collisions when a Stop hook spawns a `claude` subprocess by deriving the filename shortId from the `transcript_path` UUID when available. Parent and subprocess now write to different files while keeping names compatible with `getSessionIdShort()`.

- **Bug Fixes**
  - Derive `shortId` from the last 8 hex chars of the transcript UUID (from stdin or `CLAUDE_TRANSCRIPT_PATH`), lowercase it, and route through `sanitizeSessionId()` to match `getSessionIdShort()` and avoid collisions.
  - When stdin omits `transcript_path` or provides an empty/non-string value (or malformed JSON), fall back to `CLAUDE_TRANSCRIPT_PATH`; if still absent, fall back to `getSessionIdShort()`.
  - Add regression tests for UUID-derived `shortId`, lowercase normalization, and parity when `CLAUDE_SESSION_ID` equals the transcript UUID.

<sup>Written for commit 0c3fc7074e2efed089839e26d6deff6554856a43. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Session file naming now prefers the UUID from the session transcript (using the final 8 hex characters, normalized to lowercase), with the previous fallback preserved — reducing cross-session file conflicts.

* **Tests**
  * Added regression tests validating transcript-driven session naming, lowercase normalization, and behavior when an existing session ID matches the transcript UUID.

### DIFF
--- a/scripts/hooks/session-end.js
+++ b/scripts/hooks/session-end.js
@@ -190,14 +190,19 @@ async function main() {
 
   const sessionsDir = getSessionsDir();
   const today = getDateString();
-  // Prefer the real session UUID (first 8 chars) from transcript_path when available.
+  // Derive shortId from transcript_path UUID when available, using the SAME
+  // last-8-chars convention as getSessionIdShort(sessionId.slice(-8)). This keeps
+  // backward compatibility for normal sessions (the derived shortId matches what
+  // getSessionIdShort() would have produced from the same UUID), while making
+  // every session map to a unique filename based on its own transcript UUID.
+  //
   // Without this, a parent session and any `claude -p ...` subprocess spawned by
-  // another Stop-hook share the project-name fallback filename, and the subprocess
+  // another Stop hook share the project-name fallback filename, and the subprocess
   // overwrites the parent's summary. See issue #1494 for full repro details.
   let shortId = null;
   if (transcriptPath) {
-    const m = path.basename(transcriptPath).match(/([0-9a-f]{8})-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\.jsonl$/i);
-    if (m) { shortId = m[1]; }
+    const m = path.basename(transcriptPath).match(/([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\.jsonl$/i);
+    if (m) { shortId = m[1].slice(-8).toLowerCase(); }
   }
   if (!shortId) { shortId = getSessionIdShort(); }
   const sessionFile = path.join(sessionsDir, `${today}-${shortId}-session.tmp`);
--- a/scripts/hooks/session-end.js
+++ b/scripts/hooks/session-end.js
@@ -16,6 +16,7 @@ const {
   getDateString,
   getTimeString,
   getSessionIdShort,
+  sanitizeSessionId,
   getProjectName,
   ensureDir,
   readFile,
@@ -202,7 +203,11 @@ async function main() {
   let shortId = null;
   if (transcriptPath) {
     const m = path.basename(transcriptPath).match(/([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\.jsonl$/i);
-    if (m) { shortId = m[1].slice(-8).toLowerCase(); }
+    if (m) {
+      // Run through sanitizeSessionId() for byte-for-byte parity with
+      // getSessionIdShort(sessionId.slice(-8)).
+      shortId = sanitizeSessionId(m[1].slice(-8).toLowerCase());
+    }
   }
   if (!shortId) { shortId = getSessionIdShort(); }
   const sessionFile = path.join(sessionsDir, `${today}-${shortId}-session.tmp`);
--- a/scripts/hooks/session-end.js
+++ b/scripts/hooks/session-end.js
@@ -179,14 +179,22 @@ function mergeSessionHeader(content, today, currentTime, metadata) {
 }
 
 async function main() {
-  // Parse stdin JSON to get transcript_path
+  // Parse stdin JSON to get transcript_path; fall back to env var on missing,
+  // empty, or non-string values as well as on malformed JSON.
   let transcriptPath = null;
   try {
     const input = JSON.parse(stdinData);
-    transcriptPath = input.transcript_path;
+    if (input && typeof input.transcript_path === 'string' && input.transcript_path.length > 0) {
+      transcriptPath = input.transcript_path;
+    }
   } catch {
-    // Fallback: try env var for backwards compatibility
-    transcriptPath = process.env.CLAUDE_TRANSCRIPT_PATH;
+    // Malformed stdin: fall through to the env-var fallback below.
+  }
+  if (!transcriptPath) {
+    const envTranscriptPath = process.env.CLAUDE_TRANSCRIPT_PATH;
+    if (typeof envTranscriptPath === 'string' && envTranscriptPath.length > 0) {
+      transcriptPath = envTranscriptPath;
+    }
   }
 
   const sessionsDir = getSessionsDir();
--- a/scripts/hooks/session-end.js
+++ b/scripts/hooks/session-end.js
@@ -190,7 +190,16 @@ async function main() {
 
   const sessionsDir = getSessionsDir();
   const today = getDateString();
-  const shortId = getSessionIdShort();
+  // Prefer the real session UUID (first 8 chars) from transcript_path when available.
+  // Without this, a parent session and any `claude -p ...` subprocess spawned by
+  // another Stop-hook share the project-name fallback filename, and the subprocess
+  // overwrites the parent's summary. See issue #1494 for full repro details.
+  let shortId = null;
+  if (transcriptPath) {
+    const m = path.basename(transcriptPath).match(/([0-9a-f]{8})-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\.jsonl$/i);
+    if (m) { shortId = m[1]; }
+  }
+  if (!shortId) { shortId = getSessionIdShort(); }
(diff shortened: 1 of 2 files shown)


ratorin:
Thanks for the reviews! Pushed 93cd5f4 to address the feedback:

**P1 (cubic, greptile):** The concern about unconditional override and orphaned files is valid. Rather than guarding with `!CLAUDE_SESSION_ID` (which would bring back the collision for subprocesses that inherit `CLAUDE_SESSION_ID` from their parent), I aligned the UUID extraction with `getSessionIdShort()`'s existing `.slice(-8)` convention. The new code extracts the last 8 chars of the transcript UUID, so for any single session, the filename is identical whether it comes from `CLAUDE_SESSION_ID` or from `transcript_path`. Existing `.tmp` files are not orphaned on upgrade, and sibling subprocesses (which have a different transcript UUID) still resolve to different filenames.

**P2 (greptile):** Added `.toLowerCase()` on the extracted hex prefix so uppercase UUIDs normalize the same way `sanitizeSessionId()` does.

**P2 (coderabbitai):** Added `CLAUDE_SESSION_ID: ''` explicitly in the test env to defeat parent-process leakage.

**P2 (greptile — missing coexistence test):** Added two new regression tests:
- `normalizes transcript UUID shortId to lowercase`
- `matches getSessionIdShort when transcript UUID equals CLAUDE_SESSION_ID` — explicitly guards the backward-compat guarantee.

Full suite: `222/222` passed locally on Windows 11 + Node 20.

Please take another look, and let me know if there's anything else to tweak. First time contributing here, so happy to follow your style/scope guidance.

ratorin:
Pushed 01d8167:

- **greptile P1**: Route transcript UUID shortId through `sanitizeSessionId()` to stay byte-for-byte compatible with `getSessionIdShort()` for any non-standard IDs.
- **coderabbit P2**: Reworded the inverted inline comment to correctly describe the transcript_path branch as the one being exercised.

Tests still green: 222/222.

ratorin:
Pushed 0c3fc70: broadened the `CLAUDE_TRANSCRIPT_PATH` env fallback to cover valid-JSON-but-missing/empty/non-string `transcript_path` (not only malformed JSON). Catches the edge case coderabbit flagged without regressing the malformed-JSON branch. Tests still green: 222/222.

PR #1567, opened 2026-04-24
fix(hooks): bypass gateguard friction gates inside subagent invocations (#1548)

### LINKED ISSUE #1548: Bug: gateguard-fact-force hook session ID isolation blocks subagent Bash calls on Linux
## Problem

The `gateguard-fact-force` hook uses `process.ppid` as fallback for session ID isolation, which causes subagents to get a different session ID than the parent process on Linux. This means each subagent is treated as a "new session" by gateguard, triggering the fact-forcing gate and blocking Bash calls.

## Root Cause

In `scripts/hooks/gateguard-fact-force.js` line 32:

```js
const SESSION_ID = process.env.CLAUDE_SESSION_ID || process.env.ECC_SESSION_ID || `pid-${process.ppid || process.pid}`;
```

When Claude Code spawns a subagent (via the `Agent` tool), the child process does not inherit `CLAUDE_SESSION_ID` or `ECC_SESSION_ID` environment variables. The fallback `pid-${process.ppid}` produces different values for the parent and child processes on Linux, resulting in separate state files and independent gateguard sessions.

## Impact

- Subagents are blocked on their first Bash call every session (gateguard treats them as new sessions)
- Workaround requires setting `ECC_DISABLED_HOOKS` env var to disable gateguard entirely
- Affects Linux environments; Windows appears unaffected (likely due to different process model or env var inheritance behavior)

## Expected Behavior

Subagents spawned by Claude Code should share the same gateguard session state as the parent process, so that once the parent session has satisfied the fact-forcing gate, subagents are not re-gated.

## Suggested Fix

1. Ensure `CLAUDE_SESSION_ID` is propagated to subagent processes, OR
2. Use a more stable session identifier that is consistent across parent and child processes (e.g., based on the workspace root or a shared temp file), OR
3. Add subagent detection logic to skip the gate when running inside a delegated agent context

## Environment

- OS: Linux (Ubuntu)
- ECC version: 1.10.0
- Claude Code: latest
- Works fine on Windows

gaurav0107:
## Summary

Fixes #1548. GateGuard's fact-forcing gate re-fires for every subagent
(Task/Agent tool) call because subagents frequently get a different
`session_id` than the parent on Linux, producing a fresh empty state file.
Users report being blocked on the first Bash call of every subagent and
have to disable gateguard entirely via `ECC_DISABLED_HOOKS` as a workaround.

Claude Code's hook payload carries `parent_tool_use_id` for tool calls
spawned inside a subagent. When that field is present, the parent session
has already satisfied the fact-forcing gate for the user's instruction, so
re-gating the subagent adds friction with no safety value.

## Change

- New helper `isSubagentInvocation(data)` — checks for non-empty
  `parent_tool_use_id` (or camelCase `parentToolUseId`).
- Bypasses the Edit / Write / MultiEdit first-touch gates for subagents.
- Bypasses the routine first-Bash gate for subagents.
- **Keeps the destructive-Bash gate active for subagents.** Commands
  matching the `DESTRUCTIVE_BASH` regex still require facts — safety gate,
  not friction gate.

## Tests

6 new cases in `tests/hooks/gateguard-fact-force.test.js`:

- Subagent routine Bash allowed with fresh `session_id` + `parent_tool_use_id`
- Subagent destructive Bash still gated
- Subagent Edit allowed
- Subagent MultiEdit allowed
- camelCase `parentToolUseId` variant honored
- Empty-string `parent_tool_use_id` must NOT be treated as subagent

Result: 25/25 gateguard tests pass, full suite 1882/1882.

## Test plan

- [x] Reproduce #1548: subagent Bash denied on first call with fresh session_id
- [x] With fix: subagent Bash passes through
- [x] Top-level first Bash still gated (no regression)
- [x] Subagent `rm -rf …` still gated (safety preserved)
- [x] `npm test` suite green (1882/1882)
- [x] `npx eslint` clean on changed files

Closes #1548

---
## Summary by cubic
Prevents GateGuard from re-gating subagent tool calls by detecting subagent context and bypassing first-touch file gates, fixing first-call blocks in #1548. Bypass applies to file tools only; Bash stays gated for safety.

- **Bug Fixes**
  - Detect subagents via `agent_id`/`agentId`, with fallback to `parent_tool_use_id`/`parentToolUseId`.
  - Bypass first-touch gates for `Edit`, `Write`, and `MultiEdit` in subagents; `Bash` (routine and destructive) remains gated, with `DESTRUCTIVE_BASH` still enforced.
  - Preserve `.claude/settings.json` pass-through by running that check before the subagent bypass.
  - Add strict A/B tests for subagent behavior; annotate PASS/FAIL emoji replacement in `docs/fixes/HOOK-FIX-20260421-ADDENDUM.md` to satisfy unicode-safety CI.

<sup>Written for commit d3b1621a9f8c44296488cf04219463b07db11513. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Improvements**
  * Subagent-detected operations now bypass routine file-operation friction checks while preserving destructive-action safeguards for shell commands.

* **Tests**
  * Added tests covering subagent detection (multiple marker variants), invalid-marker rejection, multi-file and sequential subagent scenarios, and continued gating for destructive Bash commands.

* **Documentation**
  * Updated workaround guidance text; replaced emoji status indicators with plain PASS/FAIL and clarified Bash outcome notes.

### DIFF
--- a/docs/fixes/HOOK-FIX-20260421-ADDENDUM.md
+++ b/docs/fixes/HOOK-FIX-20260421-ADDENDUM.md
@@ -87,9 +87,9 @@ hook command が `"C:\Program Files\Git\bin\bash.exe" "C:\Users\...\wrapper.sh"`
 ### 回避策
 
 第1トークンを bash.exe のフルパス＋スペース付きパスにしないこと：
-1. ✅ `bash` （PATH 解決の単一トークン）— 夜fix / hooks.json パターン
-2. ✅ `.sh` 直接パス（Claude Code の .sh ハンドリングに依存）— 朝fix
-3. ❌ `"C:\Program Files\Git\bin\bash.exe" "<path>"` — 1トークン目が quoted で空白込み
+1. PASS: `bash` （PATH 解決の単一トークン）— 夜fix / hooks.json パターン
+2. PASS: `.sh` 直接パス（Claude Code の .sh ハンドリングに依存）— 朝fix
+3. FAIL: `"C:\Program Files\Git\bin\bash.exe" "<path>"` — 1トークン目が quoted で空白込み
 
 ## 結論
 
--- a/docs/fixes/HOOK-FIX-20260421-ADDENDUM.md
+++ b/docs/fixes/HOOK-FIX-20260421-ADDENDUM.md
@@ -86,6 +86,14 @@ hook command が `"C:\Program Files\Git\bin\bash.exe" "C:\Users\...\wrapper.sh"`
 
 ### 回避策
 
+<!--
+  Note: this list originally used check-mark (U+2705) and
+  cross-mark (U+274C) emoji, but the repo-wide check-unicode-safety
+  CI rejects those codepoints. PR #1567 (gateguard subagent fix)
+  was blocked by these pre-existing violations, so they are
+  replaced with plain-text PASS/FAIL to unblock CI. Not a content
+  change - semantics are identical.
+-->
 第1トークンを bash.exe のフルパス＋スペース付きパスにしないこと：
 1. PASS: `bash` （PATH 解決の単一トークン）— 夜fix / hooks.json パターン
 2. PASS: `.sh` 直接パス（Claude Code の .sh ハンドリングに依存）— 朝fix
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -328,19 +328,38 @@ function routineBashMsg() {
 
 // --- Subagent detection ---
 //
-// Claude Code's hook payload includes `parent_tool_use_id` for tool calls
-// spawned inside a Task/Agent subagent. Top-level tool calls have no such
-// parent. When a subagent is present, the main session has already satisfied
-// the fact-forcing gate for the user's request — re-gating every subagent
-// produces pure friction (see #1548) without adding any safety value.
-// Destructive Bash remains gated even in subagent context (safety gate,
-// not a friction gate).
+// Claude Code's hook payload carries `agent_id` (with `agent_type`) when the
+// hook fires from within a Task/Agent subagent. Per the documented hook
+// schema: "Present only when the hook fires from within a subagent (e.g.,
+// a tool called by an AgentTool worker). Absent for the main thread, even
+// in --agent sessions. Use this field (not agent_type) to distinguish
+// subagent calls from main-thread calls." The rest of this repo already
+// keys off `agent_id` for the same purpose (see
+// skills/continuous-learning-v2/hooks/observe.sh and tests/hooks/hooks.test.js).
+//
+// Some internal Claude Code stream events carry `parent_tool_use_id`
+// instead. We check it as a defensive fallback in case a future hook
+// payload surfaces that field, but the canonical documented signal is
+// `agent_id`.
+//
+// When a subagent is detected, the main session has already satisfied the
+// first-touch fact-forcing gate for the user's request, so re-gating
+// Edit/Write/MultiEdit on the subagent's first touch of each file produces
+// pure friction (see #1548) without adding safety value. Bash (both
+// routine AND destructive) stays gated in subagent context: the routine-
+// Bash gate is the containment net for destructive commands that slip
+// past the narrow DESTRUCTIVE_BASH regex (rm -f, mkfs, curl|sh, etc.).
 
 function isSubagentInvocation(data) {
   if (!data || typeof data !== 'object') {
     return false;
   }
-  const candidates = [data.parent_tool_use_id, data.parentToolUseId];
+  const candidates = [
+    data.agent_id,
+    data.agentId,
+    data.parent_tool_use_id,
+    data.parentToolUseId
+  ];
   for (const candidate of candidates) {
     if (typeof candidate === 'string' && candidate.trim()) {
       return true;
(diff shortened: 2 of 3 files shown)


gaurav0107:
## Update — addressing review findings + CI

Pushed commit `b81237a` addressing all blocking and major findings from the bot reviews plus a second-opinion review. Additive commit, no force-push.

### Feedback → commit mapping

| # | Source | Finding | Addressed by |
|---|--------|---------|--------------|
| 1 | **Second-opinion [BLOCKING]** | `parent_tool_use_id` is not in the documented PreToolUse hook payload | `isSubagentInvocation()` now checks `agent_id` (primary, per the Claude Code hook schema's own docstring: *"Use this field (not agent_type) to distinguish subagent calls from main-thread calls"*) and falls back to `parent_tool_use_id` for defense in depth. This also aligns with `skills/continuous-learning-v2/hooks/observe.sh:149` and `tests/hooks/hooks.test.js:2841`, which already use `agent_id`. |
| 2 | **Second-opinion [MAJOR]** | Routine-Bash subagent bypass creates a safety hole: `DESTRUCTIVE_BASH` regex misses `rm -f`, `mkfs`, `curl \| sh`, `chmod -R 777 /`, `git push -f` short flag, etc. | **Dropped the routine-Bash bypass.** Bypass now only applies to Edit/Write/MultiEdit. The routine-Bash gate stays in subagents as the containment net for commands that slip past the narrow regex. This is the "preferred fix" option from the review. |
| 3 | **Second-opinion [MAJOR]** | Edit/Write bypass skipped `isClaudeSettingsPath` — subagent could Write to `.claude/settings.json` with zero friction | The settings-path short-circuit now runs BEFORE the subagent bypass in the Edit/Write branch, preserving the existing pass-through semantics for that path. |
| 4 | **Second-opinion [MINOR]** | Tests used `notStrictEqual('deny')` — passes for any pass-through, not specifically the bypass path | All new subagent tests use strict A/B assertions: same input without subagent marker denies; same input with marker allows. |
| 5 | **Second-opinion [MINOR]** | Missing tests for non-string `agent_id`, destructive commands in subagents, two-sequential-subagent repro | Added tests 23 (routine Bash still gated in subagents), 24 (destructive Bash still gated), 27 (empty / numeric / null / whitespace `agent_id` all rejected), 28 (two sequential subagent Edits reproducing #1548). |
| 6 | **Greptile [P2]** | Test numbering out of order (tests 20-25 appeared before Test 19) | New test block (now 20-30) moved AFTER Test 19. Source order matches label order. |
| 7 | **Greptile [P2]** | Missing dedicated subagent `Write` test | Added Test 21 (A/B for Write) and Test 29 (subagent MultiEdit mixing settings path + regular file). |
| 8 | **CI — Validate Components → check-unicode-safety** | `docs/fixes/HOOK-FIX-20260421-ADDENDUM.md:90-92` contains `U+2705` and `U+274C` emojis blocking the workflow | Replaced with the same ASCII markers the CI script itself maps to (`PASS:` and `FAIL:`). **Note:** these violations exist on `main` as well, not introduced by this PR; the check scans the whole repo so the PR-branch CI inherits the failure. Flagging in case this should be a separate commit / separate PR, happy to split. |

### Key files changed

- `scripts/hooks/gateguard-fact-force.js` — `isSubagentInvocation()` widened to check `agent_id` primarily; subagent bypass removed from routine-Bash branch; settings-path short-circuit reordered above subagent bypass.
- `tests/hooks/gateguard-fact-force.test.js` — subagent block rewritten with A/B assertions and moved after Test 19; 6 new tests (empty/non-string marker rejection, #1548 repro, Write A/B, MultiEdit + settings path, top-level settings-path regression guard).
- `docs/fixes/HOOK-FIX-20260421-ADDENDUM.md` — unicode-safety fix.

### Tradeoff disclosure

The original PR bypassed routine-Bash in subagents, which fully eliminated the first-call friction from #1548 on all tool types. Dropping that bypass means subagents still see the routine-Bash gate once per session. Rationale: the safety cost of the bypass (unlimited silent passes for non-regex-matched destructive commands) outweighs the extra one-time friction for routine Bash. The primary friction source in #1548 — first-touch Edit/Write/MultiEdit on every subagent — is still fully addressed. If the maintainer prefers a wider `DESTRUCTIVE_BASH` regex + full subagent bypass instead, I'm happy to iterate.

### Test results

- `node tests/run-all.js`: **1882 / 1882** pass (full suite).
- `node tests/hooks/gateguard-fact-force.test.js`: **30 / 30** pass.
- `node scripts/ci/check-unicode-safety.js`: clean.
- `npx eslint scripts/hooks/gateguard-fact-force.js tests/hooks/gateguard-fact-force.test.js`: clean.
- `node scripts/ci/validate-hooks.js`: clean.
- `node scripts/ci/validate-no-personal-paths.js`: clean.

gaurav0107:
Added an inline HTML comment in `docs/fixes/HOOK-FIX-20260421-ADDENDUM.md` 
explaining why the check-mark/cross-mark emoji were replaced with plain-text PASS/FAIL — this makes the scope-bleed self-documenting in-file rather than only in the PR description. No content change; semantics identical.

If not aligned happy to remove this. @affaan-m

affaan-m:
Thanks for the original fix direction here. I ported the useful subagent GateGuard bypass onto current main in maintainer PR #1710, while preserving the newer mainline GateGuard hardening that had landed after this branch diverged. Closing this PR as superseded by #1710.

Issue #1588, opened 2026-04-27
bug(hooks): block-no-verify false positives when --no-verify / -n / core.hooksPath= appears inside quoted commit messages


Jamkris:
## Summary

`scripts/hooks/block-no-verify.js` runs its flag-detection regexes against the raw command string with no quote handling. As a result, perfectly legitimate commits whose **message** happens to contain `--no-verify`, `-n`, or `core.hooksPath=` are blocked.

This was caught while porting the hook to a downstream project — the review on the port flagged it and I implemented a fix there. Filing upstream so the same fix lands here if you want it.

## Reproduction

All three commands should succeed (they don't actually pass `--no-verify` to git), but the hook blocks them:

```bash
# Case 1: --no-verify discussed in a double-quoted commit message
echo '{"tool_input":{"command":"git commit -m \"fix: --no-verify edge case\""}}' \
  | node scripts/hooks/block-no-verify.js
# → exit 2, BLOCKED: --no-verify flag is not allowed with git commit.

# Case 2: -n shorthand inside a quoted message (commit subcommand only)
echo '{"tool_input":{"command":"git commit -m \"Fixed -n bug in module\""}}' \
  | node scripts/hooks/block-no-verify.js
# → exit 2

# Case 3: core.hooksPath= mentioned in a quoted message or doc string
echo '{"tool_input":{"command":"git commit -m \"doc: explain core.hooksPath= setting\""}}' \
  | node scripts/hooks/block-no-verify.js
# → exit 2
```

The first regex match comes from this region of the hook:

```js
function hasNoVerifyFlag(input, command, offset) {
  const region = input.slice(offset);
  if (/--no-verify\b/.test(region)) return true;
  if (command === 'commit') {
    if (/\s-n(?:\s|$)/.test(region) || /\s-n[a-zA-Z]/.test(region)) return true;
  }
  return false;
}

function hasHooksPathOverride(input) {
  return /-c\s+["']?core\.hooksPath\s*=/.test(input);
}
```

`region` (and `input`) include any quoted commit-message text verbatim, so the regexes match inside the message.

## Why it matters

Frequency in average repos is low, but the failure mode is corrosive:

- A user who hits a false positive will either rewrite their message (annoying, especially for security-related repos that *do* discuss these flags) or disable the hook outright. The latter defeats the whole point — `block-no-verify` is precisely the hook you don't want users disabling.
[...]

Issue #1359, opened 2026-04-11
security: PR #1341 workflow_run + fork checkout = privilege escalation vector


konghun:
## Summary

PR #1341 ("Test/ai repair") introduces `.github/workflows/ai-repair.yml` with a critical security pattern: it uses the `workflow_run` event (which runs with the **base repo's full permissions**) while checking out the **untrusted head/fork branch** and executing code from that branch.

## Vulnerable Pattern

```yaml
# Runs in base-repo context — has write permissions
permissions:
  contents: write
  pull-requests: write
  actions: read

on:
  workflow_run:
    workflows: ["CI"]
    types: [completed]

steps:
  # Checks out fork code — NOT the base repo
  - uses: actions/checkout@v4
    with:
      ref: ${{ github.event.workflow_run.head_branch }}

  # Installs and runs Python from that fork checkout
  - run: |
      pip install -r projects/ai-pipeline-repair/requirements.txt
      python -m src.main repair ...
```

## Attack Scenario

1. Attacker forks the repo and submits a PR.
2. CI runs and (deliberately or naturally) fails.
3. `ai-repair.yml` triggers via `workflow_run`, running with **base-repo `contents: write` + `pull-requests: write`** permissions.
4. The workflow checks out **the attacker's fork branch** and runs `pip install -r projects/ai-pipeline-repair/requirements.txt` + arbitrary Python from that branch.
5. Malicious `requirements.txt` or Python code can exfiltrate `ANTHROPIC_API_KEY`/`GITHUB_TOKEN`, push to any branch, or create PRs.

## Additional Issues in PR #1341

- All PR template checkboxes are unchecked (security checklist, testing, etc.).
- `npm ci --ignore-scripts || true` silently ignores install errors.
- No pinned lockfile for Python dependencies.
- Massive scope (131 files, 12,339 additions) with an empty PR description.

## Recommendation

**Do not merge PR #1341 in its current form.**

If the auto-repair workflow is desired:

1. **Never** check out or execute code from the untrusted branch inside a `workflow_run` job. The repair logic should run from the base branch only, reading logs/artifacts via the API.
2. Pin all Python dependencies with a lockfile.
3. Require all PR template checklist items to be signed off before merge.

## References

- [GitHub Security Lab: Stealing secrets from a GitHub Actions workflow](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)
[...]

PR #1589, opened 2026-04-27
fix(ci): detect unsafe checkouts even when uses: value is quoted

Jamkris:
## Summary

`extractCheckoutSteps` filters step blocks down to ones that invoke `actions/checkout`. The current regex requires the `uses:` value to be unquoted, so any step written as `uses: "actions/checkout@v4"` is silently skipped — including its `with: ref:` and `with: repository:` lines, which is exactly the surface this validator exists to inspect.

Both forms are valid YAML, and quoted values appear in the wild (some style guides quote action names; some YAML formatters add quotes when version strings start with digits). A contributor following such a style — or an attacker who notices the gap — can submit a PR with an unsafe `ref: ${{ github.event.pull_request.head.sha }}` and the validator will pass it.

Same class of issue as the original motivation for this script (#1359), just applied to a step the regex doesn't recognize as a checkout in the first place.

## Change

One-character regex update: allow an optional `'` or `"` before `actions/checkout@`. No behavior change for the unquoted form.

```diff
-    .filter(block => /uses:\s*actions\/checkout@/m.test(block.text));
+    .filter(block => /uses:\s*['"]?actions\/checkout@/m.test(block.text));
```

## Test added

`tests/ci/validate-workflow-security.test.js` gets one new fixture that mirrors the existing `pull_request_target` + `pull_request.head.sha` case but with `uses: "actions/checkout@v4"`. Without the fix this case returns exit 0 (validator silently passes); with the fix it returns non-zero and emits the expected violation.

## Test plan

- [x] `node tests/ci/validate-workflow-security.test.js` — 5/5 pass (was 4/4)
- [x] Existing tests still pass (no regression on unquoted form)
- [x] `node scripts/ci/validate-workflow-security.js` still passes against this repo's actual `.github/workflows/` (7 files clean)

## Related

- Original motivation: #1359
- Found while porting the validator to a downstream Gemini CLI extension ([reference fix in `Jamkris/everything-gemini-code#47`, commit `ce2ca96`](https://github.com/Jamkris/everything-gemini-code/pull/47/commits/ce2ca96)).
- The downstream port also fixes a separate `block-no-verify` false-positive class — filed as a follow-up issue (#1588) rather than bundling here, since it warranted a design discussion before sending a PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * CI workflow security validation now correctly recognizes action references wrapped in single or double quotes, ensuring consistent rule enforcement across YAML formats.
* **Tests**
  * Added CI validation tests covering quoted action entries in pull-request workflows to prevent regressions.

---
## Summary by cubic
Fixes a bypass where the workflow-security validator skipped `actions/checkout` when `uses:` was quoted. Now it detects quoted forms and flags unsafe `ref` values.

- **Bug Fixes**
  - Update regex to allow optional quotes: `/uses:\s*['"]?actions\/checkout@/`.
  - Add tests for both `"actions/checkout@v4"` and `'actions/checkout@v4'` to ensure unsafe refs are caught.

<sup>Written for commit e0ac4c164373259fe0ddb08c9a19d48b460a099c. Summary will update on new commits.</sup>

### DIFF
--- a/scripts/ci/validate-workflow-security.js
+++ b/scripts/ci/validate-workflow-security.js
@@ -75,7 +75,7 @@ function extractCheckoutSteps(source) {
       startLine: block.startLine,
       text: block.lines.join('\n'),
     }))
-    .filter(block => /uses:\s*actions\/checkout@/m.test(block.text));
+    .filter(block => /uses:\s*['"]?actions\/checkout@/m.test(block.text));
 }
 
 function findViolations(filePath, source) {
--- a/tests/ci/validate-workflow-security.test.js
+++ b/tests/ci/validate-workflow-security.test.js
@@ -84,11 +84,21 @@ function run() {
   // Quoted action names (`uses: "actions/checkout@v4"`) are valid YAML and
   // appear in the wild. The original `uses:\s*actions/checkout@` filter
   // skipped quoted forms, letting unsafe `with: ref:` slip past the validator.
-  if (test('detects unsafe checkout when uses: value is quoted', () => {
+  // Cover both double- and single-quoted forms since the regex's ['"]? class
+  // makes them distinct branches.
+  if (test('detects unsafe checkout when uses: value is double-quoted', () => {
     const result = runValidator({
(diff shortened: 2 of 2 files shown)


affaan-m:
Thanks for the workflow-security validator fix. I ported the quoted `actions/checkout` detection and coverage onto current `main` in #1717, so I’m closing this as superseded.

PR #1648, opened 2026-04-30
fix: parse all Claude content blocks

guyua9:
## Summary
- parse every Claude response content block instead of only inspecting the first one
- preserve text from multiple text blocks and collect all `tool_use` blocks
- add regression coverage for mixed text/tool responses and multiple tool calls

## Tests
- `PYTHONPATH=src pytest -q tests/test_claude_provider.py tests/test_types.py tests/test_resolver.py tests/test_executor.py`

---
## Summary by cubic
Fixes Claude provider parsing by iterating every response content block. We now merge all text blocks and collect every `tool_use` call with safe, normalized arguments.

- **Bug Fixes**
  - Iterate all blocks; join text; preserve order of multiple `tool_use`.
  - Coerce `input` (dict or object) to a copied dict to avoid mutation.
  - Return `tool_calls` as `None` when no tools are present.
  - Add unit tests for mixed text/tool replies, multiple tool calls, ordering, and argument copying.

<sup>Written for commit 3c724f2e1d09034efa21666f66876d97a22a32c8. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude response parsing to aggregate text from all content blocks, correctly extract tool calls, and return no tool calls when none are present; tool-call arguments are safely copied to avoid later mutation.

* **Tests**
  * Added unit tests verifying aggregation of text blocks, extraction and ordering of tool calls, argument copying, and empty-text behavior.

* **Tests**
  * Registered a "unit" pytest marker in test configuration.

### DIFF
--- a/src/llm/providers/claude.py
+++ b/src/llm/providers/claude.py
@@ -64,20 +64,32 @@ class ClaudeProvider(LLMProvider):
 
             response = self.client.messages.create(**params)
 
-            tool_calls = None
-            if response.content and hasattr(response.content[0], "type"):
-                if response.content[0].type == "tool_use":
-                    tool_calls = [
+            text_parts: list[str] = []
+            tool_calls: list[ToolCall] = []
+            for block in response.content or []:
+                block_type = getattr(block, "type", None)
+                if block_type == "text":
+                    text = getattr(block, "text", "")
+                    if text:
+                        text_parts.append(text)
+                elif block_type == "tool_use":
+                    raw_arguments = getattr(block, "input", {})
+                    arguments = (
+                        raw_arguments.copy()
+                        if isinstance(raw_arguments, dict)
+                        else getattr(raw_arguments, "__dict__", {}).copy()
+                    )
+                    tool_calls.append(
                         ToolCall(
-                            id=getattr(response.content[0], "id", ""),
-                            name=getattr(response.content[0], "name", ""),
-                            arguments=getattr(response.content[0].input, "__dict__", {}),
+                            id=getattr(block, "id", ""),
+                            name=getattr(block, "name", ""),
+                            arguments=arguments,
                         )
-                    ]
+                    )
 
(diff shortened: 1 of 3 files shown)


huyua9:
Thanks for the review. I pushed `e8060ac6` to address the CodeRabbit nitpick.

Changes:
- `ClaudeProvider.generate()` now copies dict-like `tool_use.input` arguments before storing them in `ToolCall.arguments`, so downstream mutations cannot mutate the original SDK payload by reference.
- Added a regression test proving a later mutation of the raw SDK dict does not affect the emitted `ToolCall.arguments`.

Validation:
- `python -m pytest tests/test_claude_provider.py` — 3 passed
- `git diff --check upstream/main...HEAD` — passed

guyua9:
Thanks for the follow-up review. I pushed `3c724f2` to address the latest CodeRabbit test nits:

- `make_response()` now accepts a configurable `stop_reason`, with a new pure-text regression test asserting `tool_calls is None` for an `end_turn` response.
- Added `@pytest.mark.unit` to the Claude provider tests and registered the marker via `tests/conftest.py`.
- Added explicit type annotations to the fake client/helper signatures used by these tests.

Validation:

- `python -m pytest tests/test_claude_provider.py` — 4 passed
- `git diff --check upstream/main...HEAD` — passed

affaan-m:
Thanks for the PR. I am closing this in the cleanup pass because it is either stale/non-green against current main or too broad to land safely without a dedicated review window. Reopen is welcome with a current rebase and focused scope; maintainers may also port useful pieces into a fresh PR where appropriate.

affaan-m:
Follow-up: the focused useful part of this closed PR has now been ported and merged through maintainer PR #1724: https://github.com/affaan-m/everything-claude-code/pull/1724. Thanks again for the contribution.

PR #1694, opened 2026-05-07
skills: portfolio curation — dedup, slim, drift-prone markers

vincenttpe:
## Summary

Audit-driven curation pass over 11 global skills. Three concerns addressed:

1. **Lane overlap** — `coding-standards`, `backend-patterns`, and `api-design` had duplicated REST conventions, response envelope guidance, validation examples, auth, and rate-limiting. Each now has one clear lane with explicit cross-references.
2. **Embedded fragility / tool assumptions** — `verification-loop` had inline secret regex and debug-statement grep that overlapped with `/security-review` and lint hooks. `tdd-workflow` had a stack-command table that duplicated `verification-loop` and per-language testing skills. `search-first` had `Task(...)` legacy tool naming and assumed external tools were always available without checking.
3. **Drift exposure** — 4 external-tool-dependent skills (`exa-search`, `deep-research`, `fal-ai-media`, `x-api`) had no \`drift-prone\` marker. After this PR all 11 skills the audit flagged carry a verify-first banner.

The CSP guidance in \`security-review\` is also corrected: previous text included \`'unsafe-inline'\` as a defaulted fallback (with a misleading comment about \`'strict-dynamic'\` superseding it for style-src, which it doesn't). The new text explicitly forbids blanket \`'unsafe-*'\` and frames any need for them as tech debt with a removal plan.

## Stats

| Skill | Before | After | Δ | Type |
|-------|--------|-------|---|---|
| verification-loop | 174 | 146 | -16% | rewrite |
| tdd-workflow | 238 | 174 | -27% | rewrite |
| security-review | 605 | 378 | -38% | rewrite + CSP fix |
| api-design | 523 | 431 | -18% | dedup cut |
| coding-standards | 530 | 438 | -17% | dedup cut |
| backend-patterns | 598 | 475 | -21% | dedup cut + remove unsafe rate-limiter |
| **Subtotal (rewrites/cuts)** | **2668** | **2042** | **-23%** | |
| search-first | 170 | 242 | +42% | adds Step 0 preflight + tool-name fix |
| exa-search / deep-research / fal-ai-media / x-api | — | +preamble (~12 lines each) | additive | drift markers |

Total: 11 files, +758 / -1334.

## Concrete behaviour changes (not just slimming)

- **CSP default fixed** — \`security-review\` no longer documents \`'unsafe-inline'\` as a CSP fallback. New text: \`> Do NOT default to 'unsafe-inline' or 'unsafe-eval' — they neutralize CSP's main purpose.\`
- **Naïve rate-limiter removed** — \`backend-patterns\` had a 40-line in-memory \`RateLimiter\` example that silently breaks in any multi-instance / serverless deployment (each replica has its own counter). Replaced with a pointer to \`/api-design\` which has the correct shared-store guidance.
- **Project-entrypoint preferred for verification** — \`verification-loop\` now treats \`make verify\` / \`npm run verify\` / \`just verify\` / declared task-runner targets as authoritative, falling back to the stack table only when no entrypoint exists. Honours what the project author actually wrote.
- **Tool-availability preflight** — \`search-first\` adds Step 0: explicit checks for repo grep, package managers, \`gh\`, Context7 MCP, Exa MCP, skills directory. Each has a "if missing" fallback so the skill degrades gracefully rather than silently skipping search channels and reporting "found nothing".
- **Tool naming corrected** — \`search-first\` updated \`Task(...)\` → \`Agent(...)\`. The legacy \`Task\` name still appears in older Claude Code docs but is no longer the canonical tool. Inline note added so future readers don't get confused.
- **Per-stack handoffs explicit** — every cut adds a \`> see <other-skill>\` pointer at the deletion site so users land at the right doorway.

## Convergent maintainer work

The audit also flagged \`continuous-agent-loop\` as needing stronger canonical-skill marking. While I prepared this PR, the maintainer independently rewrote that skill on \`main\` to a much shorter form that already includes the canonical claim (\`"v1.8+ canonical loop skill name. It supersedes autonomous-loops"\`). My change to that file is therefore **dropped from this PR** — the maintainer's version solves the same concern more elegantly. No regression.

## Out of scope

- \`visa-doc-translate\` full rewrite. Still FROZEN per \`~/.claude/rules/skill-preferences.md\` until a visa / NAATI domain-safety review lands. Engineering-only edits would not be safe here.

## Test plan

- [ ] Skim each \`SKILL.md\` for accidental content loss vs intent
- [ ] Verify the cross-references all point at currently-existing skills (no dangling pointers)
- [ ] Confirm CSP example block has no \`'unsafe-*'\` defaults
- [ ] Verify naïve in-memory \`RateLimiter\` is gone from \`backend-patterns\`
- [ ] Verify all 11 audit-flagged skills carry a \`> ⚠️ **Drift-prone skill.**\` banner
- [ ] Verify \`search-first\` Step 0 preflight section exists and lists each search channel with a check command + missing-fallback
- [ ] If you have a \`/skill-stocktake\` or \`harness-audit\` workflow, run it against the changes

Backups of every changed file remain in the maintainer's active runtime as \`*.bak\` for quick local revert. Not committed here.

🤖 Curation done with [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/skills/api-design/SKILL.md
+++ b/skills/api-design/SKILL.md
@@ -398,112 +398,20 @@ Accept: application/vnd.myapp.v2+json
 
 ## Implementation Patterns
 
-### TypeScript (Next.js API Route)
-
-```typescript
-import { z } from "zod";
-import { NextRequest, NextResponse } from "next/server";
-
-const createUserSchema = z.object({
-  email: z.string().email(),
-  name: z.string().min(1).max(100),
-});
-
-export async function POST(req: NextRequest) {
-  const body = await req.json();
-  const parsed = createUserSchema.safeParse(body);
-
-  if (!parsed.success) {
-    return NextResponse.json({
-      error: {
-        code: "validation_error",
-        message: "Request validation failed",
-        details: parsed.error.issues.map(i => ({
-          field: i.path.join("."),
-          message: i.message,
-          code: i.code,
-        })),
-      },
-    }, { status: 422 });
-  }
-
-  const user = await createUser(parsed.data);
-
-  return NextResponse.json(
-    { data: user },
-    {
-      status: 201,
-      headers: { Location: `/api/v1/users/${user.id}` },
-    },
-  );
-}
-```
-
-### Python (Django REST Framework)
-
-```python
-from rest_framework import serializers, viewsets, status
-from rest_framework.response import Response
-
-class CreateUserSerializer(serializers.Serializer):
-    email = serializers.EmailField()
-    name = serializers.CharField(max_length=100)
-
-class UserSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = User
-        fields = ["id", "email", "name", "created_at"]
-
-class UserViewSet(viewsets.ModelViewSet):
-    serializer_class = UserSerializer
-    permission_classes = [IsAuthenticated]
-
-    def get_serializer_class(self):
-        if self.action == "create":
-            return CreateUserSerializer
-        return UserSerializer
-
-    def create(self, request):
-        serializer = CreateUserSerializer(data=request.data)
-        serializer.is_valid(raise_exception=True)
-        user = UserService.create(**serializer.validated_data)
-        return Response(
-            {"data": UserSerializer(user).data},
-            status=status.HTTP_201_CREATED,
-            headers={"Location": f"/api/v1/users/{user.id}"},
-        )
-```
-
-### Go (net/http)
-
-```go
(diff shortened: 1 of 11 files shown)


affaan-m:
Thanks for the skill curation work. I ported the focused, low-risk parts into maintainer PR #1723 and merged it after validation: drift-prone warnings, search-first preflight/current agent naming, CSP hardening, and removal of the production-unsafe in-memory rate limiter example. I’m closing this broader PR as superseded because the full 11-skill rewrite still had unresolved cross-reference/review issues and needs a separate dedicated pass.

PR #1889, opened 2026-05-14
fix(hooks): close subshell bypass in pre-bash-dev-server-block

Jamkris:
## Summary

`scripts/hooks/pre-bash-dev-server-block.js` is bypassable today. The hook runs its leading-command and `DEV_PATTERN` check only against the top-level segments returned by `splitShellSegments`, which intentionally does not split on `$(...)`, backticks, or plain `(...)`. That leaves four documented bypass forms that all spawn a dev server while the hook returns exit `0` (allow):

| Form | Block before fix? |
| --- | --- |
| `npm run dev` | ✅ blocked |
| `$(npm run dev)` | ❌ allow |
| `` `npm run dev` `` | ❌ allow |
| `echo $(npm run dev)` | ❌ allow |
| `(npm run dev)` | ❌ allow |
| `$(echo a; npm run dev)` | ❌ allow |

Reproduced on `main` by piping a synthetic PreToolUse payload into the hook for each row.

## Fix

A small BFS walks the command and harvests bodies from `$(...)`, backticks, and plain `(...)` subshell groups, then splits each harvested body through `splitShellSegments` and feeds the result back into the existing `isBlockedDevSegment` check. The leading-command and `DEV_PATTERN` rule is **unchanged** — only the set of segments it runs against grew.

Two new exports live in `scripts/lib/shell-substitution.js`:

- `extractCommandSubstitutions(input)` — the same function originally introduced in `scripts/hooks/gateguard-fact-force.js` via PR #1853 round 2 (`25ea9a77`), extracted to a shared lib so multiple PreToolUse hooks can reuse the quote-aware parser without duplicating it. Body unchanged.
- `extractSubshellGroups(input)` — new sibling for plain `(...)` groups. Honors single-quote literal and double-quote literal-for-parens semantics (bash does not treat `(...)` inside double quotes as a subshell — only `$(...)` is honored there). Skips `$(...)` and backtick spans so we don't double-extract bodies the other helper already covers.

Both helpers are quote-aware. Single- and double-quoted forms remain untouched, so legitimate uses like `git commit -m '(npm run dev)'` or `echo "(npm run dev)"` continue to pass.

## Tests

`tests/hooks/pre-bash-dev-server-block.test.js` extended with 10 cases — 6 new bypass blocks and 4 new allow cases that explicitly prove the fix doesn't over-block:

```
✓ blocks $(npm run dev) — command substitution
✓ blocks `npm run dev` — backtick substitution
✓ blocks echo $(npm run dev) — substitution nested in argument
✓ blocks (npm run dev) — plain subshell group
✓ blocks $(echo a; npm run dev) — substitution with sequenced segments
✓ blocks (pnpm dev) — plain subshell group, alt package manager
✓ allows tmux launcher inside subshell wrapping
✓ allows single-quoted '(npm run dev)' — literal string
✓ allows double-quoted "(npm run dev)" — literal in double quotes
✓ allows single-quoted '$(npm run dev) fix' — literal in single quotes
```

`yarn lint` and `yarn test` both green locally.

## Commit layout

Split into four small commits to make review easier — each commit is independently sensible:

1. `feat(lib): extract shell command-substitution parser to shared lib` — pure extraction of `extractCommandSubstitutions` from `gateguard-fact-force.js` into `scripts/lib/shell-substitution.js`. No behavior change. `gateguard-fact-force.js` keeps its private copy for this PR; consolidating that call site onto the shared lib is a follow-up worth doing once this lands.
2. `feat(lib): add extractSubshellGroups for plain (...) subshells` — new sibling extractor. No consumer yet.
3. `fix(hooks): close subshell bypass in pre-bash-dev-server-block` — wires both helpers into the hook. This is the behavior change.
4. `test(hooks): regression coverage for dev-server-block subshell bypass` — locks in the new blocks plus the explicit allow-cases for quote literals.

## Known limitation (not fixed here)

`eval "$(echo npm run dev)"` is **not** caught by this PR. The substitution body's leading command is `echo`, and statically modeling echo's stdout to recover the executed command is out of scope. The same class affects `gateguard-fact-force.js` (e.g. `eval "$(echo rm -rf /)"`), so it's worth addressing in both hooks together as a follow-up rather than bolting an echo-special-case onto this one.

## Provenance

`extractCommandSubstitutions` originally landed via PR #1853 (`fix: harden GateGuard destructive bash tokenizer`, commit `25ea9a77`). This PR reuses that parser verbatim — extracting it to a shared lib is the only refactor — and adds a sibling for the `(...)` case that gateguard's threat model didn't need.

## Test plan

- [ ] CI lint passes
- [ ] CI tests pass
- [ ] `node tests/hooks/pre-bash-dev-server-block.test.js` — 20/20 green
- [ ] Manual: every row of the bypass table above now returns exit `2`
- [ ] Manual: every row of the allow-cases test now returns exit `0`

---
## Summary by cubic
Blocks dev server invocations wrapped in substitutions, subshells, or brace groups, with quote-aware parsing that closes previously missed bypasses; also expands detection to cover `yarn run dev` and `bun dev`. Tmux launches remain allowed.

- **Bug Fixes**
  - Scan bodies of `$(...)`, backticks, and plain `(...)` recursively and run the existing leading-command + dev-pattern check on each.
  - Handle quotes inside those bodies so `)` in strings doesn’t end groups early (e.g. blocks `$(echo ")"; (npm run dev))`).
  - Detect dev commands inside brace groups by skipping `{`/`}` when finding the leading command (blocks `{ npm run dev; }` and similar).
  - Expand the dev regex to include `yarn(?: run)? dev` and `bun(?: run)? dev`.
  - Preserves allow cases: tmux launchers and quoted literals.
  - Known limitation: `eval "$(echo npm run dev)"` still allowed; follow-up planned.

- **Refactors**
  - Added `scripts/lib/shell-substitution.js` with `extractCommandSubstitutions` and `extractSubshellGroups`; hook uses these helpers to collect nested segments (tmux rule unchanged).

<sup>Written for commit 85d33748e08e5c4a0dcffd88c26e175cd674b33a. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of dev-server invocations to catch attempts nested inside command substitutions, backticks, and subshell/grouping forms.

* **Refactor**
  * Enhanced command-line parsing logic to more thoroughly analyze nested shell constructs and avoid false negatives.

* **Tests**
  * Added comprehensive regression tests covering nested/substituted, quoted, and edge-case command patterns to ensure blocking remains effective.

[Review Change Stack](https://app.coderabbit.ai/change-stack/affaan-m/everything-claude-code/pull/1889)

### DIFF
--- a/scripts/hooks/pre-bash-dev-server-block.js
+++ b/scripts/hooks/pre-bash-dev-server-block.js
@@ -127,6 +127,8 @@ function getLeadingCommandWord(segment) {
       continue;
     }
 
+    if (token === '{' || token === '}') continue;
+
     if (/^[A-Za-z_][A-Za-z0-9_]*=.*/.test(token)) continue;
 
     const normalizedToken = normalizeCommandWord(token);
@@ -159,7 +161,7 @@ process.stdin.on('data', chunk => {
 });
 
 const TMUX_LAUNCHER = /^\s*tmux\s+(new|new-session|new-window|split-window)\b/;
-const DEV_PATTERN = /\b(npm\s+run\s+dev|pnpm(?:\s+run)?\s+dev|yarn\s+dev|bun\s+run\s+dev)\b/;
+const DEV_PATTERN = /\b(npm\s+run\s+dev|pnpm(?:\s+run)?\s+dev|yarn(?:\s+run)?\s+dev|bun(?:\s+run)?\s+dev)\b/;
 
 /**
  * Collect every command-line segment we should evaluate. Returns the top-level
--- a/scripts/hooks/pre-bash-dev-server-block.js
+++ b/scripts/hooks/pre-bash-dev-server-block.js
@@ -4,6 +4,10 @@
 const MAX_STDIN = 1024 * 1024;
 const path = require('path');
 const { splitShellSegments } = require('../lib/shell-split');
+const {
+  extractCommandSubstitutions,
+  extractSubshellGroups
+} = require('../lib/shell-substitution');
 
 const DEV_COMMAND_WORDS = new Set([
   'npm',
@@ -154,23 +158,55 @@ process.stdin.on('data', chunk => {
   }
 });
 
+const TMUX_LAUNCHER = /^\s*tmux\s+(new|new-session|new-window|split-window)\b/;
+const DEV_PATTERN = /\b(npm\s+run\s+dev|pnpm(?:\s+run)?\s+dev|yarn\s+dev|bun\s+run\s+dev)\b/;
+
+/**
+ * Collect every command-line segment we should evaluate. Returns the top-level
+ * segments first, then segments harvested from `$(...)` / backtick command
+ * substitutions and plain `(...)` subshell groups, recursively.
+ *
+ * Without this expansion the leading-command and dev-pattern check below only
+ * sees the outermost command, so wrappers like `$(npm run dev)` and
+ * `(npm run dev)` (which still spawn a dev server) sneak past.
+ */
+function collectCheckSegments(cmd) {
+  const segments = [...splitShellSegments(cmd)];
+  const queue = [cmd];
+  const seen = new Set();
+
+  while (queue.length) {
+    const current = queue.shift();
+    if (seen.has(current)) continue;
+    seen.add(current);
+
+    for (const body of extractCommandSubstitutions(current)) {
+      for (const seg of splitShellSegments(body)) segments.push(seg);
+      queue.push(body);
+    }
+    for (const body of extractSubshellGroups(current)) {
+      for (const seg of splitShellSegments(body)) segments.push(seg);
+      queue.push(body);
+    }
+  }
+
+  return segments;
+}
+
+function isBlockedDevSegment(segment) {
+  const commandWord = getLeadingCommandWord(segment);
+  if (!commandWord || !DEV_COMMAND_WORDS.has(commandWord)) return false;
+  return DEV_PATTERN.test(segment) && !TMUX_LAUNCHER.test(segment);
+}
+
 process.stdin.on('end', () => {
   try {
     const input = JSON.parse(raw);
     const cmd = String(input.tool_input?.command || '');
 
     if (process.platform !== 'win32') {
-      const segments = splitShellSegments(cmd);
-      const tmuxLauncher = /^\s*tmux\s+(new|new-session|new-window|split-window)\b/;
-      const devPattern = /\b(npm\s+run\s+dev|pnpm(?:\s+run)?\s+dev|yarn\s+dev|bun\s+run\s+dev)\b/;
-
-      const hasBlockedDev = segments.some(segment => {
-        const commandWord = getLeadingCommandWord(segment);
-        if (!commandWord || !DEV_COMMAND_WORDS.has(commandWord)) {
-          return false;
-        }
-        return devPattern.test(segment) && !tmuxLauncher.test(segment);
-      });
+      const segments = collectCheckSegments(cmd);
+      const hasBlockedDev = segments.some(isBlockedDevSegment);
 
       if (hasBlockedDev) {
         console.error('[Hook] BLOCKED: Dev server must run in tmux for log access');
--- a/scripts/lib/shell-substitution.js
+++ b/scripts/lib/shell-substitution.js
@@ -106,4 +106,114 @@ function extractCommandSubstitutions(input) {
   return substitutions;
 }
 
-module.exports = { extractCommandSubstitutions };
+/**
+ * Extract bodies of plain `(...)` subshell groups.
+ *
+ * Bash treats `(npm run dev)` as a subshell that executes its contents, but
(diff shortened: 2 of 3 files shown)


Jamkris:
Thanks for the review @greptile-apps, @coderabbitai. Pushed three commits (`70b86d8` → `e2eaf4a` → `85d3374`) addressing every actionable finding. Status below.

---

### @greptile-apps — quote-state gap in `$()` depth counters (CRITICAL — fixed)

> The two `$(...)` depth-counting loops both ignore quote state. A `")"` inside the substitution causes premature depth-zero exit; the orphaned `"` toggles `inDouble` in the outer loop, masking the subsequent `(npm run dev)` from extraction.

Reproduced before the fix:
```
$ echo '{"tool_input":{"command":"$(echo \")\"; (npm run dev))"}}' \
    | node scripts/hooks/pre-bash-dev-server-block.js; echo $?
0   # ← bypass
```

Fix: `70b86d8` — every `$(...)` and `(...)` body loop in `scripts/lib/shell-substitution.js` now tracks its own single- and double-quote state, and only treats `(`/`)` as depth delimiters when outside quotes. After the fix, the payload above exits `2`. Symmetric form `$(echo "(npm run dev)")` correctly remains allowed because bash does not honor `(...)` inside double quotes.

Regression tests in `85d3374`:
- `blocks $(echo ")"; (npm run dev)) — quoted ) does not terminate $() early`
- `blocks (echo ")"; npm run dev) — quoted ) does not terminate (...) early`
- `allows $(echo "(npm run dev)") — () inside double-quoted substitution body is literal`

### @greptile-apps — brace groups (fixed)

> Bash brace groups `{ npm run dev; }` are not handled by either extractor; `splitShellSegments` yields `{ npm run dev` whose leading command word `{` is absent from `DEV_COMMAND_WORDS`.

Fix: `e2eaf4a` — `getLeadingCommandWord` now treats bare `{` and `}` as no-op tokens (matches how shell itself parses brace groups: the braces are reserved words, not commands). The dev-pattern check then runs against the actual command word inside.

Bash requires a space after `{` and a terminator before `}` for an actual group, so `{npm run dev}` correctly remains allowed (single token `{npm`, not in `DEV_COMMAND_WORDS`). That allow case is now also explicit in the test suite.

Regression tests in `85d3374`:
- `blocks { npm run dev; }`
- `blocks echo hi && { npm run dev; }`
- `allows {npm run dev}`

---

### @coderabbitai — `yarn run dev` / `bun dev` not covered (actionable — fixed)

> The regex `DEV_PATTERN` currently misses the shorthand forms "yarn run dev" and "bun dev".

Confirmed before the fix:
```
ALLOW    yarn run dev   # yarn 1.x convention
ALLOW    bun dev        # bun bare form
```

Fix: `e2eaf4a` — `yarn(?:\s+run)?\s+dev` and `bun(?:\s+run)?\s+dev`. Same shape `pnpm(?:\s+run)?\s+dev` was already using. Regression tests in `85d3374`.

### @coderabbitai — symmetric `"$(npm run dev)"` test (nitpick — fixed)

Added in `85d3374`:
- `blocks echo "$(npm run dev)" — double-quoted substitution still substitutes`

### @coderabbitai — BFS redundancy (nitpick `💤 low value` — deferring, with reason)

You're right that the extractors recurse internally, so the BFS does double-process bodies. Two reasons I'm leaving the structure as-is for now:

1. The BFS exists for **cross-syntax** discovery — `(...)` inside `$(...)` and vice versa. Each extractor only recurses into its own syntax. If I remove `queue.push(body)`, I lose the case where `extractCommandSubstitutions` returns a body containing a plain `(...)`, and that body is never fed to `extractSubshellGroups`. The `seen` set already bounds the cost to O(unique-bodies), which for any realistic command is small.
2. Restructuring to "make the extractors non-recursive and rely solely on the BFS" is a follow-up worth doing alongside the planned consolidation of `gateguard-fact-force.js`'s private `extractCommandSubstitutions` copy onto this shared lib. Touching both at once keeps the diff and the test surface coherent. I'd rather do that as a focused follow-up than slip it into this review round.

Open to revisit if you'd prefer the restructure here.

---

### Pre-merge checks — docstring coverage warning

The new helpers in `scripts/hooks/pre-bash-dev-server-block.js` (`collectCheckSegments`, `isBlockedDevSegment`) and the new lib export `extractSubshellGroups` all carry JSDoc. The dip is on small functions like `normalizeCommandWord` / `shouldSkipOptionValue` that existed pre-PR. Happy to add docstrings to those if you want the threshold met in this PR; otherwise I'll leave it as a non-blocking warning.

---

Local: `yarn lint` clean, `yarn test` 29/29 on the dev-server-block suite + the rest of the suite still green. CI should rerun on push.

cc @affaan-m for awareness.

affaan-m:
Thanks for the contribution. I integrated this into the maintainer batch in #1905 so it could run against current `main` with the full required CI matrix. Closing this PR to keep the queue clean; your change is preserved in merge commit 375d750 and credited in the batch.

PR #1971, opened 2026-05-18
fix(hooks): fix cost-reporting bugs in ecc-metrics-bridge (double-count + tail truncation)

Jamkris:
## Summary

`scripts/hooks/ecc-metrics-bridge.js#readSessionCost` has two opposite-sign bugs that both produce a wrong cost number, then feed it into `ecc-context-monitor.js` and from there into the live model turn as `additionalContext`:

1. **Over-count** — sum of cumulative rows reports ~(N+1)/2 × the real cost (~2× over-count at 3 Stop events for a session). Threshold alarms (`COST_NOTICE_USD`, `COST_WARNING_USD`, `COST_CRITICAL_USD`) trip on phantom spend, and the agent itself is told a wrong number.
2. **Under-count** — 8 KiB tail-only read silently drops the target session's row once newer rows from other sessions push past the window; reports zero. On any non-trivial workstation this happens within minutes of normal use because `cost-tracker.js` has no rotation policy.

Both bugs reproduce in a few lines of fixture.

## Bug 1 — Over-count (cumulative double-counting)

`cost-tracker.js`'s module docblock explicitly states:

> Cumulative behavior: Stop fires per assistant response, not per session. Each row therefore represents the cumulative session total up to that point. To get per-session cost, take the last row per session_id.

`readSessionCost` then violates that contract by summing across rows. For 3 cumulative rows at 0.01, 0.02, 0.03 USD (true running total: 0.03), the bridge today reports 0.06 USD.

Repro on `main` at `b113edac`:

```
$ python3 -c '
import json
with open("/tmp/eccc/.claude/metrics/costs.jsonl", "w") as f:
    for cost, inp, out in [(0.01, 333, 166), (0.02, 666, 333), (0.03, 1000, 500)]:
        json.dump({"session_id": "S1", "estimated_cost_usd": cost,
                   "input_tokens": inp, "output_tokens": out}, f)
        f.write(chr(10))'
$ HOME=/tmp/eccc node -e 'const m = require("./scripts/hooks/ecc-metrics-bridge.js"); \
    console.log(JSON.stringify(m.readSessionCost("S1")))'
{"totalCost":0.06,"totalIn":1999,"totalOut":999}
```

Expected: `{"totalCost":0.03,"totalIn":1000,"totalOut":500}` (the last cumulative row).
Actual: 2× over-count.

Fix: replace `+=` with `=` in the matching branch so the assignment reflects the most recent row encountered. Iteration is in file order, which is event time order, so the last assignment wins — exactly the contract `cost-tracker.js` writes against.

## Bug 2 — Under-count (8 KiB tail truncation)

The function reads only the trailing 8 KiB of `~/.claude/metrics/costs.jsonl`. Once the target session's most recent cumulative row gets pushed past that window by newer rows from other sessions, the bridge silently sees zero matching rows and returns `{totalCost: 0}`. Same false signal to `ecc-context-monitor.js`, opposite sign.

Repro (S1 row at file start, 200 rows of OTHER-session noise after, total ~16 KiB):

```
$ HOME=/tmp/eccc node -e '
    const fs = require("fs");
    fs.mkdirSync("/tmp/eccc/.claude/metrics", { recursive: true });
    const s1 = JSON.stringify({session_id:"S1",estimated_cost_usd:0.5,input_tokens:500,output_tokens:250});
    const other = JSON.stringify({session_id:"OTHER",estimated_cost_usd:1,input_tokens:100,output_tokens:50});
    fs.writeFileSync("/tmp/eccc/.claude/metrics/costs.jsonl",
        [s1, ...Array(200).fill(other)].join("\\n") + "\\n");
    const m = require("./scripts/hooks/ecc-metrics-bridge.js");
    console.log(JSON.stringify(m.readSessionCost("S1")))'
{"totalCost":0,"totalIn":0,"totalOut":0}
```

Expected: `{"totalCost":0.5, "totalIn":500, "totalOut":250}` (the S1 row that exists in the file).
Actual: zero — the row is past the 8 KiB tail.

Fix: drop the `fs.openSync` + bounded `fs.readSync` + position arithmetic in favour of `fs.readFileSync(costsPath, 'utf8')` and iterate every line. Each row is ~150 bytes; even 100k rows is ~15 MB and a single sync read on PreToolUse is in the low ms. If file rotation lands in `cost-tracker.js` later, this scan becomes proportionally cheaper.

## Commit layout

Two commits, one fix each, with paired regression tests:

1. `fix(hooks): use last cumulative row for session cost in metrics bridge` — `+=` → `=`. Test: `readSessionCost returns the LAST cumulative row, not the sum (cost-tracker contract)` with three cumulative rows so the two formulas (sum vs. last-row) diverge.
2. `fix(hooks): scan full costs.jsonl when locating session row` — drop 8 KiB ceiling. Test: `readSessionCost finds session row beyond the old 8 KiB tail boundary` with the file deliberately exceeding 8 KiB before the read.

## Tests

`tests/hooks/ecc-metrics-bridge.test.js`: 13 → 15 cases (2 new). Existing `readSessionCost does not include unrelated default-session rows` test happened to pass under both bugs because it used only single-row sessions. The new tests use multi-row + multi-session fixtures so the two bugs become assertable.

Full `yarn test` green; `yarn lint` clean.

## Test plan

- [ ] CI lint passes
- [ ] CI tests pass
- [ ] Manual: bug-1 repro returns `totalCost: 0.03` (was `0.06`)
- [ ] Manual: bug-2 repro returns `totalCost: 0.5` (was `0`)
- [ ] Manual: existing `readSessionCost does not include unrelated default-session rows` test still passes

---
## Summary by cubic
Fix cost reporting in `ecc-metrics-bridge`: use the last cumulative row per session and scan the full `costs.jsonl`. This removes double-counting and tail truncation so thresholds and `additionalContext` use the right numbers, and adds aggregated, deduplicated (persisted) breadcrumbs for malformed lines or read errors.

- **Bug Fixes**
  - Use the last cumulative row per `session_id` in `readSessionCost` (stop summing across rows).
  - Read the full `~/.claude/metrics/costs.jsonl` (drop the 8 KiB tail read).
  - Aggregate and deduplicate stderr breadcrumbs for malformed lines and non-ENOENT read errors; persist dedup across hook subprocesses; stay silent on `ENOENT`.
  - Add regression tests for both bugs, malformed-line aggregation, ENOENT silence, and cross-process dedup; tests and lint pass.

<sup>Written for commit f6d7067ff5c89726d4603dec3b1cc77673ea0d12. Summary will update on new commits. <a href="https://cubic.dev/pr/affaan-m/everything-claude-code/pull/1971?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Use the latest cumulative entry per session (last matching row) when computing session costs; malformed metric lines are skipped and produce a single deduplicated aggregated warning. Missing metric files return zero totals without warnings. Cross-process deduplication ensures repeated stderr warnings are suppressed.

* **Tests**
  * Expanded tests for large files, boundary reads, malformed-entry handling, cross-process suppression, and missing-file behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/affaan-m/everything-claude-code/pull/1971?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/scripts/hooks/ecc-metrics-bridge.js
+++ b/scripts/hooks/ecc-metrics-bridge.js
@@ -91,6 +91,13 @@ function readSessionCost(sessionId) {
       fs.readSync(fd, buf, 0, readSize, Math.max(0, stat.size - readSize));
       const lines = buf.toString('utf8').split('\n').filter(Boolean);
 
+      // Each row in costs.jsonl is *already* a cumulative session total — see
+      // scripts/hooks/cost-tracker.js: "Each row therefore represents the
+      // cumulative session total up to that point. To get per-session cost,
+      // take the last row per session_id." Summing every matching row
+      // therefore double-counts: for N rows of the same session it over-
+      // reports by roughly N(N+1)/2 / N = (N+1)/2 ×. Take the last matching
+      // row instead.
       let totalCost = 0;
       let totalIn = 0;
       let totalOut = 0;
@@ -98,9 +105,9 @@ function readSessionCost(sessionId) {
         try {
           const row = JSON.parse(line);
           if (row.session_id === sessionId) {
-            totalCost += toNumber(row.estimated_cost_usd);
-            totalIn += toNumber(row.input_tokens);
-            totalOut += toNumber(row.output_tokens);
+            totalCost = toNumber(row.estimated_cost_usd);
+            totalIn = toNumber(row.input_tokens);
+            totalOut = toNumber(row.output_tokens);
           }
         } catch {
           /* skip malformed lines */
--- a/scripts/hooks/ecc-metrics-bridge.js
+++ b/scripts/hooks/ecc-metrics-bridge.js
@@ -77,46 +77,44 @@ function extractFilePaths(toolName, toolInput) {
 }
 
 /**
- * Read cumulative cost for a session from the tail of costs.jsonl.
- * Reads last 8KB to avoid scanning entire file.
+ * Read cumulative cost for a session from costs.jsonl.
+ *
+ * Scans the full file because each row is a cumulative session total
+ * (see cost-tracker.js docblock) and the row we need is the last one
+ * matching `sessionId`. The previous implementation read only the
+ * trailing 8 KiB; any session whose latest cumulative row was pushed
+ * past that window by newer rows from other sessions silently dropped
+ * to zero — the opposite sign of the double-count bug fixed in the
+ * previous commit.
+ *
+ * costs.jsonl is append-only and unbounded today (no rotation in
+ * cost-tracker.js). At a typical ~150 bytes per row, even 100k rows
+ * is ~15 MB and a single sync read on every PreToolUse hook is in
+ * the low milliseconds. If rotation lands later, this scan becomes
+ * even cheaper.
  */
 function readSessionCost(sessionId) {
   try {
     const costsPath = path.join(getClaudeDir(), 'metrics', 'costs.jsonl');
-    const stat = fs.statSync(costsPath);
-    const readSize = Math.min(stat.size, 8192);
-    const fd = fs.openSync(costsPath, 'r');
-    try {
-      const buf = Buffer.alloc(readSize);
-      fs.readSync(fd, buf, 0, readSize, Math.max(0, stat.size - readSize));
-      const lines = buf.toString('utf8').split('\n').filter(Boolean);
+    const content = fs.readFileSync(costsPath, 'utf8');
+    const lines = content.split('\n').filter(Boolean);
 
-      // Each row in costs.jsonl is *already* a cumulative session total — see
-      // scripts/hooks/cost-tracker.js: "Each row therefore represents the
-      // cumulative session total up to that point. To get per-session cost,
-      // take the last row per session_id." Summing every matching row
-      // therefore double-counts: for N rows of the same session it over-
-      // reports by roughly N(N+1)/2 / N = (N+1)/2 ×. Take the last matching
-      // row instead.
-      let totalCost = 0;
-      let totalIn = 0;
-      let totalOut = 0;
-      for (const line of lines) {
-        try {
-          const row = JSON.parse(line);
-          if (row.session_id === sessionId) {
-            totalCost = toNumber(row.estimated_cost_usd);
-            totalIn = toNumber(row.input_tokens);
-            totalOut = toNumber(row.output_tokens);
-          }
-        } catch {
-          /* skip malformed lines */
+    let totalCost = 0;
+    let totalIn = 0;
+    let totalOut = 0;
+    for (const line of lines) {
+      try {
+        const row = JSON.parse(line);
+        if (row.session_id === sessionId) {
+          totalCost = toNumber(row.estimated_cost_usd);
+          totalIn = toNumber(row.input_tokens);
+          totalOut = toNumber(row.output_tokens);
         }
+      } catch {
+        /* skip malformed lines */
       }
-      return { totalCost, totalIn, totalOut };
-    } finally {
-      fs.closeSync(fd);
     }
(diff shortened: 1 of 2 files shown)


Jamkris:
Thanks for the review @coderabbitai, @greptile-apps. Pushed two commits (`cd176504` → `8cb9dac8`) addressing both findings.

---

### @coderabbitai — silent `catch` blocks (Major — fixed)

The two `catch` blocks in `readSessionCost` swallowed every failure mode. A malformed `costs.jsonl` row, a permission error opening the file, or any other unexpected I/O failure would silently return zero cost — masking real problems and feeding stale or zero numbers into `ecc-context-monitor.js` (which then injects them as `additionalContext` into the live model turn).

Fix in `cd176504`, both fail-open-preserving:

1. **Inner `JSON.parse` catch** — count malformed lines, write **one aggregated** breadcrumb per call so a log-flooded `costs.jsonl` stays diagnosable without per-line stderr spam:

   ```
   [ecc-metrics-bridge] skipped N malformed line(s) in <path>
   ```

2. **Outer `fs.readFileSync` catch** — write a breadcrumb on real errors, **stay silent on `ENOENT`**. The "no Stop event has fired yet this session" case is genuinely normal and producing per-PostToolUse stderr noise before the first Stop would be more reviewer-visible spam than signal. All other error codes (`EACCES`, `EISDIR`, `EMFILE`, …) get:

   ```
   [ecc-metrics-bridge] failing open after <name> reading <path>: <msg>
   ```

In both cases the function still returns the zero-cost fallback so the bridge never breaks tool execution — only the diagnosability changes.

Two regression tests in `cd176504`:

- `readSessionCost writes a stderr breadcrumb when malformed lines are skipped` — feeds 4 rows (2 valid, 2 malformed), asserts the last valid row still wins AND captured stderr contains `skipped 2 malformed line(s)`.
- `readSessionCost stays silent when costs.jsonl does not exist (ENOENT)` — fresh tmp `HOME` with no metrics dir, asserts zero return AND empty stderr.

### @greptile-apps — PreToolUse / PostToolUse docblock typo (P2 — fixed)

Right — this script is wired in `hooks/hooks.json` as a PostToolUse hook (runs *after* each tool invocation to update the running session aggregate). The previous commit's docblock said "PreToolUse" by mistake. One-word fix in `8cb9dac8`.

---

Local: `node tests/hooks/ecc-metrics-bridge.test.js` 18/18, `yarn test` green, `yarn lint` clean.

cc @affaan-m

affaan-m:
Maintainer follow-up on cubic's P2 finding: pushed `ef13690f` to suppress repeated metrics warning breadcrumbs across a session. `readSessionCost()` now memoizes warning signatures per costs file for malformed-row and non-ENOENT read-error paths, so repeated PostToolUse calls do not flood stderr while still emitting a breadcrumb when the signature changes.

Local validation on the pushed head:
- `node tests/hooks/ecc-metrics-bridge.test.js` -> 18/18 passing
- `node tests/ci/validate-workflow-security.test.js` -> 16/16 passing
- `npm run lint` -> passing
- `npm test` -> 2488/2488 passing

affaan-m:
Reviewer follow-up pushed as `f6d7067f`. I removed the process-local warning cache and replaced it with a tmp-backed warning signature cache keyed by the resolved `costs.jsonl` path, so duplicate malformed-row/read-error breadcrumbs are suppressed across actual hook subprocess invocations. I also moved the `getClaudeDir()`/`costs.jsonl` resolution into the `readSessionCost()` try path and added a subprocess regression for the hook execution model.

Local validation on `f6d7067f`:
- `node tests/hooks/ecc-metrics-bridge.test.js` -> 19/19 passing
- `git diff --check` -> passing
- `npm run lint` -> passing
- `npm test` -> 2488/2488 passing

PR #1983, opened 2026-05-19
fix(lib/hooks): eliminate ENOENT + corruption race in writeBridgeAtomic and writeWarnState

Jamkris:
## Summary

`writeBridgeAtomic` (`scripts/lib/session-bridge.js`) and `writeWarnState` (`scripts/hooks/ecc-context-monitor.js`) wrote to a fixed `${target}.tmp` path before calling `renameSync`. When two processes write to the same session concurrently — e.g. PostToolUse `ecc-metrics-bridge` + the background `ecc-statusline`, or two PostToolUse subprocesses overlapping — the canonical atomic-rename race fires:

1. **A:** `writeFileSync(target.tmp, JSON_A)` — tmp file exists.
2. **B:** `writeFileSync(target.tmp, JSON_B)` — tmp file overwritten.
3. **A:** `renameSync(target.tmp, target)` — succeeds; target = JSON_B (A's payload corrupted).
4. **B:** `renameSync(target.tmp, target)` — throws ENOENT (the rename consumed the file).

Every caller wraps these in `try {} catch {}` so the ENOENT is swallowed and the user-visible symptom is just "the bridge file occasionally contains the wrong process's payload" with no diagnostic.

## Reproduction (before this PR)

Two parallel writers, each calling `writeBridgeAtomic` 500 times against the same session ID:

```
[A] errors=244   # 244 ENOENT exceptions swallowed
[B] errors=248   # ditto
```

That's ~49% of rename calls throwing in each subprocess, all silenced by the caller.

## After this PR

Same workload reports **0 errors** in both subprocesses. Tmp paths no longer collide.

## Fix

Change the fixed suffix to a per-writer unique suffix:

```js
// before
const tmp = `${target}.tmp`;

// after
const tmp = `${target}.${process.pid}.${crypto.randomBytes(4).toString('hex')}.tmp`;
```

Same shape as `writeCostWarningIfChanged` in `scripts/hooks/ecc-metrics-bridge.js` (commit `9b1d8918`). The pid + 4-byte nonce gives each writer process a distinct tmp path, so step 2 above no longer overwrites step 1's payload and step 4 no longer races step 3.

Also added: best-effort `fs.unlinkSync(tmp)` on rename failure so a writer that fails (disk full, permission, EISDIR, etc.) does not leak its tmp file. The original error is still re-thrown.

## Scope clarification

This PR closes the **atomic-rename primitive's race** only:
- ✅ ENOENT on concurrent rename eliminated
- ✅ Tmp file corruption (one writer's payload overwriting another's) eliminated
- ❌ The **read-modify-write race** in callers — two writers each read the same bridge state, increment, and write back, the second clobbering the first — is a separate concern that needs locking or per-writer logs. **Intentionally out of scope** for this PR.

The cost-tracker / metrics-bridge / context-monitor callers tolerate last-writer-wins on their cumulative aggregates today and this PR does not change that contract.

## Commit layout

Three reviewable commits:

1. `fix(lib): use unique tmp suffix in writeBridgeAtomic to eliminate ENOENT race` — the primary primitive. Includes the rename-failure cleanup.
2. `fix(hooks): use unique tmp suffix in writeWarnState (ecc-context-monitor)` — mirror the fix on the companion debounce-state writer. Three call sites in the repo now share the same atomic-write contract.
3. `test(lib): concurrent writeBridgeAtomic + tmp-cleanup regression` — two new tests:
   - spawn two child Node processes calling `writeBridgeAtomic` 200 times each, assert both exit 0 (no ENOENT) and the final file is parseable JSON belonging to one of the writers
   - pre-create a directory at the target path to force `renameSync` failure, assert the call throws AND no tmp file with the writer's `pid.<nonce>.tmp` prefix is left behind in `os.tmpdir()`

## Tests

- `tests/lib/session-bridge.test.js`: 12 → 14 cases
- Full `yarn test` green; `yarn lint` clean

## Risk

Very low. The unique-suffix pattern is the textbook fix for POSIX atomic-rename concurrency and is already used elsewhere in the same file family (`writeCostWarningIfChanged`). Reviewer concern: "doesn't the kernel handle this?" — `renameSync` is atomic *between source and destination*, but `writeFileSync` to a shared source path is not; the unique suffix is required to make the source distinct per-writer.

## Test plan

- [ ] CI lint passes
- [ ] CI tests pass
- [ ] Manual: concurrent-writer reproduction reports 0 ENOENT (was ~49%)
- [ ] Manual: rename-failure cleanup test passes (no orphan tmp files)
- [ ] Manual: existing `roundtrip write then read returns same data` test still passes (single-writer contract unchanged)

---
## Summary by cubic
Fixes the concurrent write race with unique tmp paths and adds cross-platform atomic renames with retries on Windows, eliminating ENOENT/EPERM errors and preventing file corruption. Also cleans up tmp files on failure.

- **Bug Fixes**
  - Use `${target}.${process.pid}.${crypto.randomBytes(4).toString('hex')}.tmp` instead of shared `${target}.tmp` in `scripts/lib/session-bridge.js` and `scripts/hooks/ecc-context-monitor.js`.
  - Add `renameWithRetry(tmp, target)` (5 attempts, 20–320 ms backoff) to handle Windows `EPERM`/`EACCES`/`EBUSY`; used by both writers and exported from `scripts/lib/session-bridge.js`.
  - On rename failure, attempt `fs.unlinkSync(tmp)` and rethrow the original error.

- **Tests**
  - Make the concurrency test truly concurrent and switch `assert.throws` to a regex matcher; confirms no ENOENT/EPERM, final file is valid JSON, and no tmp leaks on rename failure.

<sup>Written for commit 1f6fc9c27dc06c482469d4152ae4e1e82ff4215f. Summary will update on new commits. <a href="https://cubic.dev/pr/affaan-m/everything-claude-code/pull/1983?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented concurrent writers from clobbering each other by switching to unique atomic temp-file writes with safe rename-and-cleanup on failure; added retry handling for transient rename errors to improve reliability.

* **Tests**
  * Added tests that exercise concurrent writes and simulated rename failures to validate correctness and ensure no temporary files are leaked.

[Review Change Stack](https://app.coderabbit.ai/change-stack/affaan-m/everything-claude-code/pull/1983?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/scripts/hooks/ecc-context-monitor.js
+++ b/scripts/hooks/ecc-context-monitor.js
@@ -9,6 +9,7 @@
 
 'use strict';
 
+const crypto = require('crypto');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
@@ -61,15 +62,30 @@ function readWarnState(sessionId) {
 }
 
 /**
- * Write debounce state.
+ * Write debounce state atomically (unique-suffix tmp then rename).
+ *
+ * The tmp path includes `process.pid` plus a random nonce so concurrent
+ * PostToolUse subprocesses writing to the same session's warn-state
+ * file do not clobber each other's tmp mid-write. Without the unique
+ * suffix, two writers race over a shared `${target}.tmp` and produce
+ * either a corrupted payload or an ENOENT throw on the second rename.
+ *
+ * Same pattern as `writeBridgeAtomic` in `scripts/lib/session-bridge.js`
+ * and `writeCostWarningIfChanged` in `scripts/hooks/ecc-metrics-bridge.js`.
+ *
  * @param {string} sessionId
  * @param {object} state
  */
 function writeWarnState(sessionId, state) {
   const target = getWarnPath(sessionId);
-  const tmp = `${target}.tmp`;
+  const tmp = `${target}.${process.pid}.${crypto.randomBytes(4).toString('hex')}.tmp`;
   fs.writeFileSync(tmp, JSON.stringify(state), 'utf8');
-  fs.renameSync(tmp, target);
+  try {
+    fs.renameSync(tmp, target);
+  } catch (err) {
+    try { fs.unlinkSync(tmp); } catch { /* ignore */ }
+    throw err;
+  }
 }
 
 /**
--- a/scripts/hooks/ecc-context-monitor.js
+++ b/scripts/hooks/ecc-context-monitor.js
@@ -13,7 +13,7 @@ const crypto = require('crypto');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
-const { sanitizeSessionId, readBridge } = require('../lib/session-bridge');
+const { sanitizeSessionId, readBridge, renameWithRetry } = require('../lib/session-bridge');
 
 const CONTEXT_WARNING_PCT = 35;
 const CONTEXT_CRITICAL_PCT = 25;
@@ -81,7 +81,7 @@ function writeWarnState(sessionId, state) {
   const tmp = `${target}.${process.pid}.${crypto.randomBytes(4).toString('hex')}.tmp`;
   fs.writeFileSync(tmp, JSON.stringify(state), 'utf8');
   try {
-    fs.renameSync(tmp, target);
+    renameWithRetry(tmp, target);
   } catch (err) {
     try { fs.unlinkSync(tmp); } catch { /* ignore */ }
     throw err;
--- a/scripts/lib/session-bridge.js
+++ b/scripts/lib/session-bridge.js
@@ -8,6 +8,7 @@
  * without scanning large JSONL logs on every invocation.
  */
 
+const crypto = require('crypto');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
@@ -51,15 +52,35 @@ function readBridge(sessionId) {
 }
 
 /**
- * Write bridge data atomically (write .tmp then rename).
+ * Write bridge data atomically (write unique-suffix tmp then rename).
+ *
+ * The tmp path includes `process.pid` plus a random nonce so concurrent
+ * writers (e.g. PostToolUse `ecc-metrics-bridge` and the background
+ * `ecc-statusline`, both writing to the same session bridge) do not
+ * clobber each other's tmp file mid-write. With a fixed `.tmp` suffix
+ * two writers could both call `writeFileSync` against the same path
+ * before either reaches `renameSync`, causing one writer's payload to
+ * silently overwrite the other and the second `renameSync` to throw
+ * ENOENT once the rename consumes the file.
+ *
+ * Same pattern already used by `writeCostWarningIfChanged` in
+ * `scripts/hooks/ecc-metrics-bridge.js` (commit 9b1d8918) for the
+ * cost-warning cache; this commit applies it to the session-bridge
+ * primitive too.
+ *
  * @param {string} sessionId - Already-sanitized session ID
  * @param {object} data
  */
 function writeBridgeAtomic(sessionId, data) {
   const target = getBridgePath(sessionId);
-  const tmp = `${target}.tmp`;
+  const tmp = `${target}.${process.pid}.${crypto.randomBytes(4).toString('hex')}.tmp`;
   fs.writeFileSync(tmp, JSON.stringify(data), 'utf8');
-  fs.renameSync(tmp, target);
+  try {
+    fs.renameSync(tmp, target);
+  } catch (err) {
+    try { fs.unlinkSync(tmp); } catch { /* ignore */ }
+    throw err;
+  }
 }
 
 /**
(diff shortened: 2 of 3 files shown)


Jamkris:
Thanks for the review @greptile-apps, @coderabbitai, @cubic-dev-ai. Pushed `8149beac` addressing both findings.

---

### @greptile-apps P1 + @coderabbitai Major + @cubic-dev-ai P2 — concurrent test ran sequentially (fixed)

Right — `spawnSync` blocks until each child exits, so the two writers ran one-after-the-other and never overlapped. The race the fix targets requires both `writeFileSync` calls to interleave with each other's `renameSync`; with sequential execution it can't happen. The test would have passed against the old broken `${target}.tmp` suffix.

Fix in `8149beac`: introduce a one-off "race runner" helper that runs inside its own subprocess and uses async `spawn` to start both writers simultaneously. The runner waits for both to exit (event loop is local to the runner subprocess, so this stays compatible with the synchronous test harness used elsewhere in this file) and reports both exit codes plus stderrs on stdout. The test then calls the runner via `spawnSync` and parses the result.

Verified: with the fixed `${target}.${pid}.${nonce}.tmp` suffix the test passes; with the old fixed `${target}.tmp` suffix reintroduced the test fails as expected (one writer hits ENOENT on roughly half its rename calls).

### @greptile-apps P2 + @cubic-dev-ai P3 — `assert.throws` second arg is not a matcher (fixed)

Right — the string was silently treated as the assertion failure message (what to print when the function does *not* throw), not as an error matcher. The check passed for any thrown error.

Fix in `8149beac`: pass a regex as the second arg and keep the explanatory text as the third arg. The regex matches `EISDIR | EPERM | ENOTDIR | ENOENT` because `renameSync` of a regular tmp file onto an existing directory raises different codes on Linux / macOS / BSD — keeping the matcher portable across CI runners.

---

Local: `node tests/lib/session-bridge.test.js` 14/14, `yarn test` green, `yarn lint` clean.

The two helper files (`tests/__tmp_bridge_writer.js`, `tests/__tmp_bridge_race_runner.js`) are written and unlinked inside the test's try/finally so they never persist beyond the test run.

cc @affaan-m

Jamkris:
Thanks for the catch — the Windows matrix flagged a real OS-semantics gap that round 1's unique-tmp fix did not cover. Pushed two commits (`bcbd2ace` → `1f6fc9c2`) addressing it.

---

### Windows `MoveFileExW` is not POSIX `rename(2)` (fixed)

The round-1 unique-suffix fix is correct under POSIX: `rename(2)` is atomic between source and destination, so each writer renames its own tmp file onto the same target without conflict. Windows `MoveFileExW` is different: it fails with `EPERM` / `EACCES` / `EBUSY` when the target is currently being renamed by *another* process — a short race window that fires reliably under this hook's PostToolUse + statusline concurrency.

All 9 Windows matrix cells (Node 18 / 20 / 22 × npm / pnpm / yarn) hit the same path:

```
Error: EPERM: operation not permitted, rename
  'C:\…\ecc-metrics-test-bridge-race-….json.9504.4aef575a.tmp' ->
  'C:\…\ecc-metrics-test-bridge-race-….json'
    at writeBridgeAtomic (scripts/lib/session-bridge.js:79:8)
```

POSIX matrices (Linux + macOS) passed unchanged.

### Fix

Two commits:

1. **`bcbd2ace` fix(lib): retry rename on Windows EPERM/EACCES/EBUSY**

   New `renameWithRetry(tmp, target)` helper retries `fs.renameSync` up to 5 times on `EPERM` / `EACCES` / `EBUSY` with exponential backoff (20 ms → 320 ms total). Other error codes (`ENOENT`, `ENOSPC`, `EROFS`, …) re-throw immediately — they are not transient. POSIX runs hit the first try and exit immediately; only Windows takes the retry path.

   Backoff uses `Atomics.wait` on a throwaway `SharedArrayBuffer` so the retry path doesn't busy-spin the CPU. Verified on Node ≥ 17 that main-thread `Atomics.wait` works (50 ms call slept exactly 50 ms). A try/catch fallback to brief busy-wait covers older runtimes where `Atomics.wait` is restricted to workers.

   `writeBridgeAtomic` calls the helper instead of `fs.renameSync` and keeps its existing best-effort tmp cleanup on terminal failure.

   `renameWithRetry` is added to `module.exports` so the companion writer can adopt the same retry policy without duplicating the helper.

2. **`1f6fc9c2` fix(hooks): use shared renameWithRetry in writeWarnState (ecc-context-monitor)**

   Mirror on `writeWarnState`. Imports `renameWithRetry` from `scripts/lib/session-bridge.js` instead of duplicating the helper, so the two atomic-write sites share one retry policy.

### Verification

- Local: `node tests/lib/session-bridge.test.js` 14/14, `yarn test` green, `yarn lint` clean.
- Reasoning for Windows: round 1's race-runner test forces both child writers to overlap; under the new retry, the EPERM the Windows kernel raises during one writer's `MoveFileExW` is caught and re-tried until the other writer's rename releases the target. Maximum total retry budget (≈ 620 ms across all 5 attempts) is well above the typical Windows MoveFileExW contention window.

CI should rerun on push.

cc @affaan-m

affaan-m (approved):
Reviewed current head 1f6fc9c2. The Windows CI failure from concurrent atomic renames is addressed with the shared retry helper, focused local validation passed, and the full GitHub Actions matrix is green.

PR #2059, opened 2026-05-25
fix(commands): resolve active plugin root in /instinct-status (#2037)

### LINKED ISSUE #2037: /instinct-status falls back to manual install when CLAUDE_PLUGIN_ROOT is unset, masking the plugin's actual state
Environment:
  - ECC version: 2.0.0-rc.1 (plugin enabled in ~/.claude/settings.json)
  - Coexisting older manual install at ~/.claude/skills/continuous-learning-v2/

  What happened:

  The /instinct-status slash command template is:

  python3 "${CLAUDE_PLUGIN_ROOT}/skills/continuous-learning-v2/scripts/instinct-cli.py" status

  with a documented manual-install fallback:

  python3 ~/.claude/skills/continuous-learning-v2/scripts/instinct-cli.py status

  In Claude Code v2.1.143 sessions where the ECC plugin is enabled and active (its hooks fire correctly — confirmed via observed PreToolUse behavior), $CLAUDE_PLUGIN_ROOT was nevertheless empty in the slash command's shell context. The
  first command resolved to python3 "/skills/continuous-learning-v2/scripts/instinct-cli.py" (absolute root path), failed, and the || fallback ran the manual-install CLI.

  This meant /instinct-status reported on the manual install's view of the world even though the active hooks/observer were operating from the plugin path, leading to the path-divergence problem documented in (companion) issue #N.

  Expected:

  The slash command should robustly locate the active plugin install — for example using the same lookup logic used by ECC's hook bootstrap (scripts/hooks/plugin-hook-bootstrap.js), which already walks
  ~/.claude/plugins/cache/<marketplace>/<plugin>/<version>/ to find the active root.

  Suggested fix:

  Replace the slash command's two-line template with a Node bootstrap or Python helper that does the same plugin-root resolution the hooks already use:

  PLUGIN_ROOT=$(node -e "...same resolver as plugin-hook-bootstrap.js..." 2>/dev/null)
  python3 "${PLUGIN_ROOT:-$HOME/.claude}/skills/continuous-learning-v2/scripts/instinct-cli.py" status

  OR document explicitly that users with a manual install should remove it once they install ECC as a plugin.

  Workaround applied:

  Copied the plugin's CLI over the manual install (same workaround as the migration issue), so both code paths converge on the new XDG path.

fxdv:
## Summary

Fixes #2037.

The `/instinct-status` slash command template expanded `${CLAUDE_PLUGIN_ROOT}` directly and documented a `python3 ~/.claude/skills/continuous-learning-v2/scripts/instinct-cli.py` fallback for manual installs. When users had both an active plugin install (under `~/.claude/plugins/cache/<slug>/<org>/<version>/`) AND a legacy `~/.claude/skills/continuous-learning-v2/` directory left over from a previous manual install, an empty `CLAUDE_PLUGIN_ROOT` (which Claude Code does not always populate in slash-command shell contexts, per the issue repro) silently made the command read the stale legacy install while the active plugin hooks and observer wrote to the new XDG path.

End result, as the reporter described: a fully functional, actively learning system that appeared completely broken to the user — "No instincts found" while observations were accumulating elsewhere.

## Fix

Replace the brittle two-block template with the same inline-resolver pattern that `hooks/hooks.json` and the other slash commands (`/sessions`, `/skill-health`) already use:

```bash
ECC_ROOT="${CLAUDE_PLUGIN_ROOT:-$(node -e "var r=(()=>{...resolver...})();console.log(r)")}"
python3 "$ECC_ROOT/skills/continuous-learning-v2/scripts/instinct-cli.py" status
```

The resolver chain is: env var → standard install → known plugin roots → plugin cache walk → fallback. It's the canonical `INLINE_RESOLVE` constant exported from `scripts/lib/resolve-ecc-root.js` — **no new code is introduced**, just consistent adoption of the existing pattern across all slash commands.

## Files changed

- `commands/instinct-status.md` — canonical
- `.opencode/commands/instinct-status.md`
- `docs/zh-CN/commands/instinct-status.md`
- `docs/ja-JP/commands/instinct-status.md`
- `docs/tr/commands/instinct-status.md`
- `tests/lib/command-plugin-root.test.js` — extends the existing pattern (which already covers `sessions.md` and `skill-health.md`) with an `instinct-status.md` assertion, plus a regression guard that the legacy `~/.claude/skills/...` fallback string is no longer present.

## Test plan

- [x] `node tests/lib/command-plugin-root.test.js` — 4/4 pass (3 prior + 1 new)
- [x] `node tests/lib/resolve-ecc-root.test.js` — 23/23 pass (resolver itself is unchanged; existing tests already cover the env-unset → plugin-cache-walk path that fixes this bug)
- [x] `node tests/hooks/hooks.test.js` — 237/237 pass (sanity)
- [x] Smoke-tested the new bash block locally with `unset CLAUDE_PLUGIN_ROOT`; the resolver returns the correct active install path (falls through to `~/.claude` cleanly when no install is present, which matches the documented final-fallback behavior).

## Notes for reviewer

- All five copies are kept in sync in this PR (canonical + opencode + 3 translations). Surrounding prose in each translation is preserved; only the Implementation section's two code blocks → one code block change.
- The inline resolver blob is the verbatim string from `INLINE_RESOLVE` in `scripts/lib/resolve-ecc-root.js`, so any future tightening of the resolver only needs to update that one constant + the inlined copies in `hooks.json`, `sessions.md`, `skill-health.md`, and now `instinct-status.md` (the existing `tests/lib/command-plugin-root.test.js` already pins this consistency).
- Did not consolidate the inline blob into a shared script file because the current architecture deliberately keeps the resolver inlinable into bash (Node has to find the wrapper before it can `require()` anything from the plugin tree — chicken-and-egg). The pre-minified `INLINE_RESOLVE` export is the maintainer's solution to that constraint and is reused here as-is.

---
## Summary by cubic
Fixes stale path resolution in `/instinct-status` that could read a legacy install when `CLAUDE_PLUGIN_ROOT` was unset, causing “No instincts found.” The command now uses the shared inline resolver to always target the active plugin install (#2037).

- **Bug Fixes**
  - Replaced the two-block template with the canonical inline resolver (env var → standard install → known plugin roots → plugin cache → fallback) from `scripts/lib/resolve-ecc-root.js`.
  - Updated all copies of the command (canonical, `.opencode`, JA/TR/ZH) to use the resolver; polished zh-CN phrasing.
  - Extended tests to assert the resolver is present and the legacy `~/.claude/skills/...` fallback is not used.

<sup>Written for commit 6c99b786ee66733b24ecd372ad841e9503916b9a. Summary will update on new commits. <a href="https://cubic.dev/pr/affaan-m/ECC/pull/2059?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Documentation**
  * Updated instinct-status command docs (EN/JA/TR/ZH) to show a prioritized plugin-root resolution and run the status example using the resolved root for more reliable command examples.

* **Tests**
  * Added tests to validate the resolver is invoked correctly and to prevent regressions that reintroduce legacy hardcoded fallback paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/affaan-m/ECC/pull/2059?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.opencode/commands/instinct-status.md
+++ b/.opencode/commands/instinct-status.md
@@ -9,16 +9,15 @@ Show instinct status from continuous-learning-v2: $ARGUMENTS
 
 ## Your Task
 
-Run:
+Resolve the active ECC plugin root with the same walker `hooks/hooks.json`
+uses (env var → standard install → known plugin roots → plugin cache →
+fallback), then run the instinct CLI. This avoids reading a stale legacy
+`~/.claude/skills/continuous-learning-v2/` install when the plugin is
+active under `~/.claude/plugins/cache/...` (#2037).
 
 ```bash
-python3 "${CLAUDE_PLUGIN_ROOT}/skills/continuous-learning-v2/scripts/instinct-cli.py" status
-```
-
-If `CLAUDE_PLUGIN_ROOT` is unavailable, use:
-
-```bash
-python3 ~/.claude/skills/continuous-learning-v2/scripts/instinct-cli.py status
+ECC_ROOT="${CLAUDE_PLUGIN_ROOT:-$(node -e "var r=(()=>{var e=process.env.CLAUDE_PLUGIN_ROOT;if(e&&e.trim())return e.trim();var p=require('path'),f=require('fs'),h=require('os').homedir(),d=p.join(h,'.claude'),q=p.join('scripts','lib','utils.js');if(f.existsSync(p.join(d,q)))return d;for(var s of [['ecc'],['ecc@ecc'],['marketplaces','ecc'],['everything-claude-code'],['everything-claude-code@everything-claude-code'],['marketplaces','everything-claude-code']]){var l=p.join(d,'plugins',...s);if(f.existsSync(p.join(l,q)))return l}try{for(var g of ['ecc','everything-claude-code']){var b=p.join(d,'plugins','cache',g);for(var o of f.readdirSync(b,{withFileTypes:true})){if(!o.isDirectory())continue;for(var v of f.readdirSync(p.join(b,o.name),{withFileTypes:true})){if(!v.isDirectory())continue;var c=p.join(b,o.name,v.name);if(f.existsSync(p.join(c,q)))return c}}}}catch(x){}return d})();console.log(r)")}"
+python3 "$ECC_ROOT/skills/continuous-learning-v2/scripts/instinct-cli.py" status
 ```
 
 ## Behavior Notes
--- a/commands/instinct-status.md
+++ b/commands/instinct-status.md
@@ -10,16 +10,16 @@ Shows learned instincts for the current project plus global instincts, grouped b
 
 ## Implementation
 
-Run the instinct CLI using the plugin root path:
+Run the instinct CLI, resolving the active ECC plugin root the same way
+`hooks/hooks.json` and the other slash commands (`/sessions`, `/skill-health`)
+do — env var → standard install → known plugin roots → plugin cache → fallback.
+This avoids the divergence that happens when `CLAUDE_PLUGIN_ROOT` is unset
+while a legacy `~/.claude/skills/continuous-learning-v2/` directory still
+exists (#2037).
 
 ```bash
-python3 "${CLAUDE_PLUGIN_ROOT}/skills/continuous-learning-v2/scripts/instinct-cli.py" status
-```
-
-Or if `CLAUDE_PLUGIN_ROOT` is not set (manual installation), use:
-
-```bash
-python3 ~/.claude/skills/continuous-learning-v2/scripts/instinct-cli.py status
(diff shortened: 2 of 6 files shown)


fxdv:
Applied in `6c99b786` — changed `使用与 ... 相同的解析器` to `以与 ... 相同的解析器` so the means/method is introduced by an explicit preposition (`以`) instead of reading as a bare verb-object construction. Thanks for the LanguageTool catch.

affaan-m:
This is the right follow-up for #2037. `/instinct-status` should resolve the active plugin root explicitly instead of silently falling back to a manual install path.

Routing as a stable patch candidate after local command-root regression checks.

alexraphael07-prog:

## 🔍 PR Review for #2059

### 1. **Summary**
The PR fixes #2037 by replacing naive `${CLAUDE_PLUGIN_ROOT}` expansion in `/instinct-status` with a hardened resolver that follows the same lookup hierarchy used by other parts of the codebase: env var → standard install → known plugin roots → cache → fallback. This prevents the slash command from accidentally running a stale manual install when the plugin is active.

### 2. **Issues**
- **Bash injection risk** – the Node one-liner interpolates `home

PR #2121, opened 2026-06-02
fix(hooks): stop false loop warnings and repeated identical context warnings

### LINKED ISSUE #2120: Context monitor: false "stuck loop" warnings + the same warning repeated ~20× in one turn
## Summary

While working through one long session I hit two separate problems with the PostToolUse monitor hooks (`ecc-metrics-bridge.js` + `ecc-context-monitor.js`). Both are noise problems — they don't break anything, but they make the warnings untrustworthy, which is worse than no warning.

## Problem 1 — false "stuck loop" warnings on normal editing

**What I saw:** `LOOP WARNING: Tool 'Edit' called 3 times with same parameters` fired repeatedly while I was making *different* edits to the same file (e.g. editing several distinct spots in one source file during a feature).

**Why it happens:** `hashToolCall` in `ecc-metrics-bridge.js` fingerprints a tool call to detect loops. For `Edit`/`Write`/`MultiEdit` it builds that fingerprint from the **`file_path` only** — it ignores `old_string`/`new_string`/`content`. So every edit to the same file produces the **same hash**, and three edits to one file trip the loop detector even though each edit was completely different.

```js
} else if (toolInput?.file_path) {
  key = String(toolInput.file_path);   // <-- same file -> same hash, regardless of the change
}
```

## Problem 2 — the same warning prints ~20 times in one turn

**What I saw:** a single cost warning (`COST WARNING: Session cost is $15.56…`) printed the **identical** line ~20 times across one turn, then jumped to a new value and repeated again.

**Why it happens (two parts):**
1. The cost number comes from `costs.jsonl`, which is written by the **`cost-tracker` Stop hook** — i.e. once per *turn*. But `ecc-context-monitor` reads it on **every PostToolUse** (every tool call). So within a single long turn the figure can't change — it looks "frozen," then jumps at the next turn boundary. That part is expected given the data source.
2. The debounce in `ecc-context-monitor.js` re-emits a warning every `DEBOUNCE_CALLS` (5) tool calls **even when the warning text is unchanged**. Combined with the frozen-within-a-turn cost, that's the same stale line ~20 times.

## Suggested fix

1. **Loop hash:** include the edit content (`old_string`/`new_string`/`content`/`edits`) in the fingerprint for `Edit`/`Write`/`MultiEdit`/`NotebookEdit`, so distinct edits to the same file hash differently while identical edits still collide.
2. **Debounce:** dedupe on **message content** rather than a call counter — surface a warning only when its text changes (cost moved, new file count, new loop) or on first escalation to critical; otherwise suppress it.

I have a fix + regression tests ready and will open a PR referencing this issue.

ahnineamine:
Closes #2120.

## What

Two PostToolUse monitor noise defects (see #2120 for the plain-language writeup):

1. **False "stuck loop" warnings** — `hashToolCall` (`ecc-metrics-bridge.js`) fingerprinted `Edit`/`Write`/`MultiEdit` on `file_path` only, so distinct edits to the same file collided and tripped the loop detector. Now the hash includes the edit content (`old_string`/`new_string`/`content`/`edits`) for `Edit`/`MultiEdit`/`Write`/`NotebookEdit`; identical edits still collide as intended.
2. **Same warning repeated ~20× per turn** — `ecc-context-monitor.js` re-emitted a warning every `DEBOUNCE_CALLS` (5) tool calls even when unchanged. Because the cost figure only refreshes at Stop (turn) boundaries, one stale value printed the identical line ~20 times. Now dedupe on **message content**: surface a warning only when its text changes or on first escalation to critical; otherwise suppress. (`DEBOUNCE_CALLS` removed as now-unused.)

## Tests

- Added regression tests in `tests/hooks/ecc-metrics-bridge.test.js`: distinct edits to the same file hash differently; identical edits still match.
- `node tests/hooks/ecc-metrics-bridge.test.js` → 21 passed; `node tests/hooks/ecc-context-monitor.test.js` → 20 passed.
- Manually verified the dedupe end-to-end: first occurrence emits, identical repeats are suppressed, a changed cost re-emits.

## Risk / rollback

Hook-only change, fail-open preserved (any throw still returns input unchanged). No behavior change to genuine loops or to the first appearance of any warning. Revert the single commit to roll back.

---
## Summary by cubic
Reduced noisy PostToolUse warnings by preventing false "stuck loop" alerts and deduping repeated, identical context warnings. This cuts warning spam while keeping new or escalated issues visible.

- **Bug Fixes**
  - Loop detector: `hashToolCall` now fingerprints the full serialized edit payload (`old_string`/`new_string`/`content`/`edits`) for `Edit`/`MultiEdit`/`Write`/`NotebookEdit` and truncates the digest, so different edits—even past 2k chars—to the same file don’t collide; identical edits still collide as intended.
  - Context monitor: dedupe on warning message text and only emit on change or first escalation to critical; clear dedupe state when warnings resolve so later recurrences surface; removed `DEBOUNCE_CALLS`.
  - Tests: added regression coverage for distinct vs identical edit hashes and for large edits diverging after 2048 chars.

<sup>Written for commit 45d94aa9a26f228ecf25db66366469dd5f59e21e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/affaan-m/ECC/pull/2121?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced duplicate warning notifications by switching to message-based deduplication and clearing dedupe state when no warnings are present
  * Improved edit-loop detection to correctly distinguish different edits on the same file, preventing false loop triggers

* **Tests**
  * Added test coverage to ensure edit-operation hashing is deterministic and sensitive to content differences

### DIFF
--- a/scripts/hooks/ecc-context-monitor.js
+++ b/scripts/hooks/ecc-context-monitor.js
@@ -23,7 +23,6 @@ const COST_CRITICAL_USD = 50;
 const FILES_WARNING_COUNT = 20;
 const LOOP_THRESHOLD = 3;
 const STALE_SECONDS = 60;
-const DEBOUNCE_CALLS = 5;
 
 function isEnabledEnv(value, defaultValue = true) {
   if (value === undefined || value === null || String(value).trim() === '') {
@@ -57,7 +56,7 @@ function readWarnState(sessionId) {
   try {
     return JSON.parse(fs.readFileSync(getWarnPath(sessionId), 'utf8'));
   } catch {
-    return { callsSinceWarn: 0, lastSeverity: null };
+    return { callsSinceWarn: 0, lastSeverity: null, lastMessage: null };
   }
 }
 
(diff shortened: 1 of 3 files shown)


ManuMorem7 (changes requested):
<sub>Requested changes in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/affaan-m/ECC/pull/2121)</sub>

ManuMorem7 (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/affaan-m/ECC/pull/2121)</sub>

ManuMorem7 (approved):

codeCraft-Ritik (commented):
Nice fix! Addressing both the false loop detections and the repeated warning spam should make the monitoring output much more actionable. Including edit content in the fingerprint seems like the correct way to distinguish genuinely different edits to the same file.

PR #2161, opened 2026-06-05
feat(gateguard): add env knobs for routine bash gate + extra destructive patterns

### LINKED ISSUE #2078: Feature request: surface gateguard fact_force_bash_routine / destructive_bash_extra config knobs in the JS port
### Problem

The JS port of gateguard shipped in ECC (`scripts/hooks/gateguard-fact-force.js`) treats every Bash invocation as fact-forcing material — both destructive (`rm -rf`, `git reset --hard`) and routine (`ls`, `grep`, `python script.py`). The Python upstream `gateguard-ai` already separates these via `.gateguard.yml`:

```yaml
gates:
  fact_force_bash_destructive: true   # keep gate
  fact_force_bash_routine: false      # disable routine tax
destructive_bash_extra:
  - "supabase db reset"
  - "prisma migrate reset"
```

The JS port's own docstring acknowledges the gap (`scripts/hooks/gateguard-fact-force.js:19`):

> *"Full package with config support: pip install gateguard-ai"*

But anyone using ECC via the plugin marketplace never installs the Python package — they get the JS port and the hardcoded behavior, with no way to scope down to destructive-only without editing the bundled JS.

### Why this matters

In a long-running session, the routine-bash gate is a one-time tax (fires once per session). It's still friction:

- It surfaces on the first `ls` / `grep` / `npm install` regardless of intent.
- The fact-forcing template asks the LLM to restate context that's already in the conversation — net-neutral or net-negative for the model's next move.
- The destructive gate (`rm -rf`, `DROP TABLE`, `--force` pushes, `sudo`) is the high-value half. That one we want to keep.

Operators who run into this end up patching the JS port locally to short-circuit the routine path. Below is the patch I'm currently maintaining against ECC v1.10.0 (it survives a cache rebuild via a restore script in our repo, but every plugin update clobbers it). I'd much rather drop the patch in favor of a config knob.

### Proposed resolution

One of:

1. **Read `.gateguard.yml`** from the CWD in the JS port the same way the Python package does, honoring at minimum:
   - `gates.fact_force_bash_destructive` (default `true`)
   - `gates.fact_force_bash_routine` (default `true` to preserve current behavior)
   - `destructive_bash_extra: [str, ...]`
2. **Or** at minimum, expose two env vars that the JS port checks before firing:
   - `GATEGUARD_BASH_ROUTINE_DISABLED=1` — skip the routine gate
   - `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` — pipe-separated regex additions
3. **Or** a `.gateguard.json` next to the JS port that the hook reads at startup.

Option 1 is the cleanest — it makes the JS port a true port of the Python config surface. Option 2 is the smallest patch and would unblock us today.

#### Option 2 sketch (10-line diff against `scripts/hooks/gateguard-fact-force.js`)

```diff
-const DESTRUCTIVE_BASH = /\b(rm\s+-rf|git\s+reset\s+--hard|git\s+checkout\s+--|git\s+clean\s+-f|drop\s+table|delete\s+from|truncate|git\s+push\s+--force|dd\s+if=)\b/i;
+const _BASE_DESTRUCTIVE = '\\b(rm\\s+-rf|git\\s+reset\\s+--hard|git\\s+checkout\\s+--|git\\s+clean\\s+-f|drop\\s+table|delete\\s+from|truncate|git\\s+push\\s+--force|dd\\s+if=)\\b';
+const _EXTRA_DESTRUCTIVE = (process.env.GATEGUARD_BASH_EXTRA_DESTRUCTIVE || '').trim();
+const DESTRUCTIVE_BASH = _EXTRA_DESTRUCTIVE
+  ? new RegExp(`(${_BASE_DESTRUCTIVE})|(${_EXTRA_DESTRUCTIVE})`, 'i')
+  : new RegExp(_BASE_DESTRUCTIVE, 'i');

 // ... (in run(), inside `if (toolName === 'Bash')`, after the destructive branch)

     if (DESTRUCTIVE_BASH.test(command)) {
       // ... existing destructive gate ...
     }

+    if (process.env.GATEGUARD_BASH_ROUTINE_DISABLED === '1') {
+      return rawInput; // routine gate opted out
+    }
+
     if (!isChecked(ROUTINE_BASH_SESSION_KEY)) {
       markChecked(ROUTINE_BASH_SESSION_KEY);
       return denyResult(routineBashMsg());
     }
```

Defaults preserve current behavior (no env var set → no change). Happy to send this as a PR if option 2 looks acceptable.

### Reference: current local patch

We wrap `runGateGuard` to short-circuit on non-destructive commands. Patch is regenerated against `~/.claude/plugins/cache/everything-claude-code/everything-claude-code/<ver>/scripts/hooks/bash-hook-dispatcher.js` after every ECC update:

```js
{
  id: 'pre:bash:gateguard-fact-force',
  profiles: 'standard,strict',
  // local patch 2026-05-27: only fire on destructive Bash patterns.
  run: rawInput => {
    try {
      const data = JSON.parse(rawInput);
      const cmd = (data && data.tool_input && data.tool_input.command) || '';
      const DESTRUCTIVE = /(\brm\s+-r[fr]*\b|\bgit\s+reset\s+--hard\b|\bgit\s+(push|reset|checkout)\s+(-f|--force)\b|\bsudo\b|\bDROP\s+(TABLE|DATABASE|SCHEMA)\b|\bTRUNCATE\b|\bkubectl\s+delete\b|\bdocker\s+(rm\s+-f|system\s+prune)\b|--no-verify\b|\bgit\s+clean\s+-f\b)/i;
      if (!DESTRUCTIVE.test(cmd)) {
        return { exitCode: 0, output: '', stderr: '' };
      }
    } catch (e) { /* fall through to gateguard on parse failure */ }
    return runGateGuard(rawInput);
  },
},
```

We rebuild this after every ECC update. Happy to send a PR against the JS port if you'd accept the env-var (option 2) or `.gateguard.yml`-read (option 1) approach — let me know which shape you'd prefer to review.

### Environment

- ECC version: `v1.10.0` (cache path: `~/.claude/plugins/cache/everything-claude-code/everything-claude-code/1.10.0/`)
- Claude Code: latest
- Platform: macOS 14
- Python `gateguard-ai`: not installed (we'd prefer not to add a Python dependency just for config)

### Cross-references

- Python upstream config docs: [zunoworks/gateguard README — Configuration section](https://github.com/zunoworks/gateguard#configuration)
- JS port docstring acknowledging the gap: `scripts/hooks/gateguard-fact-force.js:19`
- Hook hardcoded path: `scripts/hooks/gateguard-fact-force.js:404-407`

gaurav0107:
## Summary

- Adds `GATEGUARD_BASH_ROUTINE_DISABLED` env var: truthy values (`1`, `true`, `on`, `yes`, `enabled`) skip the once-per-session routine bash gate in `scripts/hooks/gateguard-fact-force.js`. Destructive gate is unaffected.
- Adds `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` env var: regex source string for operator-supplied additional destructive patterns. Matches against the same quote-stripped, subshell-flattened command that the built-in `DESTRUCTIVE_SQL_DD` regex sees, so a phrase inside `$(...)` or backticks is also caught.
- Defaults preserve current behavior (both env vars unset → identical to current code path).
- Malformed `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` regex is logged once to stderr with a `[gateguard-fact-force]` prefix and treated as not configured — hooks must never crash tool execution because of operator config errors.
- Implements Option 2 from #2078 (the env-var sketch). Option 1 (`.gateguard.yml` reading) is deliberately out of scope for this PR.

## Verification

- `node tests/run-all.js` — **2619/2619 passed, 0 failed** (12 new tests added in `tests/hooks/gateguard-fact-force.test.js`)
- `npx eslint scripts/**/*.js tests/**/*.js` — **clean**
- New test coverage:
  - `GATEGUARD_BASH_ROUTINE_DISABLED=1` skips routine bash gate
  - Truthy aliases (`true`, `on`, `yes`, `enabled`, mixed case) all enable opt-out
  - Unset preserves baseline (denies first routine bash)
  - Falsy values (`0`, `false`, `off`, empty, unknown string) keep current behavior
  - Routine opt-out does NOT disable destructive bash gate (rm -rf still gated)
  - `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` custom phrase fires destructive gate
  - Alternation second member also fires
  - Invalid regex degrades to baseline (no crash)
  - Unset extra-regex preserves baseline (no false positives)
  - Custom phrase inside `$(...)` is also caught

Fixes #2078

---
## Summary by cubic
Adds env options to the Bash gateguard to disable the once-per-session routine gate and to add custom destructive regex patterns. Defaults are unchanged; destructive protections remain on. Fixes #2078.

- **New Features**
  - `GATEGUARD_BASH_ROUTINE_DISABLED`: truthy values (`1`, `true`, `on`, `yes`, `enabled`) skip the routine Bash gate; destructive gate still runs.
  - `GATEGUARD_BASH_EXTRA_DESTRUCTIVE`: regex source for extra destructive phrases; matched against quote-stripped, subshell-flattened commands. Malformed regex is logged and ignored (no crash).

- **Bug Fixes**
  - Reset warning when `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` changes so each distinct invalid regex logs once, avoiding silent failures in long-running processes.

<sup>Written for commit 9e5a59be58edd04c000853c92f5211bd832eb08a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/affaan-m/ECC/pull/2161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Operators can define custom destructive command patterns via an environment variable to extend destructive-command detection.
  * Added a configuration toggle to disable routine Bash gating while keeping destructive-command blocking active.

* **Bug Fixes**
  * Malformed custom patterns no longer crash the gate; invalid patterns are treated as unset and an error is logged only once per distinct value.

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -76,10 +76,14 @@ function getExtraDestructiveRegex() {
   if (raw === extraDestructiveCacheKey) {
     return extraDestructiveCacheRegex;
   }
+  // The env value just changed; reset the once-per-pattern warning gate
+  // so a subsequent *different* invalid regex is also reported once. The
+  // previous shape kept the flag sticky and silently swallowed the
+  // second bad pattern in a long-running process.
   extraDestructiveCacheKey = raw;
+  extraDestructiveWarnLogged = false;
   try {
     extraDestructiveCacheRegex = new RegExp(raw, 'i');
-    extraDestructiveWarnLogged = false;
   } catch (err) {
     extraDestructiveCacheRegex = null;
     if (!extraDestructiveWarnLogged) {
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -46,12 +46,57 @@ const ROUTINE_BASH_SESSION_KEY = '__bash_session__';
 const EDIT_WRITE_HOOK_ID = 'pre:edit-write:gateguard-fact-force';
 const BASH_HOOK_ID = 'pre:bash:gateguard-fact-force';
 const ECC_DISABLE_VALUES = new Set(['0', 'false', 'off', 'disabled', 'disable']);
+const ECC_ENABLE_VALUES = new Set(['1', 'true', 'on', 'enabled', 'enable', 'yes']);
 
 // SQL-keyword + dd patterns stay as a single regex — they are stable
 // phrases without shell-flag ordering concerns. Quoted strings are
 // stripped before this regex runs so a commit message mentioning
 // "drop table" no longer triggers a false positive.
 const DESTRUCTIVE_SQL_DD = /\b(drop\s+table|delete\s+from|truncate|dd\s+if=)\b/i;
+
+// Operator-supplied additional destructive patterns. Lazily compiled from
+// `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` (regex source) on first use, then
+// memoized keyed by the env-var value so a test or long-running process
+// that flips the env between calls re-reads it without paying for a
+// recompile on every invocation. A malformed regex is treated as
+// "not configured" (the gate falls back to the built-in patterns) and
+// the parse failure is logged once via `[gateguard-fact-force]` to
+// stderr — hooks must never crash tool execution because of operator
+// config errors.
+let extraDestructiveCacheKey = null;
+let extraDestructiveCacheRegex = null;
+let extraDestructiveWarnLogged = false;
+function getExtraDestructiveRegex() {
+  const raw = process.env.GATEGUARD_BASH_EXTRA_DESTRUCTIVE || '';
+  if (!raw) {
+    extraDestructiveCacheKey = '';
+    extraDestructiveCacheRegex = null;
+    return null;
+  }
+  if (raw === extraDestructiveCacheKey) {
+    return extraDestructiveCacheRegex;
+  }
+  extraDestructiveCacheKey = raw;
+  try {
+    extraDestructiveCacheRegex = new RegExp(raw, 'i');
+    extraDestructiveWarnLogged = false;
+  } catch (err) {
+    extraDestructiveCacheRegex = null;
+    if (!extraDestructiveWarnLogged) {
+      try {
+        process.stderr.write(
+          `[gateguard-fact-force] ignoring invalid GATEGUARD_BASH_EXTRA_DESTRUCTIVE regex: ${err.message}\n`
+        );
+      } catch (_) { /* stderr write failure is non-fatal */ }
+      extraDestructiveWarnLogged = true;
+    }
+  }
+  return extraDestructiveCacheRegex;
+}
+
+function isRoutineBashGateDisabled() {
+  return ECC_ENABLE_VALUES.has(normalizeEnvValue(process.env.GATEGUARD_BASH_ROUTINE_DISABLED));
+}
 
 /**
  * Strip the contents of single- and double-quoted strings so phrases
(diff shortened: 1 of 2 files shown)


gaurav0107:
Thanks @coderabbitai and @cubic-dev-ai — both findings were the same valid bug, fixed in 9e5a59be.

`extraDestructiveWarnLogged` is now reset whenever the cache key changes (right after the new `raw` is assigned, before the `try/RegExp` block), so a process that flips through `bad-pattern-a` → `bad-pattern-b` reports both. A same-process regression test was added that spies on `process.stderr.write` via `loadDirectHook()` and asserts:

- same invalid pattern warns once (sanity baseline)
- switching to a different invalid pattern emits a second warning (the bug being fixed)
- switching to a valid regex emits zero warnings (no regression on the happy path)

`node tests/hooks/gateguard-fact-force.test.js` → 102/102 passing.

PR #2238, opened 2026-06-11
feat: add Kiro as a first-class install target

robery567:
## What Changed

This makes Kiro (https://kiro.dev) a proper install target instead of the
half-wired `.kiro/` folder that was sitting in the repo. It now works the same
way the Codex, Gemini, and Zed adapters do, and it goes through the same
manifest-driven installer.

Target registration:

- Added `scripts/lib/install-targets/kiro-project.js` and registered it in `registry.js`.
- Added `kiro` to `SUPPORTED_INSTALL_TARGETS`, to both schema enums
  (`schemas/ecc-install-config.schema.json` and `schemas/install-modules.schema.json`),
  to the `platform-configs` module in the manifest, to the `.kiro` owner map in
  `helpers.js`, and to the installer help text.
- `./install.sh --profile minimal --target kiro` now plans and installs `.kiro/`.

Generator so it stays in sync:

- Added `scripts/generate-kiro-adapter.js` and `scripts/lib/kiro/generate.js`.
  They rebuild the committed `.kiro/` tree from the canonical sources so it
  doesn't drift:
  - `agents/*.md` becomes `.kiro/agents/<name>.md` for the IDE plus a
    `<name>.json` for the CLI. Tool hints are mapped to Kiro's allowedTools
    (read-only tools to `fs_read`, write tools to `fs_write`, `Bash` to `execute_bash`).
  - `skills/<id>/` is copied straight over to `.kiro/skills/<id>/` since the
    SKILL.md format is the same.
  - `rules/<ns>/*.md` becomes `.kiro/steering/*.md` with inclusion frontmatter
    added (common rules are always-on, language rules load on matching files).
  - `mcp-configs/mcp-servers.json` becomes `.kiro/settings/mcp.json.example`.
- Hooks stay as the curated `.kiro.hook` files since the canonical hooks.json
  uses Claude plugin-bootstrap commands that don't map onto Kiro's
  askAgent/runCommand model. The generator validates them. Running the generator
  with `--check` fails CI if `.kiro/` is out of date.

Coverage went from 33 to 64 agents, 43 to 262 skills, 22 to 103 steering files,
and the MCP example now lists 32 servers.

Docs, compliance, and tests:

- Added a compliance scorecard entry and regenerated the matrix.
- Added `docs/KIRO-GUIDE.md` and fixed the inventory counts in `.kiro/README.md`.
- Added Kiro to the harness list and the cross-tool FAQ in the README.
- Added `tests/lib/kiro-generate.test.js`.
- Added `.kiro/` to the `files` list in `package.json` so it ships on npm.
- Moved the search-first approval guardrail into the canonical skill so it
  applies to every target and survives regeneration.

## Why This Change

Kiro was already half-supported, but it was a second-class citizen. The `.kiro/`
folder was a static copy with its own standalone `install.sh` and had no
connection to the real installer, the target registry, or the compliance
scorecard. It only shipped part of the catalog (33 of 64 agents, 43 of 262
skills) and its README pointed at docs that were never written.

Because it was a manual copy, it would fall behind every time the canonical
agents, skills, or rules changed. Wiring it into the installer and generating it
from the same sources as everything else means it picks up the full catalog
automatically and stays current, the same as the other adapters.

## Testing Done

- [x] Manual testing completed (planned and dry-ran `--target kiro`, ran the generator, ran the `--check` guard)
- [x] Automated tests pass locally (`node tests/run-all.js`) - 2739 passed, 0 failed
- [x] Edge cases considered and tested (generator runs are idempotent, glob patterns like `*.py` are quoted so YAML doesn't read them as aliases, other platforms' folders are filtered out, skill folders without a SKILL.md are skipped)

I also ran `install-targets.test.js`, `kiro-generate.test.js`,
`harness-adapter-compliance.js --check`, the manifest/skills/personal-path
validators, the catalog check, and eslint plus markdownlint on the changed files.

## Type of Change

- [ ] `fix:` Bug fix
- [x] `feat:` New feature
- [ ] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `ci:` CI/CD changes

- [x] No secrets or API keys committed (checked ghp_, sk-, AKIA, xoxb, xoxp; the only matches are placeholder examples inside the existing security skills)
- [x] JSON files validate cleanly (agent JSON, steering frontmatter, manifests, schemas)
- [ ] Shell scripts pass shellcheck (if applicable) - no shell scripts added, the new code is Node.js
- [ ] Pre-commit hooks pass locally (if configured) - the repo has no git hooks configured

## Documentation

- [x] Updated relevant documentation (`docs/KIRO-GUIDE.md` and the `.kiro/README.md` inventory)
- [x] Added comments for complex logic (the generator module and the adapter)
- [x] README updated (harness list and cross-tool FAQ)


affaan-m:
Appreciate the effort to make Kiro a first-class target — that's a direction we're interested in. But I can't land this as-is: it commits ~91k lines / 660 files, duplicating every agent and skill into `.kiro/` as both `.json` and `.md`. That vendored copy would immediately drift from the canonical `agents/` and `skills/` sources, and a diff that size can't be meaningfully reviewed. The way ECC adds harness targets is via the install/generator manifests that render each adapter's layout at install time from the single source of truth — not by committing the generated output. If you'd like to add Kiro that way (a generator + manifest entry, no vendored tree), I'd be glad to review it. Closing this version for now.

PR #2292, opened 2026-06-18
fix(gateguard): check isDestructiveFindExec on each command segment to close compound-command bypass

### LINKED ISSUE #2291: Security: destructive find -exec bypasses gateguard via compound commands (&&, ;, |)
## Summary

`isDestructiveBash` in `scripts/hooks/gateguard-fact-force.js` calls `isDestructiveFindExec` **only on the full raw command string**. Because `isDestructiveFindExec` inspects the first token of the command, any compound command where `find` is not the first word defeats the check entirely.

## Bypass

```bash
echo x && find . -exec rm {} \;
true; find . -name "*.log" -exec rm -rf {} \;
echo start | find . -exec rm {} \;
```

**Trace through `isDestructiveBash`:**
1. `isDestructiveFindExec(raw)` → first token is `echo` / `true` → returns `false`
2. `collectExecutableBodies(raw).flatMap(splitCommandSegments)` correctly splits into `["echo x", "find . -exec rm {} \;"]`
3. For the `find . -exec rm {} \;` segment: `isDestructiveRm` → false (first token is `find`), `isDestructiveGit` → false, **`isDestructiveFindExec` is never called** → segment passes through

Result: the gateguard **does not block** the destructive command.

## Root Cause (code location)

`scripts/hooks/gateguard-fact-force.js`, inside `isDestructiveBash`, the per-segment loop (lines ~552–559):

```js
for (const segment of segments) {
  const stripped = stripQuotedStrings(segment);
  if (DESTRUCTIVE_SQL_DD.test(stripped)) return true;
  if (extra && extra.test(stripped)) return true;
  const tokens = tokenize(segment);
  if (isDestructiveRm(tokens)) return true;
  if (isDestructiveGit(tokens)) return true;
  // ❌ isDestructiveFindExec(segment) never called here!
}
```

## Fix

Add one line inside the segment loop:

```js
  if (isDestructiveFindExec(segment)) return true;   // ← add this
  const tokens = tokenize(segment);
```

## Impact

Any agent or user able to supply a Bash command can prefix a destructive `find -exec rm` with an innocuous command to bypass the gateguard safety hook entirely, achieving arbitrary file deletion on the host.

## PR

A fix with three regression tests (`&&`, `;`, `|` bypass patterns) is attached.

kapil971390:
## Summary

- `isDestructiveBash` called `isDestructiveFindExec` only on the raw full command string, not on individual segments after splitting on `&&`, `;`, `|`
- Commands like `echo x && find . -exec rm {} \;` bypassed the gateguard entirely because the first token was `echo`, not `find`
- Fix: add `if (isDestructiveFindExec(segment)) return true;` inside the per-segment loop
- Adds 3 regression tests covering `&&`, `;`, and `|` compound-command bypass patterns

Closes #2291

## Test plan

- [ ] `node tests/hooks/gateguard-fact-force.test.js` → 132 passed, 0 failed
- [ ] `echo x && find . -exec rm {} \;` is now blocked
- [ ] `true; find . -name "*.log" -exec rm -rf {} \;` is now blocked
- [ ] `echo start | find . -exec rm {} \;` is now blocked
- [ ] `find . -exec rm {} \;` (direct, no prefix) still blocked as before

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -553,6 +553,7 @@ function isDestructiveBash(command) {
     const stripped = stripQuotedStrings(segment);
     if (DESTRUCTIVE_SQL_DD.test(stripped)) return true;
     if (extra && extra.test(stripped)) return true;
+    if (isDestructiveFindExec(segment)) return true;
     const tokens = tokenize(segment);
     if (isDestructiveRm(tokens)) return true;
     if (isDestructiveGit(tokens)) return true;
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -545,15 +545,24 @@ function isDestructiveBash(command) {
   const extra = getExtraDestructiveRegex();
   if (extra && extra.test(flattened)) return true;
 
-  // Check for destructive find -exec patterns
-  if (isDestructiveFindExec(raw)) return true;
+  // Check for destructive find -exec patterns on raw body segments (before quote-stripping)
+  // so that quoted exec binaries and compound-command prefixes are both handled correctly.
+  // splitCommandSegments strips quotes before splitting, so passing its output to
+  // isDestructiveFindExec would turn `find . -exec 'rm' {} \;` into `find . -exec  {} \;`
+  // — the binary name disappears and the check returns false.  Using raw body text avoids
+  // that false-negative while also catching `&&`, `;`, `|`, and `||` compound forms.
+  const bodies = collectExecutableBodies(raw);
+  for (const body of bodies) {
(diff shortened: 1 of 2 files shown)


kapil971390:
Thanks for the review @greptile-apps! Both P2 concerns addressed in the latest commit (a9a45c78):

**P1 — quoted exec binary false-negative**: Confirmed valid. \`splitCommandSegments\` calls \`stripQuotedStrings\` internally before splitting, so passing its output to \`isDestructiveFindExec\` would erase quoted binaries like \`'rm'\`. Fixed by switching to raw body segments — \`collectExecutableBodies\` output split on \`[;|&]+\` without quote-stripping, so the exec binary name is always intact when checked.

**P2 — \`||\` separator not tested**: Also fixed. The \`[;|&]+\` split pattern already handled it in the implementation; added a regression test for \`false || find . -exec rm {} \;\`.

All 133 tests pass.

PR #2339, opened 2026-06-22
fix(clv2): escape $HOME before pgrep -f in migrate-homunculus.sh

### LINKED ISSUE #2301: bug: migrate-homunculus.sh pgrep pattern treats $HOME as regex
## Summary

In `skills/continuous-learning-v2/scripts/migrate-homunculus.sh` line 23:

```bash
if pgrep -f "${HOME}.*observer-loop\\.sh" >/dev/null 2>&1; then
```

`pgrep -f` treats its argument as an extended regular expression. `$HOME` is interpolated directly into the pattern without escaping. If `$HOME` contains regex metacharacters (e.g., `+`, `.`, `(`, `[`), the pattern will either match unintended processes or fail with a regex error.

## Examples of problematic HOME values

- `/home/user.name` — the `.` matches any character
- `/home/c++dev` — the `+` is a regex quantifier
- `/home/user (work)` — parentheses are regex grouping

## Impact

- On systems where `$HOME` contains regex metacharacters, the safety check may fail to detect a running observer (false negative → migration proceeds while observer is running, risking data corruption)
- Or it may match unrelated processes (false positive → migration is blocked unnecessarily)

## Suggested fix

Escape regex metacharacters in `$HOME` before passing to `pgrep -f`, or use a different approach:

```bash
# Option 1: escape metacharacters
escaped_home=$(printf '%s' "$HOME" | sed 's/[.+*?^${}()|[\]\\]/\\&/g')
if pgrep -f "${escaped_home}.*observer-loop\\.sh" >/dev/null 2>&1; then

# Option 2: use pgrep with a simpler pattern
if pgrep -f 'observer-loop\.sh' >/dev/null 2>&1; then
```

gaurav0107:
## Summary

- `skills/continuous-learning-v2/scripts/migrate-homunculus.sh` interpolated `$HOME` unescaped into a `pgrep -f` argument. `pgrep -f` treats its argument as an extended regular expression, so regex metacharacters in the home path changed the match semantics.
- Affected paths include `/home/user.name` (the `.` matches any character), `/home/c++dev` (`+` is a quantifier), and `/home/user (work)` (parentheses are grouping).
- Consequences: a **false negative** (a live `observer-loop.sh` is missed, the migration proceeds while the observer is running and risks registry corruption) or a **false positive** (an unrelated process matches and the migration is blocked unnecessarily).
- Fix: escape the ERE metacharacters in `$HOME` with `sed` before building the pattern, so the home prefix is matched literally while the intentional trailing `.*observer-loop\.sh` regex is preserved. The escape character class is ordered to be portable across both BSD (macOS) and GNU `sed`.

This implements Option 1 from the issue (escape metacharacters), which preserves the original intent of anchoring the check to the current user's home rather than broadening it to any user's `observer-loop.sh`.

## Verification

- `bash -n skills/continuous-learning-v2/scripts/migrate-homunculus.sh` — syntax OK.
- Escaping verified against the issue's example values on BSD `sed`:
  - `/home/user.name` -> `/home/user\.name`
  - `/home/c++dev` -> `/home/c\+\+dev`
  - `/home/user (work)` -> `/home/user \(work\)`
  - `/tmp/h[x]` -> `/tmp/h\[x\]`
- Functional check: the escaped pattern matches the literal home path, and a negative test confirms `/home/user.name` no longer over-matches `/home/userXname` (the exact false-positive class from the report).
- `node tests/run-all.js` — 2891 passed, 0 failed.
- `node scripts/ci/validate-no-personal-paths.js` — pass.
- `node scripts/ci/check-unicode-safety.js` — pass.
- `npm run security:ioc-scan` — pass (199 files inspected).

Fixes #2301

### DIFF
--- a/skills/continuous-learning-v2/scripts/migrate-homunculus.sh
+++ b/skills/continuous-learning-v2/scripts/migrate-homunculus.sh
@@ -20,7 +20,13 @@ if [ ! -d "$OLD" ]; then
 fi
 
 if command -v pgrep >/dev/null 2>&1; then
-  if pgrep -f "${HOME}.*observer-loop\\.sh" >/dev/null 2>&1; then
+  # pgrep -f treats its argument as an extended regular expression, so $HOME
+  # must be escaped before interpolation. Without this, regex metacharacters in
+  # the path (e.g. /home/user.name, /home/c++dev, /home/user (work)) would make
+  # the match over-broad or invalid, causing false negatives (observer missed,
+  # migration proceeds unsafely) or false positives (migration blocked).
+  escaped_home="$(printf '%s' "$HOME" | sed 's/[]\.[(){}+*?|^$]/\\&/g')"
+  if pgrep -f "${escaped_home}.*observer-loop\\.sh" >/dev/null 2>&1; then
     echo "Refusing to migrate: observer-loop.sh is running." >&2
     echo "Exit all Claude Code sessions, then re-run." >&2
     exit 1
--- a/tests/hooks/migrate-homunculus-home-escape.test.js
+++ b/tests/hooks/migrate-homunculus-home-escape.test.js
@@ -22,6 +22,14 @@
 
 'use strict';
 
+// migrate-homunculus.sh and this test's assertions rely on POSIX bash, sed, and
+// grep -E semantics. Skip on Windows, matching the repo convention for
+// bash-dependent clv2 tests (see tests/hooks/observe-subdirectory-detection.test.js).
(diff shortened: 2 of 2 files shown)


daltino (approved):
This PR improves safety by escaping special characters in the `$HOME` path before using it with `pgrep -f` to avoid regex misbehavior. The explanation in the code comments is clear, and the new regression test is thorough. This fix aligns well with the repo's focus on robust workflows and testing.

PR #2343, opened 2026-06-23
fix(hooks): quote args when probing Windows .cmd MCP servers via shell

phobicdotno:
## Summary

- **Symptom**: On Windows, any stdio MCP server whose `args` contain a space
  (e.g. `--codesys-path "C:\Program Files\..."`) is falsely reported as
  unhealthy by the `mcp-health-check` PreToolUse hook, causing every MCP tool
  call to that server to be blocked — even though the server itself is
  perfectly healthy.

- **Root cause**: In `probeCommandServer`, when the command falls back to a
  `.cmd` candidate the probe sets `useShell: true`. It then calls
  `spawn(tryCommand, args, { shell: true })`. With `shell: true` **and** a
  separate `args` array, Node.js concatenates all tokens without quoting
  (documented as `DEP0190`). cmd.exe then re-splits the concatenated line at
  every space, so the path `C:\Program Files\Some Server\server.exe` arrives
  at the child as `C:\Program` — the executable is not found, the process
  exits immediately, the probe times out with exit code rather than the
  expected "still running after timeout" signal, and the server is marked
  unhealthy.

- **Fix**: Added a `quoteWin(token)` helper that double-quotes any token
  containing whitespace or cmd metacharacters. In the `useShell` branch, the
  command and all args are now joined into a single pre-quoted command-line
  string and passed as the first (and only) argument to `spawn()` with no
  separate args array. The `else` branch (`shell: false`, all non-`.cmd`
  commands) is unchanged. All existing timeout/kill logic (taskkill tree-kill),
  ENOENT/EINVAL retry, stderr capture, and candidate-fallback behaviour are
  preserved.

- **Regression test added**: On Windows, the new test creates a `.cmd` shim
  that echoes its first positional argument (`%1`) to stderr and then stays
  alive, invokes the hook with `--codesys-path "C:\Program Files\..."` as an
  arg, and asserts (a) the probe marks the server healthy and (b) stderr does
  not contain the split fragment `C:\Program` alone — confirming the path
  reached cmd.exe as a single quoted token.

## Test plan

- [x] `node tests/hooks/mcp-health-check.test.js` — 24/24 pass (including
  the new Windows test)
- [x] `node tests/run-all.js` — all failures are pre-existing (missing `ajv`
  dep in test environment), none related to the changed files
- [x] Manually verified with a throwaway script that the old (unquoted)
  cmdline splits at the space and the new (quoted) cmdline does not

### DIFF
--- a/scripts/hooks/mcp-health-check.js
+++ b/scripts/hooks/mcp-health-check.js
@@ -338,6 +338,21 @@ function probeCommandServer(serverName, config) {
     // through shell mode.
     const UNSAFE_SHELL_CHARS = /[&|<>^%!()\s;]/;
 
+    // When spawning via cmd.exe (shell:true) on Windows, Node concatenates
+    // command + args WITHOUT quoting (DEP0190). An arg containing a space —
+    // such as a path under "C:\Program Files" — gets re-split by cmd.exe.
+    // Build a properly-quoted command line instead and pass it as a single
+    // string with no args array, so cmd.exe sees each token as one unit.
+    function quoteWin(token) {
+      // If the token has no characters that need quoting, return it as-is.
+      if (!/[\s"&|<>^%!();]/.test(token)) {
+        return token;
+      }
+      // Escape embedded double quotes by doubling them, then wrap in double
+      // quotes. cmd.exe uses "" as an escaped quote inside a quoted string.
+      return '"' + token.replace(/"/g, '""') + '"';
+    }
+
     function attempt(idx) {
       const tryCommand = candidates[idx];
       const isLast = idx + 1 >= candidates.length;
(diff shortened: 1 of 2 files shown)


mc856:
One nit while we're on cmd.exe quirks: %VAR% still expands inside double quotes, so an arg like `--foo "%USERNAME%"` would survive `quoteWin` but still get rewritten by the shell before the child sees it. Probably out of scope for the health probe (args are usually flags/paths), just flagging in case it bites later.

daltino (approved):
This PR addresses a critical issue with handling command arguments containing spaces in Windows `.cmd` scripts when using `shell:true`, which aligns with the documented guidelines for enhancing hooks. The solution is appropriately implemented with a quoting function and thoroughly tested with Windows-specific scenarios. The additional tests are robust and ensure behavior is accurate on Windows without affecting other platforms. Looks good to merge!

PR #2392, opened 2026-06-29
fix(hooks,lib): fix hook detection and parsing edge cases

djpjronline-netizen:
## Summary

Correctness fixes for several hooks and lib parsers, each with regression tests. Found via a focused review of `scripts/hooks` and `scripts/lib`.

## Fixes

- **`auto-tmux-dev.js`** detached one-shot `dev-*` scripts. `dev\b` matches the `dev` prefix of `npm run dev-build` / `dev-docs` (hyphen is a `\b` boundary), so finite scripts were rerouted into a detached tmux session and their output disappeared. Aligned to the corrected `(?![\w-])` pattern already used by `pre-bash-dev-server-block.js`.
- **`block-no-verify.js`** false-positive on a message-option value. For `merge`/`cherry-pick`/`am`/`rebase`, the value-skip for `-m/--message/-F/--file` was gated on `command === 'commit'`, so `git merge -m --no-verify <branch>` (where `--no-verify` is the literal commit message git consumes as the `-m` value) was blocked. Added a narrow value-skip for those message options on the non-commit subcommands. A bare `--no-verify` (no preceding `-m`/`-F`) still reaches the flag check, so real bypass attempts — including `VAR=1 git commit --no-verify` and `git merge -m "msg" --no-verify` — are still blocked.
- **`pre-bash-commit-quality.js`** — two fixes:
  - The heuristic `api[_-]?key` rule emits `error` severity (blocks the commit) on benign source like `apiKey: "process.env.API_KEY"`, `"${API_KEY}"`, `"<PLACEHOLDER>"`. Now captures the quoted value and skips flagging only when it is an *obvious* non-secret (whole-value env reference / interpolation / angle-bracket token, or a short explicit placeholder whitelist). Deliberately narrow — uppercase-hex, base32, digit-only, and mixed high-entropy values are **not** suppressed, so real non-prefixed secrets still flag.
  - The `-m` message extractor truncated at the first quote char, so `-m "fix: don't crash"` validated only `fix: don`. Made it quote-aware (capture to the matching closing quote, or the full unquoted tail).
- **`pre-compact.js`** annotated the wrong project's session. It appended the compaction marker to whichever `*-session.tmp` was newest by mtime in the shared sessions dir — frequently another project/worktree. Now selects by `**Worktree:**` header against `cwd` (then legacy `**Project:**` name), mirroring `session-start.js#selectMatchingSession`, and skips rather than annotating a foreign session. The global `compaction-log.txt` still records every event.
- **`shell-substitution.js`** double-appended a trailing backslash in an unterminated backtick span (`extractCommandSubstitutions`), corrupting the body the PreToolUse hooks scan.
- **`utils.js` `readStdinJson`** silently truncated a payload over `maxSize` (1 MB) into invalid JSON parsed as `{}`. Now stops at a clean chunk boundary and surfaces the overflow on stderr instead of dropping input silently.

## Tests

Regression tests added/extended for each fix, including a new `tests/hooks/pre-compact.test.js`. The affected suites pass locally:

```
tests/hooks/pre-compact.test.js               7 passed
tests/hooks/auto-tmux-dev.test.js            12 passed
tests/hooks/block-no-verify.test.js          28 passed
tests/hooks/pre-bash-commit-quality.test.js  14 passed
```

(`auto-tmux-dev.test.js` has some pre-existing environment-dependent flakiness around real tmux session creation in back-to-back runs, unrelated to this change; it passes 5/5 in isolation.) Each fix was additionally validated with isolated behavioral assertions covering the no-false-negative guards (bypasses still blocked, real secrets still flagged).

### DIFF
--- a/scripts/hooks/auto-tmux-dev.js
+++ b/scripts/hooks/auto-tmux-dev.js
@@ -37,8 +37,12 @@ function run(rawInput) {
     const cmd = input.tool_input?.command || '';
 
     // Detect dev server commands: npm run dev, pnpm dev, yarn dev, bun run dev
-    // Use word boundary (\b) to avoid matching partial commands
-    const devServerRegex = /(npm run dev\b|pnpm( run)? dev\b|yarn dev\b|bun run dev\b)/;
+    // Trailing (?![\w-]) rather than \b: \b treats a hyphen as a word boundary,
+    // so `dev\b` matches the `dev` prefix of distinct scripts like `dev-build` /
+    // `dev-docs` and would wrongly detach those one-shot scripts into tmux. The
+    // lookahead still matches the dev server (`dev`, `dev:ssr`, ...) but not a
+    // `dev-<suffix>` script. Mirrors pre-bash-dev-server-block.js DEV_PATTERN.
+    const devServerRegex = /(npm run dev|pnpm( run)? dev|yarn dev|bun run dev)(?![\w-])/;
 
     if (devServerRegex.test(cmd)) {
       // Get session name from current directory basename, sanitize for shell safety
--- a/scripts/hooks/block-no-verify.js
+++ b/scripts/hooks/block-no-verify.js
@@ -251,6 +251,24 @@ function isCommitNoVerifyShortFlag(value) {
   return value === '-n' || /^-n[a-zA-Z]/.test(value);
 }
 
+// Message/value options shared by the non-'commit' subcommands that still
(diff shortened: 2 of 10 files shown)


djpjronline-netizen:
Superseded by #2405 — this PR auto-closed when I renamed the head branch on my fork (cross-fork PRs don't follow GitHub branch renames the way same-repo PRs do). #2405 has the identical correctness-only changes under a clearer branch name. Apologies for the noise.

PR #2406, opened 2026-06-30
fix: canonicalize codex and opencode install payloads

austinmao:
## Summary
- Keep Codex install planning on canonical `.codex` payloads without pulling in foreign platform files
- Keep OpenCode install planning on canonical `.opencode` payloads while avoiding `.opencode/commands` duplication
- Add regression coverage for both payload planners
- Add changelog entry under Unreleased

## Verification
- `node tests/lib/install-targets.test.js`
- `node tests/scripts/doctor.test.js`
- `npm test` → 2939 passed, 0 failed

## Notes
- The repo still reports `2.0.0` in `package.json` and `VERSION`, so the gstack version-bump helper could not parse the base version as a 4-part release. I left the version unchanged and documented the change in `CHANGELOG.md` instead.


austinmao:
Closing because this change was shipped against the wrong repository.

Issue #2142, opened 2026-06-04
GateGuard fact-force restate-retry loop amplifies model repetition traps in long sessions


Furinelle:
## Context

`gateguard-fact-force` requires restating 4 facts before each Bash/Edit/Write, then blocks the operation once and forces a retry. In a long session this produced **10+ near-identical "state facts → blocked → restate → retry" blocks** in the same context window.

## Problem

This highly repetitive, near-duplicate structure measurably raises the odds of the underlying model dropping into a **degenerate repetition loop**. In one long session I observed **3 separate collapses** where the model emitted a single filler token (`court`) thousands of times until manually interrupted. The collapse consistently started at the transition right before a guarded tool call.

The guard itself is well-intentioned, but its forced repetition is a contributing factor to model instability in long sessions.

## Reproduction

1. Run a long autonomous session (50+ tool calls) with GateGuard enabled
2. Each guarded Bash/Edit/Write triggers: state facts → blocked → restate same facts → retry
3. After 10+ such cycles, the model enters a degenerate single-token repetition loop that cannot self-recover

## Suggestions

- **Don't re-trigger the full fact-force on a *retry* of the same operation** — the facts were stated one step prior
- Add a per-session cache: "facts already established for file X" to avoid restating identical facts
- Document `ECC_GATEGUARD=off` / `ECC_DISABLED_HOOKS` more prominently for long autonomous sessions

## Environment

ECC `2.0.0-rc.1`, Claude Code, macOS, claude-opus-4-8 model

affaan-m:
#2161 (merged) added env knobs to tune/disable the routine-bash gate, which mitigates the restate-retry amplification. The deeper fix for the loop dynamic is still open — keeping this issue for that.

mvanhorn:
Implemented in https://github.com/affaan-m/ECC/pull/2192.

PR #2432, opened 2026-07-03
feat(gateguard): add GATEGUARD_EXEMPT_GLOBS path exemptions

Zalasyu:
## Motivation

The fact-forcing gate fires once per **first-touched file** per session. In a build-heavy session spanning many files, every new file costs a deny→retry round-trip — including trees where the gate's questions ("who imports this?", "what data schema?") carry no signal: test files (collected, never imported), generated artifacts, and scratch dirs. In one real feature session this was ~15 first-touch denials, ~11 of them on `tests/**` / output-artifact / scratch paths.

The existing knobs don't cover this: `GATEGUARD_FACT_FORCE_FULL_DENIALS` (#2142) only shortens the *message* after N denials; there's no way to say "this *class of path* isn't worth gating."

## Change

Add an opt-in, comma-separated glob allowlist read from `GATEGUARD_EXEMPT_GLOBS`. A matching `Edit`/`Write`/`MultiEdit` target skips the first-touch gate. **Destructive-Bash and routine-Bash gates are untouched.**

- **Default-off:** unset env → byte-for-byte prior behavior.
- **Fail-open:** a malformed glob is dropped, never throws (mirrors `getExtraDestructiveRegex`).
- **Memoized** on the env value (same idiom as the extra-destructive cache).
- Globs: `*` within a segment, `**` across segments, `?` one char; matched against the normalized (forward-slash, lowercased) path via the existing `normalizeForMatch`.

```json
"env": { "GATEGUARD_EXEMPT_GLOBS": "**/tests/**,**/_bmad-output/**,**/scratchpad/**" }
```

## Implementation

- New `getExemptMatchers()` (memoized glob→RegExp compile) + `isExemptPath()`, placed next to `getExtraDestructiveRegex`.
- Wired into both the `Edit`/`Write` branch and the `MultiEdit` loop, alongside the existing `isClaudeSettingsPath` allow.

## Tests

Adds 4 cases to `tests/hooks/gateguard-fact-force.test.js` (exempt match, non-match still gated, multiple comma-separated globs incl. `Write`, default-off when unset). **All 144 gateguard tests pass** (140 prior + 4 new). `check-unicode-safety` and `validate-no-personal-paths` pass.

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable path exemptions, allowing certain file paths to skip fact-forcing checks.
  * Introduced glob-style matching for exempt paths, including multiple patterns.

* **Bug Fixes**
  * Improved handling of first-touch file operations on exempt paths so they are allowed instead of blocked.
  * Invalid exemption patterns are now ignored safely without interrupting the workflow.

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -94,6 +94,47 @@ function getExtraDestructiveRegex() {
   return extraDestructiveCacheRegex;
 }
 
+// Operator-supplied path exemptions. Comma-separated globs (`GATEGUARD_EXEMPT_GLOBS`)
+// matched against the normalized (forward-slash, lowercased) file path. First-touch
+// fact-forcing is skipped for a matching Edit/Write/MultiEdit target — intended for
+// low-import-value trees (tests, generated artifacts, scratch dirs) where "who imports
+// this / what schema" carries no signal. Memoized on the env value; fail-open (a
+// malformed pattern is dropped, never throws). `*` matches within a path segment,
+// `**` across segments, `?` a single char.
+let exemptCacheKey = null;
+let exemptCacheRegexes = null;
+function getExemptMatchers() {
+  const raw = process.env.GATEGUARD_EXEMPT_GLOBS || '';
+  if (raw === exemptCacheKey) {
+    return exemptCacheRegexes;
+  }
+  exemptCacheKey = raw;
+  exemptCacheRegexes = raw
+    .split(',')
+    .map(s => s.trim())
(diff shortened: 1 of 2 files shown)


daltino (approved):
This PR introduces a new feature allowing path-based exemptions (`GATEGUARD_EXEMPT_GLOBS`) to skip first-touch fact-forcing for specified file paths, which is useful for noise reduction in low-value areas like test files. The implementation looks clean, and the functionality is well-tested with clear test cases. Overall, this aligns with the contribution guidelines for useful automation hooks.

PR #2444, opened 2026-07-04
feat(skills): add i18n-sync — agent-driven locale translation with deterministic validation

berkayyalcin7:
## What

Adds `skills/i18n-sync` (origin: community): a workflow that makes the agent itself the translation layer for JSON locale files, while delegating everything deterministic to [locakit](https://www.npmjs.com/package/locakit) (MIT, zero-dependency CLI): key diffing, lockfile-based staleness tracking, placeholder/glossary validation and safe JSON writes.

## Why this belongs in ECC

Existing AI localization tools (lingo.dev, Languine, i18n-ai-translate…) all ship their own LLM pipeline — API keys, SaaS accounts, per-token billing. An ECC user already has a capable agent attached to their codebase. This skill uses that:

- **Real usage context**: the skill instructs the agent to grep each pending key and read the surrounding component before translating — a nav label, a page title and an error toast get different translations even for the same source string ("Home" → nav label, not the building).
- **Hallucination containment**: `locakit apply` rejects any key that doesn't exist in the source locale.
- **CI-enforceable quality**: `locakit check` validates placeholder parity (`{name}`, `{{count}}`, `%s`, `$t()`), glossary retention, and language-pack rules (Turkish pack ships first: suffix-after-placeholder vowel-harmony risk, dotless-uppercase-i).
- **Staleness tracking**: source copy changes mark affected keys stale per target via a lockfile — the "English says passkey, German still says password" class of bug.

## Validation

Dogfooded end-to-end on a production Next.js 16 site ([berkayyalcin.dev](https://berkayyalcin.dev)): 175 keys translated tr→en in a single diff→translate→apply→check pass; `locakit check` reported 0 errors, 0 warnings. PR with the full flow: berkayyalcin7/berkayyalcin-dev#22

## Format compliance

- Frontmatter: `name`, `description`, `metadata.origin: community` ✓
- "When to Activate", practical workflow, out-of-scope and Related sections ✓
- No hooks/scripts added; the skill is a single SKILL.md ✓

### DIFF
--- a/skills/i18n-sync/SKILL.md
+++ b/skills/i18n-sync/SKILL.md
@@ -1,0 +1,131 @@
+---
+name: i18n-sync
+description: >
+  Context-aware localization workflow for JSON locale files. Detects missing or
+  stale translation keys with the locakit CLI, reads how each key is used in the
+  codebase to infer real UI context, translates in the project's tone, and writes
+  results back with placeholder/glossary validation. No translation API keys —
+  the agent itself is the translator. Use when adding locale keys, adding a new
+  language, or when the user asks to sync/translate i18n files.
+metadata:
+  origin: community
+---
+
+# i18n Sync
+
+Translate application locale files the way a human localizer would: by looking
+at where each string appears in the product, not just the string itself. The
+deterministic parts (diffing, lockfile tracking, validation, safe JSON writes)
+are delegated to [locakit](https://www.npmjs.com/package/locakit); this skill
+supplies the judgment.
+
+## When to Activate
+
+- New keys were added to the source locale and target languages need catching up
+- A new target language is being introduced
+- Source copy changed and existing translations may be stale
+- The user asks to "translate", "localize" or "sync" locale/i18n JSON files
+- CI failed on `locakit check`
+
+## Requirements
+
+- `locakit.config.json` in the project root (run `npx locakit init` if absent)
+- Locale files matching the config's `files` template, e.g. `locales/{locale}.json`
+
+## Workflow
+
+### 1. Discover pending work
+
+```bash
+npx locakit diff --json
+```
+
+Returns, per target language, every `missing` and `stale` key with its source
+text, plus the project `context` and `glossary` from the config. If empty,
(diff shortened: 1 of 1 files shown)


haelyra:
Thank you for the thoughtful i18n workflow, and apologies for the review delay. The combination of agent judgment with deterministic diffs, placeholder checks, glossary enforcement, and stable key ordering is a strong direction. One correction to earlier automation feedback: `locakit` does exist on npm.

The external package still needs hardening before ECC can recommend running it against untrusted repositories:

- **Contain every path to the project root.** Repository-controlled path templates and locale names can currently escape the project and read or write arbitrary JSON files.
- **Validate config, locale, and patch data.** Reject `__proto__` and other prototype-pollution keys, malformed locale identifiers, and unexpected patch shapes before rebuilding objects.
- **Use atomic writes.** Locale and lock files are written directly despite the skill's safe-write claim, so interruption can corrupt project data.
- **Pin a hardened release.** Replace bare `npx locakit` with an exact reviewed version and remove shell-fragile single-quoted JSON piping.
- **Add the repository trust boundary and real hook contract.** Treat translation content/config as untrusted, and either implement the optional hook through ECC's actual hook-input contract or remove it.

Please cover traversal, dangerous keys, interruption recovery, malicious config, and cross-platform behavior. If you do not have bandwidth, we will add the i18n-sync idea to ECC's task queue while preserving your contribution as the source.

chenhz01:
Following up on the locakit correction in the maintainer review: one determinism gap remains in the skill's contract with it. The workflow invokes `npx locakit` unpinned, but `diff`/`check` behavior — placeholder grammar, glossary rules, the Turkish pack — is versioned, so a minor release can change what CI enforces between two runs of the same skill. Pinning in the skill body keeps validation a known quantity:
```bash
npx locakit@^1 check
```
with a note to review the changelog before bumping. That preserves the zero-config onboarding (`npx locakit init` still works) while making the skill's verdicts reproducible, which matters since the pitch is deterministic validation. Happy to open a PR with working code if useful.

PR #2310, opened 2026-06-19
feat(skills): add project-context skill — shared agent briefing file

t0kubetsu:
Closes #2308 (partial — see also sibling PRs for `dev-team` and `story-lifecycle`)

## Summary

Adds `skills/project-context/SKILL.md`, a community skill inspired by the BMAD Method's `project-context.md` contract.

- Creates and maintains a `PROJECT-CONTEXT.md` file at the repo root
- Every ECC agent reads this file at session start so context does not degrade across sessions, agents, or team members
- Eliminates the need to re-explain goals, constraints, tech stack, and team context to each new agent

**When to invoke**: when starting a new project, when agents produce inconsistent outputs due to missing shared context, or when onboarding a codebase that lacks a CLAUDE.md.

## Also included

- **fix(opencode/tools)**: normalize path separators to forward slashes on Windows
- **test(lib/github-coordination)**: branch coverage tests raising coverage to 80.0 %

## Test plan

- [ ] CI passes on `ubuntu-latest` and `windows-latest`
- [ ] `node tests/run-all.js` passes locally
- [ ] Skill file follows frontmatter conventions

PR #2309, opened 2026-06-19
feat(skills): add dev-team skill — multi-persona collaborative session

t0kubetsu:
Part of #2308 (partial — see also sibling PR #2311 for `story-lifecycle`)

## Summary

Adds `skills/dev-team/SKILL.md`, a community skill inspired by the BMAD Method's "party mode".

- Launches PM, Architect, Developer, and QA as parallel subagents responding to the same topic in one turn
- Reads `PROJECT-CONTEXT.md` from the repo root when present, so all personas share the same baseline
- Produces a structured synthesis that names tensions explicitly rather than averaging them
- Pairs with `council` (adversarial decisions) and `story-lifecycle` (convert output into epics)

**When to invoke**: when the user says "what would the team think about this", "give me all perspectives", or is starting a story and wants multi-role input before writing code.

Also registers the skill in the `workflow-quality` install module (`manifests/install-modules.json`) and the npm publish allowlist (`package.json` `files`), as required by `tests/ci/validators.test.js` and `tests/scripts/npm-publish-surface.test.js`.

> Rebased onto `main` as a skill-only diff per review feedback — the Windows path-separator fix and github-coordination coverage tests previously bundled here landed separately in #2459.

## Test plan

- [ ] CI passes on `ubuntu-latest` and `windows-latest`
- [x] `node tests/run-all.js` passes locally (3352 tests; sole failure is `tests/scripts/manual-hook-install-docs.test.js`, which also fails on current `main`)
- [ ] Skill renders correctly in ECC skill picker
- [x] Skill file follows frontmatter conventions (name, description, metadata.origin)

PR #2459, opened 2026-07-06
fix(opencode/tools): normalize backslashes cross-platform + github-coordination branch coverage

t0kubetsu:
## Summary

Splits out the shared fixes that were duplicated verbatim across the `dev-team` (#2309), `story-lifecycle` (#2311), and closed `project-context` (#2310) sibling PRs, per [affaan-m's review comment on #2309](https://github.com/affaan-m/ECC/pull/2309#issuecomment-4880738219):

> it bundles the same two unrelated changes (opencode format-code Windows fix + github-coordination tests) as #2311, which makes the trio mutually conflicting. Please rebase to a skill-only diff; put the shared fixes in one separate small PR and both can land.

- **fix(opencode/tools)**: `buildFormatterCommand`'s `path.normalize(...).split(path.sep).join("/")` only splits on the platform's native separator, so backslash paths passed on non-Windows hosts (where `path.sep` is `/`) were left unconverted. Adds an explicit `.replace(/\\/g, "/")` to close the gap on Linux CI.
- **test(lib/github-coordination)**: adds branch coverage tests for `parsing.js`/`state.js`, `policy.js`, and `store.js`, raising branch coverage to 80.0%.

Once this lands, `dev-team` and `story-lifecycle` can each rebase to a skill-only diff on top of it.

## Test plan

- [x] `node tests/opencode-tools.test.js` passes
- [x] `node tests/lib/github-coordination-branches.test.js` passes
- [x] `node tests/lib/github-coordination-policy.test.js` passes
- [x] `node tests/lib/github-coordination-store.test.js` passes

### DIFF
--- a/.opencode/tools/format-code.ts
+++ b/.opencode/tools/format-code.ts
@@ -107,7 +107,7 @@ function buildFormatterCommand(formatter: Formatter, filePath: string, cwd?: str
   // Normalize to forward slashes so the emitted command is identical on every
   // platform. `path.normalize` yields backslashes on Windows, which broke the
   // command string (and Windows CI); all formatter CLIs accept `/` on Windows.
-  const normalizedPath = path.normalize(filePath).split(path.sep).join("/")
+  const normalizedPath = path.normalize(filePath).split(path.sep).join("/").replace(/\\/g, "/")
 
   // Build command based on formatter and platform
   const commands: Record<Formatter, string> = {
--- a/.opencode/tools/format-code.ts
+++ b/.opencode/tools/format-code.ts
@@ -107,7 +107,7 @@ function buildFormatterCommand(formatter: Formatter, filePath: string, cwd?: str
   // Normalize to forward slashes so the emitted command is identical on every
   // platform. `path.normalize` yields backslashes on Windows, which broke the
   // command string (and Windows CI); all formatter CLIs accept `/` on Windows.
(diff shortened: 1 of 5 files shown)


t0kubetsu:
Addressed the bot findings:

- **Greptile (P2)**: dropped the now-redundant `split(path.sep).join("/")` in `buildFormatterCommand` — `.replace(/\\/g, "/")` alone covers every case that step handled, on every platform.
- **CodeRabbit**: strengthened the "dependency issue not found" test in `github-coordination-branches.test.js` to actually assert the `stderr` warning it claims to cover (via a temporary `process.stderr.write` spy), instead of only checking the return value.

Skipped one, with reason: CodeRabbit's suggestion to extract the shared `test(name, fn)` runner into a common `tests/lib/test-helpers.js` — that inline pattern is the existing convention across 160+ test files in this suite already, so singling out just these 3 new files would be inconsistent with the rest of the codebase rather than a real fix.

All 4 affected test files still pass (93 tests, 0 failures).

affaan-m (approved):
Verified on a clean merged tree against current main: backslash normalization handles the POSIX-hosted Windows-path case, all new coordination tests are discovered by the suite, and full coverage passes (88.41% statements/lines, 80.35% branches, 93.05% functions). The remaining local test-helper nit matches the repository standalone-suite convention and is non-blocking.

PR #2462, opened 2026-07-06
fix(commands): probe for auto-update.js in auto-update ECC_ROOT resolver

karimfarahat:
## Summary

`/ecc:auto-update` fails with `MODULE_NOT_FOUND` on machines that installed ECC as a Claude Code marketplace plugin with only the hooks-runtime module applied to `~/.claude` (e.g. `install-apply --target claude --modules hooks-runtime`).

The command's inline `ECC_ROOT` resolver delegates to `resolveEccRoot()` with the default probe (`scripts/lib/utils.js`). The hooks-runtime module copies `scripts/lib/` into `~/.claude`, so that partial install satisfies the probe and the resolver short-circuits to `~/.claude` — before the `plugins/marketplaces/` candidates are ever checked. The command then tries to run `~/.claude/scripts/auto-update.js`, which doesn't exist there:

```
Error: Cannot find module '/Users/<user>/.claude/scripts/auto-update.js'
```

The fix uses the resolver's existing `probe` option so the inline resolver only accepts a root that actually contains the script the command runs:

```js
resolveEccRoot({probe:p.join('scripts','auto-update.js')})
```

With this, partial installs no longer shadow the full plugin root, and the resolver correctly falls through to `~/.claude/plugins/marketplaces/everything-claude-code`.

Applied to `commands/auto-update.md` and its ja-JP / zh-CN translations.

## Type
- [ ] Skill
- [ ] Agent
- [ ] Hook
- [x] Command

## Testing

- New regression test in `tests/lib/resolve-ecc-root.test.js`: a partial install (`scripts/lib/utils.js` only) plus a full marketplace root — asserts the default probe still prefers the standard install, and the `scripts/auto-update.js` probe skips the partial install.
- New test in `tests/lib/command-plugin-root.test.js`: asserts all three `auto-update.md` variants embed the shared inline resolver exactly once and pass the probe (fails on the pre-fix docs).
- Full suite: `node tests/run-all.js` — 2973 tests, 0 failures.
- Validators: `node scripts/ci/validate-commands.js` (93 files OK), `npm run command-registry:check` (up to date).
- Reproduced the failure and verified the fix end-to-end on macOS with a hooks-runtime-only `~/.claude` install: the patched snippet resolves to the marketplace root and `auto-update.js --dry-run` succeeds.

## Checklist
- [x] Tested with Claude Code
- [x] Clear descriptions


daltino (commented):
**DECISION: APPROVE**

This PR resolves an issue with the `auto-update` command by improving the probing logic for the `auto-update.js` file to ensure it avoids false positives from incomplete installs. The changes enhance robustness without introducing unnecessary complexity. Tests are added to verify the new behavior, and relevant changes are reflected across documentation files in multiple languages. Nice work!

karimfarahat:
Thank you @daltino!

affaan-m (changes requested):
The explicit `scripts/auto-update.js` probe is still the right fix, but this branch cannot merge cleanly into current `main`: `tests/lib/resolve-ecc-root.test.js` conflicts with the newer partial-root coverage from #2544.

Please rebase/update onto current `main`, retain the #2544 partial-root tests, and add this PR's custom `scripts/auto-update.js` probe regression after them. The current test wording should also describe the surviving failure shape accurately: a partial install has resolver evidence but lacks the script that auto-update will execute (the older hooks-runtime-only case was partially superseded by #2577).

After resolving, please rerun `node tests/lib/resolve-ecc-root.test.js`, `node tests/lib/command-plugin-root.test.js`, and the command validators. The focused PR-head tests currently pass (27/27 and 5/5), and I found no security blocker in the fixed literal probe.

karimfarahat:
Rebased onto current main (4e973d3). Changes as requested:

- Retained all four #2544 partial-root tests; the auto-update probe regression now sits after them in the Custom Probe section.
- Reworded the test (and the matching comment in command-plugin-root.test.js) to describe the surviving failure shape: a partial install can carry full resolver evidence (script tree plus sentinel ECC skill) yet still lack the script auto-update executes. The fixture now builds exactly that root via the updated setupStandardInstall helper, so the first assertion documents that the default probe rightly accepts it, and the second asserts the explicit scripts/auto-update.js probe skips it.

Reran as requested:
- node tests/lib/resolve-ecc-root.test.js: 31 passed, 0 failed
- node tests/lib/command-plugin-root.test.js: 5 passed, 0 failed
- node scripts/ci/validate-commands.js: 94 command files validated
- npm run command-registry:check: up to date

haelyra (approved):
Thank you for the focused rebase and follow-through. The exact auto-update probe is present in every maintained command variant, the #2544 partial-root coverage is retained, and the resolver, command-root, command-validation, and registry checks are all green. The earlier requested changes are satisfied, and this is ready to merge.

Issue #842, opened 2026-03-23
[Windows] Observer Haiku agent cannot read temp files and never writes instinct files


AlvaroRojas:
## Environment

- OS: Windows 11 Pro (Git Bash / MSYS2)
- ECC version: 1.9.0
- Claude Code: latest
- `ECC_OBSERVER_ALLOW_WINDOWS=true`

## Bug Description

The observer loop runs successfully on Windows (observations accumulate, Haiku agent spawns), but **zero instinct files are ever created** due to two issues in `observer-loop.sh`.

### Bug A: Temp files created in `/tmp` are inaccessible to the spawned Haiku session

`observer-loop.sh` lines 58 and 63 create temp files via:

```bash
analysis_file="$(mktemp "${TMPDIR:-/tmp}/ecc-observer-analysis.XXXXXX.jsonl")"
prompt_file="$(mktemp "${TMPDIR:-/tmp}/ecc-observer-prompt.XXXXXX")"
```

The spawned `claude --print` session cannot read these paths. On Windows/Git Bash, `/tmp` maps to a Windows temp directory that the child Claude session treats as outside its allowed scope. The Haiku agent outputs:

> "I cannot access the file at `/tmp/ecc-observer-analysis.XXXXXX.jsonl` because security restrictions limit my file operations to the project directory"

This happens ~50% of analysis cycles. The other ~50% the agent can read the file (nondeterministic).

**Suggested fix:** Use `$PROJECT_DIR` for temp files instead of `$TMPDIR`/`/tmp`:

```bash
analysis_file="${PROJECT_DIR}/ecc-observer-analysis-$$.jsonl"
prompt_file="${PROJECT_DIR}/ecc-observer-prompt-$$.txt"
```

### Bug B: Haiku agent asks for permission instead of writing instinct files

When the Haiku agent **can** read the analysis file, it identifies patterns correctly but then outputs:

> "Would you like me to save these as instinct files in your project's instinct directory? This will help future sessions automatically apply these patterns."

Instead of calling the Write tool, it asks for confirmation — which nobody can answer in `--print` mode. This happens 100% of the time when reading succeeds.

**Suggested fix:** Add explicit instructions to the prompt (around line 65):

```
IMPORTANT: You MUST write files directly using the Write tool. Do NOT ask for permission or confirmation. Do NOT output summaries or ask the user anything. Just read the observations, identify patterns, and write the instinct files.
```

## Observer Log Evidence

[...]

PR #2464, opened 2026-07-06
fix(observer): repair daemon boot (stale resolver name) and close analysis stdin (#2452)

### LINKED ISSUE #2452: continuous-learning-v2 observer broken on Windows/Git Bash: daemon boot + analysis both fail (two regressions)
## Summary

On Windows (Git Bash / MSYS2) the continuous-learning-v2 observer is broken by two independent regressions: the background daemon fails to boot, and even when booted its analysis step exits 1 before reading any observations. Observations keep being recorded, but no instincts are ever produced.

Repo path references below are relative to the plugin root (`skills/continuous-learning-v2/...`).

## Bug 1 — observer daemon fails to boot (function-name mismatch)

`agents/start-observer.sh` sources `scripts/lib/homunculus-dir.sh` and calls:

```sh
CONFIG_DIR="$(_ecc_resolve_homunculus_dir)"
```

but `scripts/lib/homunculus-dir.sh` only defines `_clv2_resolve_homunculus_dir` (the `_ecc_` name no longer exists). Every launch dies with:

```
start-observer.sh: line 40: _ecc_resolve_homunculus_dir: command not found
```

The rest of the tree (`hooks/observe.sh`, `scripts/detect-project.sh`, `scripts/migrate-homunculus.sh`) already uses `_clv2_resolve_homunculus_dir`, so `start-observer.sh` is the lone stale caller.

**Fix:** rename the call in `start-observer.sh` to `_clv2_resolve_homunculus_dir`.

## Bug 2 — analysis claude call exits 1 before reading observations (Windows)

`agents/observer-loop.sh` spawns the analysis process in the background with the prompt passed via `-p` while stdin stays open:

```sh
... claude --model "..." --max-turns "$max_turns" --print \
    --allowedTools "Read,Write" \
    -p "$prompt_content" >> "$LOG_FILE" 2>&1 &
```

On Git Bash the backgrounded child inherits an open stdin, so claude prints `Warning: no stdin data received in 3s, proceeding without it` and then exits 1 — logged as `Claude analysis failed (exit 1); retaining observations for retry` — before it ever reads the analysis file. The result: observations accumulate forever, instincts are never written.

**Fix that works for us:** write `$prompt_content` to an absolute-path temp file and feed it on stdin instead of `-p`, keeping `claude_pid` as the real PID so the watchdog still works:

```sh
observer_prompt_file="${PROJECT_DIR}/.observer-prompt.$$"
printf '%s' "$prompt_content" > "$observer_prompt_file"
... claude --model "..." --max-turns "$max_turns" --print \
    --allowedTools "Read,Write" \
    < "$observer_prompt_file" >> "$LOG_FILE" 2>&1 &
claude_pid=$!
# ... rm -f "$observer_prompt_file" after wait
```

After both fixes the full observe -> analyze -> instinct cycle completes on Windows (verified: instinct files written on trigger).

## Environment

- OS: Windows 11
- Shell: Git Bash (MSYS2)
- Plugin channel: `autoUpdatesChannel: latest`
- `ECC_OBSERVER_ALLOW_WINDOWS=true` set (so it is past the #295 Windows gate)

Happy to open a PR if the suggested fixes look right.

mc856:
## Summary
- fix the all-platform observer boot regression: `start-observer.sh:40` still called `_ecc_resolve_homunculus_dir`, but the shared lib was renamed to `_clv2_resolve_homunculus_dir` (observe.sh, detect-project.sh, and migrate-homunculus.sh were updated; start-observer.sh was missed) — under `set -e` every launch dies with exit 127, so this breaks daemon boot everywhere, not only on Windows
- close stdin (`< /dev/null`) on the backgrounded analysis `claude` call in `observer-loop.sh`: on Git Bash/MSYS2 the child inherits an open stdin, waits on it, warns "no stdin data received", and exits 1 before reading the analysis file; this is the only backgrounded claude call site in the repo
- keep the `-p` prompt flag rather than reverting to a stdin redirect, preserving the Windows-compat decision from #842 (the fix suggested in #2452 would undo it)
- add two source invariant guards to `tests/hooks/hooks.test.js`: every `*_resolve_homunculus_dir` call site must match a function the shared lib defines, and the backgrounded claude call must close stdin

## Verification
- pre-fix repro on macOS: `start-observer.sh status` dies with `line 40: _ecc_resolve_homunculus_dir: command not found`, exit 127
- post-fix: full start → status → stop cycle completes in a sandboxed `CLV2_HOMUNCULUS_DIR`
- `node tests/run-all.js` on node 22: 2973 passed, 0 failed
- shellcheck on both scripts: only two pre-existing info-level findings on untouched lines
- note: no Windows machine available — bug 1 is verified end-to-end on macOS (it is platform-independent); bug 2's fix is the standard stdin-close idiom and is behavior-neutral on macOS/Linux, but has not been exercised on Git Bash

Fixes #2452

### DIFF
--- a/skills/continuous-learning-v2/agents/observer-loop.sh
+++ b/skills/continuous-learning-v2/agents/observer-loop.sh
@@ -248,13 +248,16 @@ PROMPT
   # Pass prompt via -p flag instead of stdin redirect for Windows compatibility (#842).
   # prompt_content is already loaded in-memory so this no longer depends on the
   # mktemp absolute path continuing to resolve after cwd changes (#1296).
+  # stdin is explicitly closed with </dev/null: on Git Bash/MSYS2 the backgrounded
+  # child otherwise inherits an open stdin, and claude waits on it, warns
+  # "no stdin data received", and exits 1 before reading the analysis file (#2452).
   # Model is configurable via ECC_OBSERVER_MODEL (defaults to haiku for cost efficiency);
   # e.g. ECC_OBSERVER_MODEL=opus for higher-quality instinct extraction. Heavier models are
   # slower — consider raising ECC_OBSERVER_TIMEOUT_SECONDS (default 120s) so the watchdog
   # doesn't kill the analysis mid-run.
   ECC_SKIP_OBSERVE=1 ECC_HOOK_PROFILE=minimal claude --model "${ECC_OBSERVER_MODEL:-haiku}" --max-turns "$max_turns" --print \
     --allowedTools "Read,Write" \
-    -p "$prompt_content" >> "$LOG_FILE" 2>&1 &
+    -p "$prompt_content" < /dev/null >> "$LOG_FILE" 2>&1 &
   claude_pid=$!
 
   (
--- a/skills/continuous-learning-v2/agents/start-observer.sh
+++ b/skills/continuous-learning-v2/agents/start-observer.sh
@@ -37,7 +37,7 @@ PYTHON_CMD="${CLV2_PYTHON_CMD:-}"
 
 # shellcheck disable=SC1091
 . "${SKILL_ROOT}/scripts/lib/homunculus-dir.sh"
-CONFIG_DIR="$(_ecc_resolve_homunculus_dir)"
+CONFIG_DIR="$(_clv2_resolve_homunculus_dir)"
 if [ -n "${CLV2_CONFIG:-}" ]; then
   CONFIG_FILE="$CLV2_CONFIG"
 elif [ -f "${CONFIG_DIR}/config.json" ]; then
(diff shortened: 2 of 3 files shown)


daltino (approved):
This PR resolves critical issues in the observer scripts, including fixing a stale resolver name and properly closing stdin to prevent analysis errors. The changes look minimal yet impactful, adding robustness to daemon boot and maintaining compatibility with shared lib-defined resolvers. Additionally, the comprehensive new test ensures future reliability. Nice work!

mc856:
Dedup note: #2511 (opened Jul 13) proposes the same one-line rename for Bug 1.

Issue #2290, opened 2026-06-18
suggest-compact: context % doubled on 400k-window models (Opus 4.x)


George323ru:
## Bug

`resolveContextWindowTokens()` in `scripts/lib/transcript-context.js` only recognizes two context-window sizes: the 200k standard and 1M (detected via the `[1m]` model marker or a token count above 200k). Models with a **400k** window (e.g. Claude Opus 4.x) match neither and fall through to the 200k default.

As a result the `suggest-compact` hook reports roughly **double** the real usage — e.g. `[StrategicCompact] Context ~177k tokens (88% of 200k window)` when the real figure is ~44% of a 400k window.

## Proposed fix

Honor an explicit window override at the top of `resolveContextWindowTokens()`:

```js
const env = (typeof process !== 'undefined' && process.env) || {};
const envWindow = Number.parseInt(
  env.ECC_CONTEXT_WINDOW_TOKENS || env.CLAUDE_CODE_AUTO_COMPACT_WINDOW || '', 10);
if (Number.isInteger(envWindow) && envWindow > 0) return envWindow;
```

This adds an `ECC_CONTEXT_WINDOW_TOKENS` knob and also picks up Claude Code's own `CLAUDE_CODE_AUTO_COMPACT_WINDOW`, so it works with zero extra config for users who already set it. `formatWindowLabel()` already renders arbitrary sizes (`400k`), so no change is needed there.

## Test note

`tests/lib/transcript-context.test.js` should clear both `ECC_CONTEXT_WINDOW_TOKENS` and `CLAUDE_CODE_AUTO_COMPACT_WINDOW` for isolation — a globally-set `CLAUDE_CODE_AUTO_COMPACT_WINDOW` otherwise leaks into the default-window (200k) assertions and fails them.

## Environment

- ECC 2.0.0
- Claude Code on Opus 4.x (400k window), `CLAUDE_CODE_AUTO_COMPACT_WINDOW=400000` set

affaan-m:
[...]

PR #2465, opened 2026-07-06
fix(suggest-compact): recognize default-1M models without a [1m] marker (#2461)

### LINKED ISSUE #2461: suggest-compact: context % overstated on newer large-window models without a "[1m]" marker (e.g. claude-fable-5)
## Bug

`resolveContextWindowTokens()` in `scripts/lib/transcript-context.js` detects a large context window only via:
1. an explicit `ECC_CONTEXT_WINDOW_TOKENS` / `CLAUDE_CODE_AUTO_COMPACT_WINDOW` env override (added in #2290), or
2. a `[1m]` marker in the model id, or
3. the observed token count already exceeding 200k.

A model whose id carries none of these — a large-window model without the `[1m]` suffix, observed with `model: "claude-fable-5"`, and with usage still below 200k — falls through all three checks and is misclassified as a 200k window.

Observed in session: `[StrategicCompact] Context ~187k tokens (93% of 200k window) - consider /compact at the next logical boundary`, while the session's actual reported usage (from the CLI status line) was ~19%. 187k / 0.19 ≈ 984k, i.e. the real window is ~1M — the hook overstated usage by roughly 5x.

This is the same failure class as #2290 (400k Opus models falling through to the 200k default), but the env-override fix from #2290 does not help here unless the user has manually set `CLAUDE_CODE_AUTO_COMPACT_WINDOW`/`ECC_CONTEXT_WINDOW_TOKENS` — and nothing in the hook or its docs prompts a user to do so for a *new* model family. The `[1m]` marker and the "tokens > 200k" heuristic are both static/stale: neither scales to new model ids as Anthropic ships additional model families with large windows.

## Proposed fix

Add a small model-id-prefix → window-size table as a third fallback, checked before the final 200k default, so common large-window model families are recognized without requiring an env var:

```js
// scripts/lib/transcript-context.js
const KNOWN_MODEL_WINDOW_TOKENS = [
  // [id-substring, window-tokens] — checked in order, first match wins
  ['claude-fable-5', 1000000],
  ['claude-mythos-5', 1000000],
  // extend as new large-window model families ship
];

function resolveContextWindowTokens(tokens, model) {
  // ...existing env override and [1m] marker checks...

  if (typeof model === 'string') {
    const known = KNOWN_MODEL_WINDOW_TOKENS.find(([substr]) => model.includes(substr));
    if (known) return known[1];
  }

  if (Number.isFinite(tokens) && tokens > STANDARD_CONTEXT_WINDOW_TOKENS) {
    return LARGE_CONTEXT_WINDOW_TOKENS;
  }

  return STANDARD_CONTEXT_WINDOW_TOKENS;
}
```

Table entries are best-effort and expected to lag new releases, so this doesn't replace the env-override escape hatch — it just removes the "you must know to set an env var for your new model" trap for the common case. Also worth a doc callout (README/CHANGELOG) telling users on unlisted large-window models to set `CLAUDE_CODE_AUTO_COMPACT_WINDOW` explicitly.

## Test note

`tests/lib/transcript-context.test.js` should add a case for a model id that is large-window but carries neither `[1m]` nor exceeds 200k tokens (e.g. `claude-fable-5` at 187k tokens), asserting the window resolves correctly once the model is in the table — and asserting it still falls back to 200k for a genuinely-unknown small-window model at low token counts (no false positives).

## Environment

- ECC 2.0.0
- Claude Code on `claude-fable-5` (Claude 5 family / Mythos-class tier), large context window, no `[1m]` marker in the reported model id
- `CLAUDE_CODE_AUTO_COMPACT_WINDOW` / `ECC_CONTEXT_WINDOW_TOKENS` not set

Related: #2290 (same failure class, different model family — 400k Opus 4.x)

mc856:
## Summary
- `resolveContextWindowTokens()` only detected a large window via the env overrides (#2290), the `[1m]` model-id marker, or observed tokens already above 200k — newer families whose default window is 1M (e.g. `claude-fable-5`) carry no marker, so below 200k of usage the hook reported ~187k of real usage as 93% of a 200k window when real usage was ~19% of 1M (~5x overstated)
- add a small `KNOWN_MODEL_WINDOW_TOKENS` substring → window table, checked after the `[1m]` marker and before the >200k heuristic
- the table deliberately lists only families where 1M is the default rather than an opt-in (fable/mythos 5); opt-in families (Opus/Sonnet 4.x) keep signaling via `[1m]`, so 200k-default sessions on those are not overstated in the other direction
- env overrides remain the escape hatch for unlisted large-window models
- sync the window-detection description in both strategic-compact SKILL.md copies with the new detection path; the broader README callout suggested in the issue is left out deliberately — the env-override docs from #2290 already cover unlisted models

## Verification
- new test: `claude-fable-5` at 187k tokens resolves to the 1M window (so the hook reports ~19% instead of 93%)
- new test: an unknown model at the same token count still falls back to the 200k window (no false positives)
- `node tests/run-all.js` on node 22: 2973 passed, 0 failed

Fixes #2461

### DIFF
--- a/.agents/skills/strategic-compact/SKILL.md
+++ b/.agents/skills/strategic-compact/SKILL.md
@@ -31,7 +31,7 @@ Strategic compaction at logical boundaries:
 
 The `suggest-compact.js` script runs on PreToolUse (Edit/Write) and combines two signals:
 
-1. **Context size (primary)** — Reads the latest `usage` record from the session transcript (`transcript_path` in the hook payload) and sums `input_tokens + cache_read_input_tokens + cache_creation_input_tokens` (the true context size of the turn). Suggests `/compact` at a window-scaled threshold — 160k tokens on a 200k window, 250k on a 1M window (detected from a `[1m]` model marker, or inferred when observed tokens already exceed 200k) — and re-reminds after every additional 60k tokens of context growth
+1. **Context size (primary)** — Reads the latest `usage` record from the session transcript (`transcript_path` in the hook payload) and sums `input_tokens + cache_read_input_tokens + cache_creation_input_tokens` (the true context size of the turn). Suggests `/compact` at a window-scaled threshold — 160k tokens on a 200k window, 250k on a 1M window (detected from a `[1m]` model marker, a known default-1M model family, or inferred when observed tokens already exceed 200k) — and re-reminds after every additional 60k tokens of context growth
 2. **Tool-call count (secondary)** — Counts tool invocations in session; suggests at a configurable threshold (default: 50 calls), then every 25 calls after
 
 ## Hook Setup
--- a/scripts/lib/transcript-context.js
+++ b/scripts/lib/transcript-context.js
@@ -6,9 +6,9 @@
  *
  * - `input_tokens + cache_read_input_tokens + cache_creation_input_tokens`
  *   partition the prompt, so their sum is the true context size of the turn.
- * - The context window is detected from the model id (`[1m]` marker) or from
- *   the observed token count (anything above 200k implies a 1M window even
- *   when logs drop the suffix).
+ * - The context window is detected from the model id (`[1m]` marker or a
+ *   known large-window model family) or from the observed token count
+ *   (anything above 200k implies a 1M window even when logs drop the suffix).
  * - Thresholds are window-scaled and env-overridable; re-reminders fire in
  *   fixed token "buckets" above the threshold so the suggestion only repeats
  *   after real context growth.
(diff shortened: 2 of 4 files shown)


daltino (commented):
**DECISION: APPROVE**

This PR improves the detection logic for 1M context window models, accounting for default large-window model families without relying on the `[1m]` marker. The updates to code and tests are clear, targeted, and adhere to the repo's contribution guidelines. Tests demonstrate proper handling of various scenarios, ensuring robust functionality. Minor changes to documentation accompany the logic updates, keeping everything consistent. Good work!

mc856:
Dedup note: #2497 (opened Jul 11) overlaps on the transcript-context fix for #2461.

affaan-m:
Closing as superseded by merged #2468. Current main already contains the boundary-aware known-model detection and stronger 36-case regression coverage, including false-positive guards such as `claude-fable-5-mini`; reviving this branch would conflict in all four changed files and reintroduce broader substring matching. Thank you for identifying the default-1M model gap.

PR #2467, opened 2026-07-06
feat: Plan Canvas, a browser review canvas for plans

haelyra:
Plan Canvas is a local browser surface for reviewing plans. Your agent writes a plan, opens it in the browser, and you point at the exact thing you want changed, chat back, and hit Approve plan or Request changes. The agent blocks on one CLI call that returns your feedback as JSON, so the whole loop happens without you retyping "change this part" into the terminal.

Requested by the community. Inspired by [lavish-axi](https://github.com/kunchenguid/lavish-axi) by [@kunchenguid](https://github.com/kunchenguid). This is an original, ECC-native build of that idea rather than a port, credit to the original creator for the concept.

Plan Canvas reviewing a plan on the left while the agent works in the terminal on the right

## What it does

- Renders `.claude/plans/*.plan.md` (and any local `.md` or `.html`) in an ECC styled browser canvas
- Click any block or select text to attach an annotation, or just chat from the side rail
- Approve plan and Request changes verdicts map straight onto the `/plan` confirmation gate, so approval can happen in the browser instead of the terminal
- Live reload on save, so the agent edits the plan and you see it update
- Mermaid diagrams render themed to ECC, with an offline fallback to the source

## How it fits ECC

- CLI at `scripts/plan-canvas.js` (bin `ecc-plan-canvas`), modeled on the existing `control-pane` loopback server
- Loopback only with Host and Origin allowlisting (shared `scripts/lib/loopback-guard.js`, also adopted by control-pane)
- Zero new dependencies, plain CommonJS, matches the repo conventions
- Ships as a skill (`skills/plan-canvas`), a `/plan-canvas` command, and a SessionStart hook that resurfaces open reviews
- Registered across package.json, install manifests, catalog, command registry, and docs

## Cross harness

- The core is a CLI that emits JSON, so any harness that runs a shell command and reads stdout drives it the same way (Claude Code, Codex, Cursor, Gemini, OpenCode, Copilot)
- Codex surface added under `.agents/skills/plan-canvas` with an `openai.yaml` interface manifest, alongside peers like tdd-workflow and e2e-testing
- No Claude only dependency in the core loop, the SessionStart hook is an additive convenience

## Testing

- 130 plus new tests, including a full workflow end to end run through the real CLI and a real detached server (open, annotate, verdict, reply, live reload, end, reopen rules, stop)
- Whole gauntlet green: 3098 of 3098 tests, plus all CI validators, ESLint, and markdownlint
- Mermaid rendering verified with a real headless browser render

## Try it

```bash
ecc-plan-canvas open .claude/plans/your-feature.plan.md
ecc-plan-canvas await .claude/plans/your-feature.plan.md
```

## One note

Mermaid loads from a pinned CDN in the browser, the same approach lavish-axi takes, rather than vendoring a few MB into the package. It degrades to showing diagram source when offline and can point at a local mirror via `ECC_PLAN_CANVAS_MERMAID_URL`. Happy to switch to a vendored copy if we would rather be fully self contained.


daltino (approved):
This PR introduces "Plan Canvas," a handy browser-based tool for reviewing and annotating plans and visual artifacts. The implementation aligns well with contribution guidelines, includes updates to the necessary skill manifests and metadata, and adds steps for skill registration in the PR template. Changes to descriptions in marketplace and plugin JSON are consistent. Overall, this seems like a solid and thoughtful addition!

PR #2468, opened 2026-07-07
fix(suggest-compact): recognize large-window model families without a [1m] marker

### LINKED ISSUE #2461: suggest-compact: context % overstated on newer large-window models without a "[1m]" marker (e.g. claude-fable-5)

girish-kanjiyani7:
## Summary

Fixes #2461.

`resolveContextWindowTokens()` in `scripts/lib/transcript-context.js` only detected a 1M context window via (1) the `ECC_CONTEXT_WINDOW_TOKENS` / `CLAUDE_CODE_AUTO_COMPACT_WINDOW` env override, (2) a `[1m]` marker in the model id, or (3) observed tokens already exceeding 200k. Large-window models whose ids carry none of these — e.g. `claude-fable-5` at 187k tokens — fell through to the 200k default, and the strategic-compact hook reported ~93% usage when real usage was ~19% (roughly 5x overstated).

## Change

- Add a `KNOWN_MODEL_WINDOW_TOKENS` substring table (`claude-fable-5`, `claude-mythos-5` → 1M), checked after the env override and `[1m]` marker and before the tokens-above-200k heuristic, per the approach proposed in #2461.
- Substring match covers dated/region-prefixed ids (e.g. `us.anthropic.claude-fable-5-20260115-v1:0`).
- Precedence is unchanged for existing behavior: env overrides still win; unknown model ids still fall back to the 200k default (no false positives).

## Tests

Five new cases in `tests/lib/transcript-context.test.js` (written first, verified red → green):

- `claude-fable-5` at 187k tokens → 1M window (the reported repro)
- `claude-mythos-5` → 1M window
- dated/prefixed variant id → 1M window
- env override beats the table
- unknown model id at 187k tokens → stays 200k

`node tests/run-all.js`: 2976 passed, 0 failed. ESLint clean on both changed files.

### DIFF
--- a/scripts/lib/transcript-context.js
+++ b/scripts/lib/transcript-context.js
@@ -28,6 +28,16 @@ const DEFAULT_TRANSCRIPT_TAIL_BYTES = 256 * 1024;
 const MAX_TOKEN_SETTING = 10000000;
 const LARGE_WINDOW_MODEL_MARKER = '[1m]';
 
+// Known large-window model families whose ids carry no `[1m]` marker (#2461).
+// Substring match against the model id — covers dated/region-prefixed variants
+// (e.g. `us.anthropic.claude-fable-5-20260115-v1:0`). Checked in order, first
+// match wins. Best-effort and expected to lag new releases; the env override
+// remains the escape hatch for unlisted models.
+const KNOWN_MODEL_WINDOW_TOKENS = [
+  ['claude-fable-5', LARGE_CONTEXT_WINDOW_TOKENS],
+  ['claude-mythos-5', LARGE_CONTEXT_WINDOW_TOKENS]
+];
+
 /**
  * Read the trailing `tailBytes` of a file as UTF-8.
  * Returns null when the file is missing or unreadable.
@@ -132,9 +142,10 @@ function readLatestContextTokens(transcriptPath, options = {}) {
 
 /**
  * Detect the context window size for a turn.
- * 1M when the model id carries the `[1m]` marker, or when the observed token
- * count already exceeds the standard 200k window (covers logs that drop the
- * suffix); otherwise the standard 200k window.
+ * 1M when the model id carries the `[1m]` marker, matches a known large-window
+ * model family, or when the observed token count already exceeds the standard
+ * 200k window (covers logs that drop the suffix); otherwise the standard 200k
+ * window.
  */
 function resolveContextWindowTokens(tokens, model) {
   // Explicit window override wins: 400k models (e.g. Opus 4.x) match neither the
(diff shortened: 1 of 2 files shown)


daltino (approved):
This PR clearly addresses the issue of recognizing large-window model families without relying on the `[1m]` marker by introducing a lookup mechanism for known models. The code changes in `transcript-context.js` are well-commented, and the logic appears thoughtful, accounting for potential edge cases like prefixed/dated model IDs. The additions to the test suite provide good coverage for the new functionality, ensuring its correctness. Everything aligns with the contribution guidelines, and the change is well-documented.

PR #2479, opened 2026-07-09
feat(llm): add MiniMax provider

octo-patch:
Reason: add target provider/model to the existing provider registry.

## Summary
- Register MiniMax as an LLM provider with MiniMax-M3 as the default.
- Expose MiniMax-M3 and MiniMax-M2.7 with their verified context and modality metadata.
- Support configurable global and China endpoints through the existing `MINIMAX_BASE_URL`.
- Route OpenAI-compatible URLs to chat completions and Anthropic-compatible URLs to messages while normalizing the SDK base so the final path contains one `/v1/messages`.
- Reject empty responses consistently across both protocol paths.

## Test plan
- `.venv/bin/python -m pytest -q` (101 passed)
- Isolated Ruff lint, annotation, and format checks for the affected Python files
- `.venv/bin/mypy src/llm/providers/minimax.py`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,11 +37,13 @@ ATLAS_API_KEY=
 # ATLAS_BASE_URL=https://api.atlascloud.ai/v1
 # ATLAS_MODEL=anthropic/claude-sonnet-4.6
 
-# --- Optional: MiniMax (OpenAI-compatible) ----------------------------------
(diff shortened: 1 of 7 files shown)


octo-patch:
This PR has been superseded by #2510, rebuilt from the latest `main` with a clean single-commit history. Please continue review there.

Issue #2222, opened 2026-06-10
run-with-flags.js echoes truncated stdin on >1MB payloads, blocking all large-file edits


oleksiirek:
## Summary

`scripts/hooks/run-with-flags.js` caps stdin at `MAX_STDIN = 1024 * 1024` (1 MB). When the hook payload exceeds the cap, every fallthrough path still does `process.stdout.write(raw)` with the **truncated** string. The harness parses hook stdout as JSON, gets a document cut mid-stream, and treats it as a hook failure — blocking the tool call.

Net effect: any Edit/Write whose hook payload exceeds 1 MB (large source files, large markdown docs) is permanently blocked by every registered pre-write hook (`pre:write:doc-file-warning`, `pre:insaits-security`, `pre:governance-capture`, `pre:config-protection`, `pre:mcp-health-check`, ...), with errors like:

```
Hook "node .../run-with-flags.js pre:write:doc-file-warning ..." returned invalid JSON. The command was blocked for safety.
```

## Why the existing kill-switches don't help

The truncated-echo happens in `main()` **before** the `isHookEnabled()` check, so `ECC_HOOK_PROFILE=minimal` / `ECC_DISABLED_HOOKS=...` still emit the broken payload:

```js
async function main() {
  const [, , hookId, relScriptPath, profilesCsv] = process.argv;
  const { raw, truncated } = await readStdinRaw();   // raw may be truncated here

  if (!hookId || !relScriptPath) {
    process.stdout.write(raw);                       // <-- invalid JSON if truncated
    process.exit(0);
  }

  if (!isHookEnabled(hookId, { profiles: profilesCsv })) {
    process.stdout.write(raw);                       // <-- same
    process.exit(0);
  }
  ...
```

## Repro

```bash
python3 -c "import json,sys; sys.stdout.write(json.dumps({'tool_input':{'content':'x'*1200000}}))" \
  | node scripts/hooks/run-with-flags.js "pre:write:doc-file-warning" "scripts/hooks/doc-file-warning.js" "standard,strict"
# stdout ends mid-JSON -> harness rejects the hook output
```

Reproduced on v1.9.0 and v2.0.0 (the code is identical in both).

## Suggested fix

Fail open when the payload was truncated — empty stdout + exit 0 means "no opinion" in the hooks protocol:

```js
  const { raw, truncated } = await readStdinRaw();

  if (truncated) {
    process.stderr.write(
[...]

PR #2485, opened 2026-07-09
fix(hooks): stop echoing hook input to stdout — transcript bloat (100s of MB/week)

AlbertChiu777:
## Problem

Every wrapper layer in the hook chain echoes the full hook stdin JSON (`tool_input` + `tool_response` + `originalFile` content) back to stdout on the "no opinion" path. Claude Code records hook stdout in the session transcript, so each PostToolUse/PreToolUse call appends **18–75KB of duplicate JSON** — on an active project this added up to **~700MB of transcript bloat in one week** (`hook_success` 457MB + `async_hook_response` 245MB; `PostToolUse:Edit` alone 343MB), slowing session loading and post-hoc analysis.

None of these echoes are load-bearing: empty stdout + exit 0 already means "no opinion" to Claude Code — the same contract this repo's own truncation fix (#2222) and `bash-hook-dispatcher` comment ("Echoing the unmodified input event back to stdout fails Claude Code's hook-output JSON schema validation") rely on.

## Root causes (one per layer — all needed)

| Layer | Echo path |
|---|---|
| `run-with-flags.js` | 8 skip/disabled/dry-run/traversal/not-found/error paths echoed raw; `resolveHookResult()` forwarded a hook's `return rawInput` (the passthrough convention used by most hooks) verbatim; `resolveLegacySpawnStdout()` substituted raw for empty child stdout |
| `plugin-hook-bootstrap.js` | `passthrough()` echoed raw when the child produced no stdout; 3 early-exit paths echoed raw |
| `session-start-bootstrap.js` | same pattern on the SessionStart path |
| `auto-tmux-dev.js` | **the trigger for per-Bash echoes**: its passthrough returned `JSON.stringify(input)` instead of the raw input. Live hook stdin carries a trailing newline, so the re-serialization differs by one byte → `bash-hook-dispatcher` flags `rawModified` → the full input event is echoed on *every single Bash call* |
| `bash-hook-dispatcher.js` | belt-and-braces: never emit output byte-equal to the original input event |
| `hooks/hooks.json` | the 7 inline Stop/SessionEnd wrappers had `else process.stdout.write(raw)` + an unresolved-root echo fallback |

## What still flows through (verified)

- Blocking: `config-protection` exit 2 + stderr — unchanged
- `permissionDecision` deny JSON (gateguard) — forwarded
- `additionalContext` injection (`doc-file-warning`, `ecc-context-monitor`) — forwarded
- `auto-tmux-dev` dev-server rewrite — a real modification, still emitted

## Verification

- 74KB synthetic `PostToolUse:Edit` payload through the full chain (bootstrap → run-with-flags → hook): **0 bytes stdout, exit 0** (previously echoed all 74KB)
- Live before/after instrumentation: PreToolUse bash dispatcher stdout went from 653–1697 bytes per Bash call to **0**
- Full suite: `node tests/run-all.js` — 3105/3107 pass; the 2 remaining failures (`install-apply` ENOBUFS, env-dependent) reproduce identically on a clean `main` checkout
- `npx eslint` clean on all touched files

## Test updates

Allow/skip paths are now asserted as **silent** (`stdout === ''`) instead of raw echo, per the new contract. Added regressions: verbatim passthrough for trailing-newline stdin (`auto-tmux-dev`), and "a hook with a real opinion still emits its output" (`run-with-flags`).


mc856 (approved):
Verified locally on macOS (node 22):

- Full suite on the PR head: 3105/3106. The single failure (`install-apply` "full profile dry-runs include delivery-gate", spawnSync ENOBUFS) reproduces identically on a clean upstream/main checkout — env-dependent, not this diff.
- The byte-equality guards (`stdout !== raw`) assume hooks pass the original string through. Checked that assumption across the tree: `git grep` over `scripts/hooks/` shows auto-tmux-dev was the only hook re-serializing (`JSON.stringify(input)`); every other passthrough returns `rawInput` verbatim, so the guards hold for the current hook set.
- Spot-checked the opinion paths: config-protection exit 2, gateguard deny JSON, and additionalContext injection all emit stdout ≠ raw and flow through `resolveHookResult` unchanged.

The trailing-newline diagnosis in auto-tmux-dev also explains why the echo fired on every single Bash call. LGTM.

mc856 (commented):
#2494 landed on the 19th and it moves the ground under this one — quietly, since it doesn't surface as a conflict.

That PR routed PostToolUse off `run-with-flags` onto the new `posttooluse-dispatcher`, and main's `hooks/hooks.json` now sets `ECC_POSTTOOLUSE_PASSTHROUGH='1'` inline in both dispatcher commands (sync and async). The dispatcher honors it:

```js
function resolveMainStdout(raw, result, options = {}) {
  if (result.stdout) return result.stdout;
  if (options.truncated || result.exitCode !== 0 || !options.passthrough) return '';
  return raw;
}
```

So on the no-opinion path PostToolUse still echoes the raw payload on main today — the behavior this PR removes everywhere else. It was the right call in #2494 (parity with main as it stood), but it wants undoing when this one lands.

Two consequences:

- This PR's `hooks.json` diff only touches the Stop and SessionEnd entries, so it merges clean and stays green — but `PostToolUse:Edit`, the 343MB of your 700MB, keeps echoing afterwards.
- The verification bullet above ("74KB synthetic `PostToolUse:Edit` payload through the full chain (bootstrap → run-with-flags → hook): 0 bytes stdout") no longer describes the shipping path — PostToolUse doesn't reach `run-with-flags` anymore.

The fix is dropping `process.env.ECC_POSTTOOLUSE_PASSTHROUGH='1';` from those two commands in `hooks/hooks.json` — the dispatcher treats unset as no-echo (strict `=== '1'`), so nothing else has to change. Worth re-running that 74KB check through the dispatcher after.

Flagging since my approval predates #2494 and doesn't cover any of this.

haelyra:
Thank you for this! The transcript-bloat fix is a really useful direction, and we appreciate the careful work across the hook paths. One small thing needs to be updated before we can merge it: after #2494, the PostToolUse dispatcher still sets `ECC_POSTTOOLUSE_PASSTHROUGH='1'` in `hooks/hooks.json`, so please remove that from both PostToolUse dispatcher commands and rerun the focused PostToolUse/no-echo check. Once that is handled, this should be much easier for us to move forward.

affaan-m (changes requested):
This no longer fixes the shipping PostToolUse path after #2494. Current main routes PostToolUse through posttooluse-dispatcher and both registered commands set ECC_POSTTOOLUSE_PASSTHROUGH=1, so no-opinion Edit responses still echo the raw payload. Please rebase onto current main (including newly merged #2493), remove the two passthrough env assignments, reconcile the Stop/no-op tests with #2493, add a dispatcher-level 74KB no-echo regression, and run the current full CI matrix. The existing approval predates #2494 and the head has not changed since.

haelyra:
Thank you again for the careful production diagnosis here, and apologies for how long this sat. The transcript-bloat problem is real and the measurements in this PR are genuinely useful. Since this branch was opened, #2494 moved the main PostToolUse path to posttooluse-dispatcher, #2493 changed the Stop wrappers, and the current head now conflicts without fixing the shipping PostToolUse passthrough. The focused bootstrap portion also overlaps #2380.

Because bringing this branch current now requires rebuilding and revalidating several hook boundaries rather than a small patch, we’re going to close this implementation and carry the no-echo work into the maintainer task queue with attribution. A future focused patch against the current dispatcher—including the 74 KB no-echo regression and proof that real deny/context output still passes through—would still be welcome.

PR #2486, opened 2026-07-09
fix(hooks): context-monitor noise — loop-detection false positives + per-call cost-warning spam

AlbertChiu777:
## Problem

Two independent noise sources in the PostToolUse context monitor (`ecc-context-monitor.js`) inject agent-facing warnings into context on nearly every tool call. Both were observed live: a `LOOP WARNING: Tool called N times with same parameters` that fired for consecutive Bash calls with *different* content, and a `COST NOTICE` re-injected on almost every call once a session crossed $5.

### 1. LOOP WARNING false positives

- `hashToolCall()` in `ecc-metrics-bridge.js` hashed only `command.slice(0, 160)` for Bash — distinct long commands sharing a 160-char prefix (heredocs, long one-liners) collide, so consecutive **different** calls look like a stuck loop. Fixed by hashing the full command and truncating the *digest*, exactly the treatment the Edit/Write branch already received for the same bug class (its comment even warns about slicing input).
- `LOOP_THRESHOLD = 3` against the 5-entry `recent_tools` ring buffer fires on legitimate repetition (retries, polling). Raised to 5 = *all* of the last 5 calls must be identical. The warning text ("in last 5 calls") already said 5.

### 2. COST NOTICE / warning spam

`run()` dedupes on exact message text, but the dollar figure embedded in the text moves on nearly every call — so text-based dedupe considers each one "new information" and re-emits per tool call for the rest of the session. Context warnings have the same defect via the remaining-% figure.

Fixed by deduping on a stable **per-tier key** (`cost:notice|warning|critical`, `context:warning|critical`, `scope`; loop messages are already stable keys): each tier fires exactly once, and re-fires only on genuine escalation (notice → warning → critical, escalation-to-critical fast path unchanged). The existing `ECC_CONTEXT_MONITOR_COST_WARNINGS` opt-out is untouched.

State file field `lastMessage` → `lastKey`; legacy state files simply cause one re-emit after upgrade, then converge.

## Tests

- Loop: 5-of-5 identical fires, 4-of-5 does not (new); threshold test updated
- Hash: long shared-prefix Bash commands (diverging only after 160 chars) hash differently (new, mirrors the existing >2048-char Write test)
- Dedupe: run()-level tier test — notice fires once, stays silent as cost ticks within the tier, re-emits on tier escalation (new)
- Full suite: 3104/3105; the 1 failure (`install-apply` ENOBUFS) reproduces identically on a clean `main` checkout
- `npx eslint` clean on all touched files

Sibling PR #2485 addresses the transcript-bloat echo in the same hook chain; the two are file-disjoint and independently mergeable.


affaan-m (dismissed):
The overall fix and regression coverage look sound. One project-level blocker remains before merge: this path mutates the object returned by `readWarnState` in place:

```js
warnState.lastSeverity = topSeverity;
warnState.lastKey = dedupeKey;
writeWarnState(sessionId, warnState);
```

ECC marks immutability as a critical coding rule. Please construct the next state instead:

```js
writeWarnState(sessionId, {
  ...warnState,
  lastSeverity: topSeverity,
  lastKey: dedupeKey,
});
```

That keeps state transitions explicit and avoids coupling this caller to whether `readWarnState` ever returns a cached/shared object. I found no other merge blocker; this should be ready for another look once that small change lands.

haelyra:
Thank you for digging into both of these noise paths, and sorry this sat for so long. Hashing the full Bash input and deduplicating by warning tier are the right fixes. I refreshed the branch onto current main and carried the small immutable warn-state update in 31e60cf7.

The focused context-monitor suite passes 22/22 and the metrics-bridge suite passes 23/23. Required hosted CI is now running on the exact updated head; we will merge it through the serial main-branch gate once green.

PR #2492, opened 2026-07-10
fix(skill-stocktake): Windows Git Bash/MSYS compatibility for stocktake scripts + two skill-doc fixes

EricT1230:
## Summary

Fixes six Windows (Git Bash / MSYS) compatibility bugs in the skill-stocktake scripts, plus two skill-doc defects found while running a full stocktake of 186 installed skills on Windows 11.

### skill-stocktake scripts

| Bug | Symptom on Windows | Fix |
|---|---|---|
| Process substitution `<(...)` fed to jq (`scan.sh`, `save-results.sh`) | Native jq.exe cannot open MSYS `/proc/<pid>/fd` paths → `Could not open file /proc/NNN/fd/63` | Feed via stdin (`scan.sh`) / real temp file (`save-results.sh`) |
| `--argjson skills "$all_skills"` with a full inventory (`scan.sh`) | `Argument list too long` (~32KB argv limit) | Pass the array via stdin |
| `dp="${file/#$HOME/~}"` canonicalization (`scan.sh`, `quick-diff.sh`) | The bare `~` in the replacement is tilde-expanded back to `$HOME` (no-op), while MSYS argument conversion turns POSIX paths into `C:/`-style before jq stores them → quick-diff never matches stored paths and reports **every** file as new | `cygpath -m` on Windows; escaped `\~` fallback elsewhere |
| Global dir == project dir (`scan.sh`, `quick-diff.sh`) | Invoking from `$HOME` makes `$PWD/.claude/skills` resolve to the global dir → every skill double-counted (242 files reported as 484) | Skip the project pass when both resolve to the same physical dir |
| Auxiliary `.md` assets (`quick-diff.sh`) | `results.json` records one entry per skill (its SKILL.md), but the scanner enumerates all `*.md` → `references/`, `rules/`, `agents/` files are forever misreported as `is_new` | Judge known-ness per skill via its SKILL.md anchor |

### Skill docs

- **browser-qa**: tool family was written as `mChild__claude-in-chrome__*`; the real prefix is `mcp__claude-in-chrome__*`, so an agent following the skill literally could never invoke the browser tools.
- **autonomous-agent-harness**: the Setup Guide instructed installing `@anthropic/memory-mcp-server`, `@anthropic/scheduled-tasks-mcp-server`, `@anthropic/computer-use-mcp-server` and POSTing to `https://api.anthropic.com/dispatch` — none of these npm packages or the hosted endpoint exist. Rewritten to point at what actually ships (Claude Code built-in memory, native scheduled tasks / OS schedulers with headless `claude -p`, Claude in Chrome or Playwright MCP, and `claude -p` inside CI for the dispatch pattern).

## Test plan

- [x] `bash -n` clean on all three scripts
- [x] End-to-end on Windows 11 Git Bash + jq 1.8.2 against a 186-skill install: `scan.sh` produces a correct single-count inventory; `quick-diff.sh` returns `[]` on a fresh baseline, and exactly the edited files (no false positives) after edits; `save-results.sh` merges and stamps `evaluated_at` correctly
- [x] `node tests/run-all.js`: 3037/3049 pass; the 12 failures (block-no-verify, config-protection, gateguard suites) reproduce identically on a clean checkout of `main` on this machine, i.e. pre-existing local-environment failures unrelated to this change (no test coverage exists for the stocktake bash scripts)
- [ ] Behavior on macOS/Linux is unchanged by design (`command -v cygpath` gates every Windows path; verifying on a POSIX runner would be a welcome double-check)

Side note: while committing, the local block-no-verify hook falsely rejected a commit message containing the text "bash -n clean" — a live reproduction of the failing `allows -n discussed in a quoted commit message` test case in the suite.

### DIFF
--- a/skills/autonomous-agent-harness/SKILL.md
+++ b/skills/autonomous-agent-harness/SKILL.md
@@ -119,13 +119,13 @@ Trigger Claude Code agents remotely for event-driven workflows.
 **Dispatch patterns:**
 
 ```bash
-# Trigger from CI/CD
-curl -X POST "https://api.anthropic.com/dispatch" \
-  -H "Authorization: Bearer $ANTHROPIC_API_KEY" \
-  -d '{"prompt": "Build failed on main. Diagnose and fix.", "project": "/repo"}'
+# Trigger from CI/CD — run Claude Code headless inside the pipeline
+# (e.g., a GitHub Actions step; there is no hosted Anthropic "dispatch" endpoint)
+claude -p "Build failed on main. Read the CI logs in ci-failure.log, diagnose, and fix." \
+  --allowedTools "Read,Edit,Bash,Grep,Glob"
 
 # Trigger from webhook
-# GitHub webhook → dispatch → Claude agent → fix → PR
+# GitHub webhook → your listener → spawns `claude -p` in the repo → fix → PR
 
 # Trigger from another agent
 claude -p "Analyze the output of the security scan and create issues for findings"
@@ -186,24 +186,29 @@ description: Persistent task queue for autonomous operation
 
 ## Setup Guide
 
-### Step 1: Configure MCP Servers
+### Step 1: Wire Up Memory, Scheduling, and Browser Control
 
-Ensure these are in `~/.claude.json`:
+Use the capabilities that actually ship with Claude Code — Anthropic does not
+publish npm packages for these:
+
+- **Memory**: Claude Code's built-in persistent memory
+  (`~/.claude/projects/<project>/memory/` + `MEMORY.md` index). For a graph
+  layer, add a community memory MCP server of your choice to `mcpServers`
+  after reviewing its source.
+- **Scheduling**: Claude Code's native scheduled/background tasks (Cron tools
+  in supported clients) or plain OS schedulers (Windows Task Scheduler,
+  cron/launchd) invoking `claude -p "..."`.
+- **Browser / computer control**: the Claude in Chrome extension
+  (`mcp__claude-in-chrome__*` tools) or a Playwright MCP server.
+
+Example `mcpServers` entry (only add servers you have vetted):
 
 ```json
 {
   "mcpServers": {
-    "memory": {
+    "playwright": {
       "command": "npx",
-      "args": ["-y", "@anthropic/memory-mcp-server"]
-    },
-    "scheduled-tasks": {
-      "command": "npx",
-      "args": ["-y", "@anthropic/scheduled-tasks-mcp-server"]
-    },
-    "computer-use": {
-      "command": "npx",
-      "args": ["-y", "@anthropic/computer-use-mcp-server"]
+      "args": ["-y", "@playwright/mcp@latest"]
     }
   }
 }
--- a/skills/browser-qa/SKILL.md
+++ b/skills/browser-qa/SKILL.md
@@ -98,7 +98,7 @@ credentials/tokens/PII before saving any screenshot.
 ## Integration
 
 Works with any browser MCP:
-- `mChild__claude-in-chrome__*` tools (preferred — uses your actual Chrome)
+- `mcp__claude-in-chrome__*` tools (preferred — uses your actual Chrome)
 - Playwright via `mcp__browserbase__*`
 - Direct Puppeteer scripts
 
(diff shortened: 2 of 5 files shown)


mc856 (commented):
The Windows-side diagnosis checks out — reproduced on bash 5.3: an unquoted `~` in the patsub replacement gets re-expanded to `$HOME` (bash 5.2 changed how the replacement string is expanded). But `\~` only works on >= 5.2: on bash <= 5.1 the backslash survives as a literal `\~`. macOS ships /bin/bash 3.2 and these scripts use `#!/usr/bin/env bash`, so on a stock Mac quick-diff now emits `"path": "\~/…", "is_new": true` for every already-assessed skill — reproduced end-to-end with a seeded results.json baseline that matches fine on main. So as written this moves the bug from one bash family to the other rather than fixing it.

A form that behaves the same on 3.2 and 5.3 (tested on both):

```bash
tilde='~'
dp="${file/#$HOME/$tilde}"
```

(the result of a variable expansion is never subject to tilde expansion, in any bash). Same change applies at scan.sh:121 and quick-diff.sh:71 and :90.

The other five Windows fixes and both doc corrections verified fine. Small aside: the zh-CN mirrors of the two corrected docs still carry the old text (docs/zh-CN/skills/browser-qa/SKILL.md:77, docs/zh-CN/skills/autonomous-agent-harness/SKILL.md:123).

affaan-m (changes requested):
Thanks for tackling the Windows path issue. I reproduced a blocking cross-platform regression in the current implementation:

1. `skills/skill-stocktake/scripts/quick-diff.sh` and `scan.sh` use `${value/#$HOME/\~}` as the non-Windows fallback. Stock macOS Bash 3.2 preserves the backslash, producing `\~/.claude/...` instead of the stored `~/.claude/...` form. Please replace through a literal variable (for example, `tilde='~'; display_path=${value/#$HOME/$tilde}`) and add coverage for Bash 3.2-compatible behavior.\n2. The autonomous harness example feeds attacker-influenceable CI logs to a write-capable `claude -p` session (`Read,Edit,Bash,Grep,Glob`). Please split this into a read-only, secret-free diagnosis pass followed by a separately approved constrained remediation step.\n3. Pin the `@playwright/mcp` example to an exact vetted version instead of `@latest`, consistent with the repository supply-chain policy.\n4. Update the zh-CN autonomous-agent-harness mirror, which still contains the removed dispatch endpoint.\n\nThe core Windows normalization direction remains useful; these items need resolution before merge.

haelyra:
Thank you again for the detailed Windows investigation, and I’m sorry this has waited so long. The process-substitution, argv-limit, duplicate-directory, and path-identity fixes are valuable, and the browser tool-prefix correction is also valid.

The current head still has the previously requested blockers:

- **Preserve stock macOS Bash path identity.** The non-Windows ${value/#$HOME/\\~} replacement produces a literal backslash on Bash 3.2. Use a variable containing the literal tilde and add focused cross-version coverage so stored paths remain ~/.claude/....
- **Handle project-only installs quietly.** Both stocktake scripts cd into the global directory without confirming it exists. Skip that comparison when the directory is absent so a project-only setup does not emit a spurious error.
- **Keep CI diagnosis read-only and isolated.** The current autonomous-harness example gives attacker-influenceable logs Edit and Bash access. Separate untrusted-log diagnosis from any explicitly approved remediation.
- **Pin and bound MCP dependencies.** Replace mutable @playwright/mcp@latest and vague community-server guidance with a vetted exact version and least-privilege boundaries.
- **Update the zh-CN mirror.** It still documents the nonexistent Anthropic dispatch endpoint and package flow.

If you do not have bandwidth for another pass, that is completely understandable—tell us and we can carry the useful Windows work into the coordinated stocktake task queue with #2610, #2640, and #2729.

chenhz01:
The `cygpath -m` fix corrects new writes but breaks reads of everything already stored: existing saved results hold MSYS/POSIX-style paths, so the first `quick-diff.sh` run after this lands reports every previously-scanned file as new — the exact false-positive class the PR fixes, reintroduced by migration. Suggest a format-version key in the saved JSON plus a one-shot normalizer (or matching on both path forms during a transition window). Second, same collision note from the other direction: #2729 rewrites `scan.sh` and touches `save-results.sh` for the same MSYS symptoms via a different mechanism (streamed JSON, zero-padded tmp names), so the two PRs need an explicit merge-order decision before either lands. Happy to take a crack at the migration shim if useful.

Issue #2514, opened 2026-07-14
bug(hooks/pre-bash-tmux-reminder): misplaced `?` makes every `yarn <anything>` trigger the tmux reminder


thejesh23:
### Summary

The tmux-reminder hook uses a regex where the `?` on the yarn subcommand group is misplaced. As written, the yarn branch of the alternation matches every command that starts with `yarn ` (including `yarn build`, `yarn dev`, `yarn add`, `yarn workspaces …`) — not just `yarn install` / `yarn test` as intended by the sibling branches.

### Location

`scripts/hooks/pre-bash-tmux-reminder.js:16`

```js
/(npm (install|test)|pnpm (install|test)|yarn (install|test)?|bun (install|test)|cargo build|make\b|docker\b|pytest|vitest|playwright)/
```

The `?` after `(install|test)` applies only to the yarn subcommand group, so:

- `yarn (install|test)?`  =  `yarn ` followed by optional `install|test`  =  matches `yarn ` + anything.

Every other package-manager branch (`npm`, `pnpm`, `bun`) requires the subcommand — only yarn is over-broad.

### Expected

Only genuinely long-running commands like `yarn install` and `yarn test` (parity with `npm (install|test)`, `pnpm (install|test)`, `bun (install|test)`) should trigger the "Consider running in tmux" reminder.

### Actual

Every `yarn <anything>` call outside tmux surfaces the reminder in additional context. This spams the additional-context channel for very short commands like `yarn --version`, `yarn add foo`, `yarn workspaces list`, etc.

### Reproduction

```bash
node -e "
const {run} = require('/tmp/ECC/scripts/hooks/pre-bash-tmux-reminder.js');
delete process.env.TMUX;
for (const cmd of ['yarn install', 'yarn test', 'yarn add react', 'yarn build', 'yarn dev', 'yarn --version']) {
  const r = run(JSON.stringify({ tool_input: { command: cmd } }));
  console.log(cmd.padEnd(20), '->', typeof r === 'object' ? 'REMINDER FIRED' : 'no reminder');
}
"
```

All six lines print `REMINDER FIRED`, but only the first two should.

### Suggested fix

Drop the stray `?`:

```diff
-/(npm (install|test)|pnpm (install|test)|yarn (install|test)?|bun (install|test)|cargo build|make\b|docker\b|pytest|vitest|playwright)/
[...]

Issue #2513, opened 2026-07-14
bug(ecc_dashboard): `Refresh Data` doesn't repopulate Rules or Commands trees


thejesh23:
### Summary

Clicking **Refresh Data** in the ECC Dashboard updates the tab-count labels and the status bar, but the Rules and Commands trees are never repopulated. The dialog then reports "Data refreshed successfully!" — so a user sees a success message while the two trees still show pre-refresh rows.

### Location

`ecc_dashboard.py`, `refresh_data()` (around lines 854-877):

```python
def refresh_data(self):
    self.project_path = self.path_entry.get()
    self.agents = load_agents(self.project_path)
    self.skills = load_skills(self.project_path)
    self.commands = load_commands(self.project_path)
    self.rules = load_rules(self.project_path)

    self.notebook.tab(0, text=f"Agents ({len(self.agents)})")
    self.notebook.tab(1, text=f"Skills ({len(self.skills)})")
    self.notebook.tab(2, text=f"Commands ({len(self.commands)})")
    self.notebook.tab(3, text=f"Rules ({len(self.rules)})")

    # Repopulate
    self.populate_agents(self.agents)
    self.populate_skills(self.skills)
    # <-- self.populate_rules(self.rules) is missing
    # <-- self.command_tree is not repopulated at all

    ...
    messagebox.showinfo("Success", "Data refreshed successfully!")
```

`populate_rules()` already exists (used by `create_rules_tab` and `filter_rules`) but is never called from `refresh_data`. For Commands there is no reusable populate method at all — the initial population is inlined in `create_commands_tab` (~lines 659-661) with no equivalent in the refresh path.

### Expected

After **Refresh Data**:
- Agents tree — refreshed (already works)
- Skills tree — refreshed (already works)
- **Rules tree — refreshed (currently broken)**
- **Commands tree — refreshed (currently broken)**

### Actual

Rules and Commands trees show stale rows from initial load. Tab counts update correctly, which makes the divergence between "N Commands" in the tab label and the actual visible rows especially confusing.

### Reproduction

1. Launch `python ecc_dashboard.py`
2. Add or remove a file under `commands/` or `rules/` (or point Project Path at a different repo)
3. Click **Refresh Data**
4. Look at the Agents and Skills tabs — rows updated
5. Look at the Commands and Rules tabs — rows still stale, but tab counts and "Data refreshed successfully!" popup both lie about it

### Suggested fix

[...]

PR #2515, opened 2026-07-14
fix(llm/providers/claude): attach cache_control to system block, not top-level

### LINKED ISSUE #2512: bug(llm/providers/claude): `cache_control` passed as top-level Messages API param
### Summary

`ClaudeProvider.generate()` sets `cache_control` as a top-level parameter to `Anthropic().messages.create()`. `cache_control` is a **per-content-block** field on the Anthropic Messages API — it must be attached to a text block, tool definition, or system prompt block, not to the top-level request. Passing it top-level causes the SDK/API to reject the call, so every invocation of the Claude provider fails.

### Location

`src/llm/providers/claude.py` around line 70:

```python
params: dict[str, Any] = {
    "model": model,
    "messages": api_messages,
    "max_tokens": input.max_tokens if input.max_tokens else 16000,
    "cache_control": {"type": "ephemeral"},   # <-- invalid at top level
}
...
response = self.client.messages.create(**params)
```

### Expected

`cache_control` should either be removed (if prompt caching isn't wanted here) or attached to the last content block of the system prompt / to tool definitions per the Anthropic docs:

```python
system=[{"type": "text", "text": sys_text, "cache_control": {"type": "ephemeral"}}]
```

### Actual

- Anthropic Python SDK's `messages.create()` doesn't accept `cache_control` as a keyword arg, so calls raise `TypeError` (or the API returns `400 unknown_parameter` depending on SDK version).
- The provider never returns a successful response — the whole Claude codepath is broken.

### Reproduction

```python
from llm.providers.claude import ClaudeProvider
from llm.core.types import LLMInput, Message, Role

provider = ClaudeProvider(api_key="sk-ant-...")
provider.generate(LLMInput(messages=[Message(role=Role.USER, content="hi")]))
# -> Error (top-level cache_control is not a valid Messages API param)
```

The existing unit tests in `tests/test_claude_provider.py` don't catch this because `FakeClient.messages.create(**_params)` accepts any kwargs and ignores them.

### Notes

- Other providers in `src/llm/providers/` (openai, ollama, astraflow, atlas) don't set any equivalent field, so this isn't a repo-wide pattern.
- Fix should keep `cache_control` only where the API accepts it (or drop it), and add a regression test that asserts the kwargs actually sent to `messages.create` don't include a top-level `cache_control`.

Severity: **Major** — provider is non-functional on every request.

thejesh23:
## What Changed

- Removed the top-level `cache_control: {\"type\": \"ephemeral\"}` from the params dict in `ClaudeProvider.generate()`.
- When a system prompt is present, `system` is now sent as a list of content blocks with `cache_control` attached to the (single) block, preserving the ephemeral prompt-caching behavior via the API's supported shape.
- Hardened the unit test scaffold: `FakeMessages.create` now records `last_params`, and two new regression tests assert:
  1. `cache_control` never appears as a top-level param.
  2. When a system prompt is set, `cache_control` rides on the last system content block.

## Why This Change

The Anthropic Messages API does not accept `cache_control` as a top-level request parameter — it's a per-content-block field. Passing it top-level raises `TypeError` in the Anthropic Python SDK (kwargs validated against `messages.create()`'s signature) or a `400 unknown_parameter` from the API. Every `ClaudeProvider.generate()` call was failing on the wire.

Existing tests didn't catch this because `FakeMessages.create(**_params)` accepted any kwargs and ignored them.

## Testing Done

- [x] Automated tests pass locally — `pytest tests/test_claude_provider.py -v` shows **6/6 passed** (4 existing + 2 new regression tests).
- [x] Edge cases considered and tested — the second regression test asserts the "no system prompt" path does not sneak `cache_control` back in at the top level.

## Type of Change

- [x] `fix:` Bug fix

## If you changed dependencies or `package.json`

No dependency changes.

Not applicable — pure library bugfix in `src/llm/providers/claude.py`.

## Documentation

No documented behavior changed. The public `ClaudeProvider.generate()` signature is unchanged; the wire-level `system` shape now uses the block form the API expects.

## Linked issue

Fixes #2512

---

Related independent PRs opened alongside this one (different files, no conflicts):
- fix(ecc_dashboard): repopulate Rules and Commands trees on Refresh Data — #2513
- fix(hooks): remove stray '?' in tmux-reminder yarn regex — #2514

### DIFF
--- a/src/llm/providers/claude.py
+++ b/src/llm/providers/claude.py
@@ -67,10 +67,15 @@ class ClaudeProvider(LLMProvider):
                 "model": model,
                 "messages": api_messages,
                 "max_tokens": input.max_tokens if input.max_tokens else 16000,
-                "cache_control": {"type": "ephemeral"},
             }
             if system_parts:
-                params["system"] = "\n\n".join(system_parts)
+                params["system"] = [
+                    {
+                        "type": "text",
+                        "text": "\n\n".join(system_parts),
+                        "cache_control": {"type": "ephemeral"},
+                    }
+                ]
             if input.tools:
                 params["tools"] = [tool.to_anthropic_tool() for tool in input.tools]
             if not _uses_adaptive_thinking_only(model):
--- a/tests/test_claude_provider.py
+++ b/tests/test_claude_provider.py
@@ -135,7 +135,8 @@ def test_generate_does_not_pass_cache_control_as_top_level_param() -> None:
     # When a system prompt is present, cache_control should ride on the last
     # system content block so ephemeral prompt caching still works.
     system = params.get("system")
-    assert isinstance(system, list) and system, "system should be a non-empty list of blocks"
+    assert isinstance(system, list), "system should be sent as a list of content blocks"
+    assert system, "system content-block list should not be empty"
     assert system[-1].get("cache_control") == {"type": "ephemeral"}
 
 
--- a/tests/test_claude_provider.py
+++ b/tests/test_claude_provider.py
@@ -10,8 +10,10 @@ from llm.providers.claude import ClaudeProvider
 class FakeMessages:
     def __init__(self, response: SimpleNamespace) -> None:
         self.response = response
+        self.last_params: dict[str, Any] = {}
 
-    def create(self, **_params: object) -> SimpleNamespace:
+    def create(self, **params: object) -> SimpleNamespace:
+        self.last_params = dict(params)
         return self.response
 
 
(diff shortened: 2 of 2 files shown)


thejesh23:
Addressed CodeRabbit review in `5bfd9f2`: split the composite `isinstance(system, list) and system` assertion into two focused asserts to satisfy Ruff PT018. 6/6 tests still pass locally; `ruff check --select PT018` is clean.

rjjelley:
Local review from Codex automation: this is the strongest fix for #2512. It removes the invalid top-level Anthropic `cache_control` parameter, preserves prompt caching by attaching cache control to the structured system block, and includes regression coverage. Focused local check passed: `tests/test_claude_provider.py` 6/6. I attempted to merge with expected head `5bfd9f2310fede6a00d66124c00f7d6b5255f4ba`, but the GitHub integration lacks merge permission (`403 Resource not accessible by integration`). Recommendation: merge as the canonical fix for #2512.

Issue #2506, opened 2026-07-12
Two hardening gaps in new scripts: dashboard-web binds all interfaces; agent-data-home lacks project-config containment


Michael-WhiteCapData:
Vendoring a recent `main` into a downstream config repo, we security-reviewed the new `scripts/` and found two missing-containment gaps worth hardening. Neither is a backdoor and both are in opt-in features, but each is exploitable in the wrong environment. Sharing upstream so they can be fixed at source.

### 1. `scripts/dashboard-web.js` — binds all interfaces, no auth (HIGH)
`server.listen(PORT, cb)` is called with no host arg, so Node binds `0.0.0.0`/`::` (all interfaces) rather than loopback. Unlike `scripts/lib/plan-canvas/server.js` (which binds `127.0.0.1` and enforces `loopback-guard.js` Host/Origin checks), the dashboard has no Host validation, no Origin check, and no auth. It serves agent prompt bodies, skill/command/rule content, and MCP server topology (env **names**; values are masked). Anyone on the same LAN/VPN segment can browse it uncredentialed.
**Suggested fix:** bind `process.env.ECC_DASHBOARD_HOST || '127.0.0.1'`, and/or reuse `loopback-guard.js` the way `plan-canvas/server.js` does.

### 2. `scripts/lib/agent-data-home.js` — project-local config can redirect the data-persistence root (HIGH)
`readProjectConfigAt()` reads `<projectRoot>/.cursor/ecc-agent-data.json` (attacker-controllable in any cloned/forked repo) and takes its `agentDataHome` verbatim through `expandHomePath()`, which honors an absolute path unchanged or resolves a relative one (incl. `../`) against the project with no upper-bound check — none of the `assertWithinTrustedRoot`/`isWithinRoot` containment that `scripts/lib/path-safety.js` applies elsewhere (and which cites the same GHSA-hfpv-w6mp-5g95 class). It's reached automatically via the `scripts/hooks/cursor-session-env.js` SessionStart hook. A malicious repo shipping a relative `agentDataHome` can silently redirect ECC's session-data writes into its own tree, where a later `git commit` would exfiltrate them; an absolute value redirects writes outside the project.
**Suggested fix:** resolve `agentDataHome` and require containment within an allow-listed root (home subtree or explicit opt-in), mirroring `path-safety.js`.

[...]

PR #2518, opened 2026-07-14
fix(security): harden dashboard-web — loopback-only bind + Host/Origin guard (#2506)

pangerlkr:
Closes gap 1 of #2506.

## Problem
`scripts/dashboard-web.js` was binding to all interfaces (0.0.0.0 / :: / OS default) with no Host or Origin validation, making the dashboard reachable from any network interface and vulnerable to DNS-rebinding attacks from local malicious pages.

### 1. Loopback-only bind by default
```js
const HOST = process.env.ECC_DASHBOARD_HOST || '127.0.0.1';
server.listen(PORT, HOST, () => { ... });
```
Users who need LAN/container access can set `ECC_DASHBOARD_HOST=0.0.0.0`; the default posture is now localhost-only.

### 2. Host header guard (DNS-rebinding protection)
```js
const { buildAllowedHostnames, isAllowedHostHeader, isAllowedOrigin } = require('./lib/loopback-guard');
const ALLOWED = buildAllowedHostnames(HOST);
// In request handler (before routing):
if (!isAllowedHostHeader(req.headers['host'], ALLOWED)) { res.writeHead(421); ... }
```
Blocks requests where the Host header doesn't match the known loopback hostnames, consistent with the pattern already used in `plan-canvas/server.js`.

### 3. Origin guard (cross-origin protection)
```js
if (req.headers['origin'] && !isAllowedOrigin(req.headers['origin'], ALLOWED)) { res.writeHead(403); ... }
```
Rejects cross-origin requests from disallowed origins.

## Gap 2
`agent-data-home.js` path-containment hardening (the second gap in #2506) will follow in a separate commit on this branch.

## Testing
- `node scripts/dashboard-web.js` still starts and serves on `http://localhost:3456`
- Requests with spoofed Host headers (e.g. `Host: attacker.com`) now receive 421
- `ECC_DASHBOARD_HOST=0.0.0.0 node scripts/dashboard-web.js` still works for users who need it

### DIFF
--- a/scripts/dashboard-web.js
+++ b/scripts/dashboard-web.js
@@ -12,6 +12,7 @@
 const fs = require('fs');
 const path = require('path');
 const http = require('http');
+const { buildAllowedHostnames, isAllowedHostHeader, isAllowedOrigin } = require('./lib/loopback-guard');
 
 function parsePort(v) {
   const n = parseInt(String(v), 10);
@@ -20,6 +21,8 @@ function parsePort(v) {
 }
 const PORT = parsePort(process.argv[2] || process.env.ECC_DASHBOARD_PORT || '3456');
 const ROOT = path.resolve(__dirname, '..');
+const HOST = process.env.ECC_DASHBOARD_HOST || '127.0.0.1';
+const ALLOWED = buildAllowedHostnames(HOST);
 
 function readFrontmatter(p) {
   try {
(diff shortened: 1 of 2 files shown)


rjjelley:
Local review from Codex automation: I would request changes before merging this PR. `git diff --check origin/main...HEAD` reports trailing whitespace in `scripts/dashboard-web.js`, the Host/Origin guard formatting is cramped into one line, and `scripts/lib/agent-data-home.js` discards the canonical path returned by the containment check. Please also add negative tests for rejected Host/Origin headers and project-config containment before treating #2506 as fully fixed.

daltino (approved):
This PR strengthens security for the dashboard by restricting the bind to loopback-only (`127.0.0.1`) and adding Host/Origin validation mechanisms. It introduces helper functions to handle hostname safety and origin checks, aligning with secure server practices. Changes are concise, follow existing structure, and appropriately enhance safety.

affaan-m (changes requested):
Thanks for tackling both findings. I verified this branch against the current issue and CI, and there are still blocking gaps:

1. `agentDataHome` is bounded only by the user's home directory. A cloned repository under `~/...` can still set a relative path such as `.ecc-data`, redirect session persistence into its own Git tree, and preserve the original exfiltration path. Project-controlled config needs a user-owned trusted root (for example `~/.cursor/ecc`) or an explicit user-controlled allowlist; `$HOME` alone is too broad.
2. The branch's `agent-data-home` tests still expect the now-rejected `/tmp` and project-relative behavior, so the runtime matrix is failing. Please rewrite those assertions around the intended trust boundary and fallback behavior.
3. The dashboard change needs negative Host/Origin regression tests, a loopback default assertion, and `git diff --check` cleanup. The current `0.0.0.0` option also does not define safe LAN Host/auth semantics.
4. Please rebase onto current `main`; this branch predates several overlapping changes.

I recommend loopback-only dashboard behavior unless authenticated LAN exposure is deliberately designed, and allowing repository config only beneath a dedicated user-owned ECC data root. I am preparing a clean current-main implementation so the security issue is not left open.

affaan-m:
Superseding implementation opened in #2585. It carries both #2506 fixes, rejects unauthenticated wildcard/LAN binding, and includes the related install-state/source/symlink boundary hardening found during security review. I’m leaving this PR open until #2585 clears CI and lands, then I’ll close it with the replacement link. Thank you for the original fix and issue follow-through.

affaan-m:
Closing as superseded by #2585, now merged as 382060905e2270c140b24a39dcd7759cc85ec501. The merged fix includes the loopback bind and Host/Origin guard from this PR, plus no-store/generic error handling and the related install-lifecycle containment hardening, with the full 44-check matrix green.

PR #2529, opened 2026-07-18
fix: detect Anthropic API keys (sk-ant-...) in pre-commit secret scan

Oleksandr-Kh4:
## Problem

The secret scan in `scripts/hooks/pre-bash-commit-quality.js` never matches real Anthropic API keys. Their format is `sk-ant-api03-...` and contains hyphens, which terminate the `sk-[a-zA-Z0-9]{20,}` character class long before the 20-char threshold, so the OpenAI pattern silently passes them through. Keys of the primary Claude Code audience slip into commits unnoticed.

## Fix

- Add a dedicated `sk-ant-[a-zA-Z0-9_-]{20,}` pattern (checked before the generic OpenAI one).
- Extend the staged-secrets test with a realistic Anthropic key fixture and assert the new warning.

## Validation

`node tests/hooks/pre-bash-commit-quality.test.js` -> Passed: 9, Failed: 0. Red-green confirmed: the fixture key is missed by the old pattern and caught by the new one.

Found while running ECC project-scope in production against Claude-API-heavy repos.

### DIFF
--- a/scripts/hooks/pre-bash-commit-quality.js
+++ b/scripts/hooks/pre-bash-commit-quality.js
@@ -108,6 +108,7 @@ function findFileIssues(filePath) {
       
       // Check for hardcoded secrets (basic patterns)
       const secretPatterns = [
+        { pattern: /sk-ant-[a-zA-Z0-9_-]{20,}/, name: 'Anthropic API key' },
         { pattern: /sk-[a-zA-Z0-9]{20,}/, name: 'OpenAI API key' },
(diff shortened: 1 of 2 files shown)


daltino (approved):
This PR adds a detection rule for Anthropic API keys (`sk-ant-...`) in the pre-commit secret scan, ensuring better security coverage. The implementation in `pre-bash-commit-quality.js` is consistent with existing patterns, and a corresponding test update confirms functionality. Looks good! 🎉

PR #2533, opened 2026-07-19
docs: fix stale command references in COMMANDS-QUICK-REF.md

latreon:
## Summary
- `COMMANDS-QUICK-REF.md` claimed "59 commands" and documented 11 retired commands (`/tdd`, `/eval`, `/verify`, `/e2e`, `/docs`, `/claw`, `/context-budget`, `/devfleet`, `/orchestrate`, `/prompt-optimize`, `/rules-distill`) that moved to `legacy-command-shims/` in favor of skills, while omitting ~36 real commands added since (`orch-*`, `prp-*`, `react-*`, `jira`, `hookify*`, etc).
- Rebuilt it against the actual 84 files in `commands/`, added a "Retired Commands" table mapping each dead slash command to its replacement skill, and fixed the dead references in the Quick Decision Guide.

Split out of #2532 into its own PR (translated-docs fix moved to a separate PR).

## Test plan
- [x] `node scripts/ci/validate-commands.js` — 84/84 command files validate clean (unaffected by this doc-only change, ran to confirm no drift)
- [x] `node tests/run-all.js` — 2688/2690 pass (2 pre-existing failures unrelated to this change — from untracked local scaffold dirs under `.agents/skills/source-command-*`)

PR #2532, opened 2026-07-19
docs: fix stale command references in quick-ref and translated docs

latreon:
## Summary
- `COMMANDS-QUICK-REF.md` claimed "59 commands" and documented 11 retired commands (`/tdd`, `/eval`, `/verify`, `/e2e`, `/docs`, `/claw`, `/context-budget`, `/devfleet`, `/orchestrate`, `/prompt-optimize`, `/rules-distill`) that moved to `legacy-command-shims/` in favor of skills, while omitting ~36 real commands added since (`orch-*`, `prp-*`, `react-*`, `jira`, `hookify*`, etc). Rebuilt it against the actual 84 files in `commands/`, added a "Retired Commands" table mapping each dead slash command to its replacement skill, and fixed the dead references in the Quick Decision Guide.
- The same 11 retired commands were still documented as translated per-command pages under `docs/{ja-JP,zh-CN,zh-TW,ko-KR,pt-BR,es,tr}/commands/`. Removed the 47 stale files (count differs per locale — not all locales had all 11 translated). `docs/ja-JP/commands/README.md` also listed 4 of them directly; updated those bullets to point at the replacement skills.

## Test plan
- [x] `node scripts/ci/validate-commands.js` — 84/84 command files validate clean
- [x] `node tests/lib/locale-install.test.js` — 7/7 pass (translated-docs install paths unaffected)
- [x] `node tests/ci/catalog.test.js` — 4/4 pass
- [x] `node tests/run-all.js` — 2688/2690 pass (the 2 failures are pre-existing, unrelated to this change — caused by untracked local scaffold dirs under `.agents/skills/source-command-*`, not touched by this PR)

PR #2534, opened 2026-07-19
docs: remove stale translated docs for retired commands

latreon:
## Summary
- 11 commands (`/tdd`, `/eval`, `/verify`, `/e2e`, `/docs`, `/claw`, `/context-budget`, `/devfleet`, `/orchestrate`, `/prompt-optimize`, `/rules-distill`) were retired to `legacy-command-shims/` in favor of skills, but their translated per-command doc pages were still committed under `docs/{ja-JP,zh-CN,zh-TW,ko-KR,pt-BR,es,tr}/commands/`.
- Removed the 47 stale files (count differs per locale — not all locales had all 11 translated).
- `docs/ja-JP/commands/README.md` also listed 4 of them directly; updated those bullets to point at the replacement skills instead.

Split out of #2532 into its own PR (COMMANDS-QUICK-REF.md fix moved to a separate PR: #2533).

## Test plan
- [x] `node tests/lib/locale-install.test.js` — 7/7 pass (translated-docs install paths unaffected by removing these stale files)
- [x] `node tests/ci/catalog.test.js` — 4/4 pass
- [x] `node tests/run-all.js` — 2688/2690 pass (2 pre-existing failures unrelated to this change — from untracked local scaffold dirs under `.agents/skills/source-command-*`)


daltino (commented):
**DECISION: APPROVE**

This PR removes stale, translated documentation for retired commands, which aligns with maintaining an up-to-date and relevant documentation set. The changes appear straightforward, consisting only of file deletions, and comply with the contribution guidelines under "Cross-Harness and Translations" about maintaining relevant content. No further action is needed.

haelyra:
Thank you for spotting that the translated command documentation drifted after the skills-first migration. That diagnosis is useful. The current deletion-only shape is not the right landing path, though: ECC intentionally retains these slash commands under `legacy-command-shims/commands` for compatibility, so removing 47 localized pages would make those supported migration entry points less accessible, and the PR leaves locale indexes and directory maps pointing at removed files. We are going to close this version rather than revive a large stale deletion across every locale. A focused follow-up should convert each affected page into a short compatibility redirect to its canonical skill, update every locale index and tree reference, and add link/inventory validation so the shims remain discoverable without duplicating obsolete workflow prose. We appreciate the careful inventory and would be glad to see that smaller migration-focused approach.

PR #2535, opened 2026-07-19
feat(skills): complete Codex skill surface for source-command scaffolds

latreon:
## Summary
`tests/run-all.js` was failing on 3 untracked `.agents/skills/source-command-*` scaffolds (`add-language-rules`, `database-migration`, `feature-development`) left behind by a botched generator run:

- `SKILL.md`'s `name:` field was quoted (`tests/ci/codex-skill-surface.test.js` requires it unquoted and matching the folder name exactly).
- The content baked in the wrong project name — `"...in everything-Codex"` instead of `everything-claude-code` (the source `.claude/commands/*.md` files these were mirrored from already had it right).
- The required `agents/openai.yaml` companion file was missing entirely.

Fixed all three: unquoted the frontmatter, corrected the project name in both the description and body text, and added `agents/openai.yaml` with `display_name`/`short_description`/`default_prompt` so each scaffold satisfies the Codex skill surface contract.

While verifying, this surfaced a latent bug in `tests/scripts/install-apply.test.js`: it spawns the full-profile `--json` dry-run via `execFileSync` with no `maxBuffer` override, so it inherits Node's 1MB default. That dry-run lists every managed file in the repo, and adding these 6 files pushed its output past the limit (`spawnSync ENOBUFS`) — any future skill/agent addition could trip the same wall. Fixed with an explicit 16MB `maxBuffer`.

Diff is intentionally minimal — 7 files, 153 insertions, 0 deletions.

## Test plan
- [x] `node tests/ci/codex-skill-surface.test.js` — 4/4 pass (was failing 2/4 before this change)
- [x] `node tests/scripts/install-apply.test.js` — 30/30 pass (was failing 1/30 before the maxBuffer fix, reproduced consistently across repeated runs)
- [x] `node scripts/ci/validate-commands.js` — 94/94 clean
- [x] `node tests/run-all.js` — 3110/3110 pass

### DIFF
--- a/.agents/skills/source-command-add-language-rules/SKILL.md
+++ b/.agents/skills/source-command-add-language-rules/SKILL.md
@@ -1,0 +1,44 @@
+---
+name: source-command-add-language-rules
+description: Workflow command scaffold for add-language-rules in everything-claude-code.
+---
+
+# source-command-add-language-rules
+
+Use this skill when the user asks to run the migrated source command `add-language-rules`.
+
+## Command Template
+
+# /add-language-rules
+
+Use this workflow when working on **add-language-rules** in `everything-claude-code`.
+
+## Goal
+
+Adds a new programming language to the rules system, including coding style, hooks, patterns, security, and testing guidelines.
+
+## Common Files
+
+- `rules/*/coding-style.md`
+- `rules/*/hooks.md`
+- `rules/*/patterns.md`
+- `rules/*/security.md`
+- `rules/*/testing.md`
+
+## Suggested Sequence
+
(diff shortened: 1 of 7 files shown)


haelyra:
Thank you for tracing the validator failure and spotting the possible output-buffer pressure. On clean current main, these three source-command directories do not exist and CI passes; they were local generator residue, so committing them would turn that residue into shipped legacy scaffolds outside ECC’s skills-first surface. We are closing this version. A focused maxBuffer hardening PR would be welcome if the failure reproduces on clean current main.

PR #2536, opened 2026-07-19
fix(scripts): detect modern bun.lock, ignore stray root lockfiles

latreon:
## Summary
Bun switched its default lockfile from the binary `bun.lockb` to the text-based `bun.lock`, but `detectFromLockFile()` in `scripts/lib/package-manager.js` only ever checked for `bun.lockb` — a project using modern Bun would never be detected as using Bun at all.

- Made `bun.lock` the primary recognized lockfile, kept `bun.lockb` as a recognized legacy alias via a new `lockFileAliases` field, and updated `detectFromLockFile()` to check both.
- Added `/bun.lock` and `/bun.lockb` to `.gitignore` at the repo root — yarn is this repo's canonical package manager (`package.json` `"packageManager"`), so a lockfile from someone running `bun install` locally against this repo shouldn't show up in `git status`.
- Added a regression test for the modern `bun.lock` path alongside the existing `bun.lockb` one.

## Test plan
- [x] `node tests/lib/package-manager.test.js` — 116/116 pass (was 115, added 1 new test)
- [x] `node tests/run-all.js` — 3111/3111 pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,6 +99,11 @@ ecc2/target/
 # Generated lock files in tool subdirectories
 .opencode/package-lock.json
 .opencode/node_modules/
+
+# yarn is the canonical package manager for this repo (see package.json
+# "packageManager"); ignore stray lockfiles from running another manager locally
+/bun.lock
+/bun.lockb
 assets/images/security/badrudi-exploit.mp4
 
 .aider*
--- a/scripts/lib/package-manager.js
+++ b/scripts/lib/package-manager.js
@@ -43,7 +43,10 @@ const PACKAGE_MANAGERS = {
   },
   bun: {
     name: 'bun',
-    lockFile: 'bun.lockb',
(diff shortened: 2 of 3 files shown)


daltino (commented):
**DECISION: APPROVE**

This PR makes focused updates to improve lockfile handling for Bun, aligning with its modern lockfile format while retaining compatibility for legacy files. The addition to `.gitignore` ensures only the canonical package manager's files are tracked. The updates to the detection logic and tests are thorough and well-structured, adhering to the repo's conventions. Well done!

haelyra:
Thank you for this! Supporting both the current bun.lock format and the legacy bun.lockb file fixes package-manager detection without changing the existing priority rules. The recorded checks are green, and the combined current-main test run passes in full, so I’m merging it now. We appreciate the careful regression coverage.

PR #2549, opened 2026-07-23
fix: instinct promote leaves orphaned project-scoped copy

wouterschoonees:
## Summary
- `instinct promote <id>` (and the auto cross-project variant) writes the promoted instinct into the global personal dir but never removes it from the project-scoped source file it came from.
- `load_all_instincts()` dedupes by ID with project-scope taking precedence over global on collision, so the leftover project copy silently shadows the new global entry - `status` under-reports the global count (and shows the instinct as still project-scoped) until the stale file is deleted by hand.
- Reproduced by importing a multi-instinct file, promoting one ID by hand, and diffing `status` output before/after manually removing the stale source file.

## Fix
- Adds `_remove_instinct_from_source()`: re-parses the instinct's source file, strips just the promoted block, rewrites the remaining blocks in the same frontmatter format, or deletes the file outright if it was the only instinct in it.
- Wired into `_promote_specific` (single-ID promote).
- Wired into `_promote_auto` as well - it iterates every project entry that contributed to the cross-project candidate, not just the highest-confidence one used for the merged output content, since each project's copy independently shadows the same global ID.

## Test plan
- [x] `py_compile` on the patched file
- [x] Isolated smoke test (`CLV2_HOMUNCULUS_DIR` pointed at a temp dir): imported a 2-instinct file into a scratch git repo, promoted one ID, confirmed the promoted instinct disappears from project scope and appears under global in `status`, while the un-promoted sibling in the same source file survives untouched
- [ ] Maintainers: no existing automated test coverage found for `_promote_specific`/`_promote_auto`; happy to add one if there's a preferred harness/location for `instinct-cli.py` tests

### DIFF
--- a/skills/continuous-learning-v2/scripts/instinct-cli.py
+++ b/skills/continuous-learning-v2/scripts/instinct-cli.py
@@ -1327,8 +1327,14 @@ def _remove_instinct_from_source(source_file_str: str, instinct_id: str) -> None
     if not source_file.exists():
         return
 
-    remaining = [i for i in parse_instinct_file(source_file.read_text(encoding="utf-8"))
-                 if i.get('id') != instinct_id]
+    parsed = parse_instinct_file(source_file.read_text(encoding="utf-8"))
+    if not any(i.get('id') == instinct_id for i in parsed):
+        # Target not found in a successful parse - either already removed, or the
+        # file is malformed/foreign content. Either way, don't touch it: a parse
+        # failure must never be treated as proof the file is safe to delete.
+        return
+
+    remaining = [i for i in parsed if i.get('id') != instinct_id]
 
     if not remaining:
         source_file.unlink()
--- a/skills/continuous-learning-v2/scripts/instinct-cli.py
+++ b/skills/continuous-learning-v2/scripts/instinct-cli.py
@@ -1314,6 +1314,39 @@ def _show_promotion_candidates(project: dict) -> None:
         print(f"  Run `instinct-cli.py promote` to promote these to global scope.\n")
 
 
+def _remove_instinct_from_source(source_file_str: str, instinct_id: str) -> None:
+    """Strip one instinct block from its project-scoped source file after promotion.
+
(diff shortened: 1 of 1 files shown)


daltino (commented):
This PR addresses a significant issue where project-scoped copies of instincts remain after promotion, causing problems with global instinct recognition. The added function `_remove_instinct_from_source` appears to handle the cleanup effectively but could benefit from clarifications in its documentation and potentially additional safeguards. Consider adding a description of what happens if the instinct ID isn't found and test coverage to validate edge cases.

affaan-m:
Thanks for surfacing this. The shadowing bug is real, but this patch still has data-integrity risks around rewriting mixed/malformed source files. I opened #2587 as an owner-side replacement that preserves non-target blocks byte-for-byte, uses an atomic same-directory replace for mixed files, and adds explicit/auto promotion regression coverage. Recommended disposition: close this PR as superseded by #2587 once that replacement lands.

affaan-m (changes requested):
The shadowing bug is real, but this implementation needs data-safety coverage before merge. `_source_file` is treated as a trusted deletion/write target even though project files and symlinks can change after discovery; constrain it to the current project instinct roots, reject final/parent symlink escapes, and revalidate immediately before the filesystem effect. Re-serializing parsed dictionaries also drops comments/preamble and can change quoting/formatting for surviving blocks; remove the target block without lossy reconstruction. Finally, if cleanup fails after the global file is written, a retry currently stops at “already exists globally” and cannot repair the stale shadow, so make the operation recoverable/transactional and report cleanup failures. Please add automated tests for single- and multi-instinct sources, malformed/unreadable files, retry after partial failure, and path/symlink containment.

haelyra:
Thank you so much for catching this! The issue was real, and the safer replacement has now landed in #2587 with atomic cleanup and regression coverage. We’re closing this one as superseded, and we really appreciate the contribution.

wouterschoonees:
Thx for the feedback

*Regards*
*Wouter Schoonees*
0829980222

PR #2567, opened 2026-07-24
feat(skills): add contract-first collaboration workflow

Seekers2001:
## Summary

- add a `contract-first` skill for consumer/provider API and event collaboration
- center the workflow on one machine-checkable boundary artifact, consumer-first design, provider verification, and an explicit contract-change protocol
- wire the skill into package exports, selective installation, GitAgent configuration, and the repository catalogs

## Why

ECC already provides strong guidance for API design quality and regression testing, but it does not currently cover the coordination gap before integration: two independently reasonable implementations can still drift on field names, status codes, event shapes, or error semantics.

This skill stays narrow. It does not introduce a new contract framework or generator. It encourages teams to reuse their existing schema format and pinned tooling while establishing a clear source of truth and evidence-based handoff between consumers and providers.

## Validation

- skill frontmatter and structure validation
- `node scripts/ci/validate-skills.js --strict`
- Unicode safety check
- personal-path safety check
- JSON parse validation for changed manifests
- 40 catalog and package-wiring assertions
- `git diff --check`
- independent forward test using the skill on a sample integration scenario

## Scope note

The full repository test suite was not run locally because this contribution was prepared from a partial checkout while GitHub downloads were unstable. The repository's strict skill validator and targeted catalog/package checks pass; CI on the complete checkout remains the final integration check.


affaan-m (approved):
Verified the skill stays narrowly focused on a single machine-checkable boundary artifact, includes consumer/provider evidence gates and an explicit change protocol, and now treats embedded contract content as untrusted data with allowlisted references, pinned least-privilege generators, and generated-diff review. Packaging/catalog counts are wired and all 40 checks pass.

PR #2571, opened 2026-07-25
Restore guide covers and switch the partner row to brand marks

affaan-m:
The three guide covers had been replaced with generated placeholder cards, so the Shorthand and Longform guides lost their actual artwork. This restores the real headers and normalizes all three to 1200x500 so they sit back to back at the same size.

Partner row now uses each brand's mark rather than its wordmark, at a uniform height of 60:
- Atlas Cloud: the "A" mark, extracted from their own wordmark SVG (light + dark)
- Moonshot AI (Kimi): the "K" logomark, both background variants as supplied by Moonshot
- Itô: the Itô glyph only, not the "Itô Markets" lockup

Kimi links carry the partner tracking parameter Moonshot asked for (`platform.kimi.ai?aff=ecc`). The Coding Plan link is held back until new purchases reopen.

Framing restored to what the partners actually asked for: Itô partners with ECC on compute; Moonshot AI (Kimi) are open source friends, not business sponsors.

https://claude.ai/code/session_0149VwNuynam6rvEfcMmiHHa


daltino (commented):
The PR focuses on design updates by restoring guide covers and switching sponsor logos to brand marks. The changes in the README look fine, but it would be helpful to verify that the updated image paths render correctly. Also, confirm that the added SVG for the sponsor logo aligns with the existing branding. Let the contributor know if testing was carried out to ensure these visuals display as intended.

PR #2575, opened 2026-07-25
feat(skills): add Solana + multi-network x402 via upstream x402 packages

notorious-d-e-v:
## Summary

Adds a third integration path to the `agent-payment-x402` skill — **Solana + multi-network EVM via the upstream x402 packages** — alongside the existing Base (`agentwallet-sdk`) and X Layer (OKX) options. The skill previously had no Solana coverage and no pointer to the canonical protocol implementation.

The new Option C centers on the actively maintained [`x402-foundation/x402`](https://github.com/x402-foundation/x402) monorepo and the packages it publishes, keeping the guidance facilitator-neutral:

- **Buyer side** — wrap the agent's HTTP client with `@x402/fetch` / `@x402/axios` and register both `eip155:*` and `solana:*` schemes, mirroring the upstream [client examples](https://github.com/x402-foundation/x402/tree/main/examples/typescript/clients). One client pays whichever chain the 402 offers.
- **Seller side** — upstream middleware (`@x402/express|hono|next|fastify` pinned at `2.19.0`, Python `x402`, Go `github.com/x402-foundation/x402/go/v2`); one route can advertise Base and Solana simultaneously, per the upstream [server examples](https://github.com/x402-foundation/x402/tree/main/examples/typescript/servers).
- **Facilitators** — the packages' default `x402.org` facilitator for testnets (Base Sepolia + Solana devnet; upstream docs say not for mainnet), and the [upstream facilitator list](https://docs.x402.org/dev-tools/facilitators) for production, with PayAI suggested (multi-network incl. Solana mainnet, no API keys) and CDP noted (Coinbase-hosted, KYT/OFAC).
- **Discovery** — the bazaar-extension `/discovery/resources` endpoint on both CDP and PayAI, plus [pay.sh](https://pay.sh) for Solana-payable services.
- **Safety rails** — explicit user confirmation before the first paid request (mirroring Option B), a defined fail-closed budget gate invoked immediately before every paid call, USDC mint validation with the concrete mainnet/devnet addresses from upstream constants, and no unpinned `npx` commands.

### Translations
- `docs/ja-JP` — mirrored.
- `docs/zh-CN` — resynced to full parity (it predated the X Layer / Options structure entirely), then the Solana/upstream content mirrored in. Flagging in case you'd prefer a minimal zh-CN scope.

## Type
- [x] Skill

## Testing
Ran the repo CI gauntlet locally: `validate-skills` (279 dirs), `check-unicode-safety` (incl. CJK docs), `catalog:check`, `validate-no-personal-paths`, `security:ioc-scan`, and `markdownlint` on all three files — all pass. `tests/run-all.js` green except one pre-existing install-config test needing `ajv` (local deps not installed); nothing skill-related.

All technical claims verified against `x402-foundation/x402` `main` as of 2026-07-24: package names/versions on npm (2.19.0), the x402.org facilitator's testnet-only scope, the facilitator list, USDC mint constants (`typescript/packages/mechanisms/svm/src/constants.ts`), the Go module path, and both discovery URLs.

## Checklist
- [x] Follows format guidelines (mirrors the existing Option A/B structure)
- [x] Tested with the repo's CI validators
- [x] No sensitive info (keys sourced from env; no personal paths)
- [x] Clear descriptions; inline frontmatter `description` (no block scalar)

## Disclosure
I contribute to PayAI (one of the facilitators referenced). The guidance points at the upstream x402 packages and docs as the canonical path, lists x402.org for testnets, and names both PayAI and CDP for production/discovery — happy to adjust the balance further if you'd like.


codeCraft-Ritik (commented):
Nice addition. Using the upstream x402 packages as the canonical third integration path makes the skill more broadly useful, especially with Solana and multi-network EVM support.

notorious-d-e-v:
Verified this end to end on Solana devnet rather than only against the docs.

Installed the published packages (`@x402/core|evm|svm|express|fetch@2.19.0`), ran the seller middleware and the buyer client from Option C, and settled a real payment:

```
HTTP 200
PAYMENT-RESPONSE: { "success": true, "network": "solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1",
  "transaction": "3dF8CbLSeytRBCHGGsmMwuriuKppx81RKDugo4DnwbaMvYmU27426S7egg9ku3PhSJANk4dr6HS72EBDRdrq2FnM" }
```

On-chain (devnet, slot 478814353, `err: null`): `transferChecked` of 0.001 USDC on mint `4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU`, buyer −0.001 / seller +0.001. The 10,001-lamport fee was paid by `CKPKJWNdJEqa81x7CkZ14BVPiY6y16Sxs7owznqtWYp5` — the address the 402 advertised as `extra.feePayer` — while **both wallets held 0 SOL**. That is the "fund USDC, not SOL" claim demonstrated rather than asserted.

The first attempt failed with `transaction_simulation_failed`, which turned out to be the seller having no USDC associated token account. The upstream client derives the destination ATA (`findAssociatedTokenPda`) and builds only a `transferChecked` — it never creates the account — and I could not find this called out in the upstream docs. The second commit documents it with a pre-flight check and three ways to create the account, since it will most often hit freshly provisioned or custodial payout wallets.

Also confirmed live, which is what the facilitator guidance in the skill is based on: `x402.org/facilitator/supported` advertises Solana **devnet** and no Solana mainnet, while `facilitator.payai.network/supported` carries Solana mainnet across 32 networks.

Happy to paste the full transcript (decoded 402 with both network legs, the budget-gate output, both `/supported` probes) if useful.

daltino (approved):
This PR adds support for Solana and multi-network EVM to the existing x402 payment execution skill, enhancing its versatility for multi-chain scenarios. The updates are consistent across all language documentation files (EN, JA, and ZH-CN) and align with the repo's contribution guidelines. The added descriptions are clear and provide a valuable extension to support upstream packages.

codeCraft-Ritik (commented):
Nice addition. Pointing users toward the upstream x402 implementation while keeping the existing Base and X Layer options intact makes the skill more complete without disrupting current workflows.

haelyra:
Thank you for the thoughtful x402 expansion, and sorry for the review delay. Supporting the upstream protocol across Solana and EVM networks, including token-account discovery and real devnet coverage, is a strong direction for ECC.

We need one focused security pass before this can land:

- **Enforce policy against the actual payment challenge.** The current `maxCost` example is not connected to the server-provided amount, asset, or network before signing, so a response above the configured budget can still be accepted. Success is a client policy or before-payment hook that validates all three fields and refuses the signature when any boundary is exceeded.
- **Refresh and test the supported upstream contract.** The branch pins an older x402 release while the upstream API has moved. Please update to the current supported package contract and add adversarial tests for over-budget amounts, the wrong mint or asset, and the wrong network.
- **Keep facilitator guidance vendor-neutral.** PayAI can remain a useful example, but contributor-affiliated infrastructure should not appear as the production default. Present it alongside neutral self-hosted or upstream-supported options and disclose the relationship.

If you do not have bandwidth, we will keep the multi-network x402 idea in the maintainer queue and can carry the contained policy patch forward after the next release. We appreciate the work and the concrete network coverage.

notorious-d-e-v:
Thanks — all three are addressed in b78ea72.

**1. Policy enforced against the actual challenge.** You were right: the old example compared the budget to a number the caller passed in, so a server quoting any amount would still have been paid. It now registers a `PaymentPolicy` on the client, which filters the server-provided requirements before a signature exists and checks all three fields:

```ts
client.registerPolicy((_x402Version, requirements) =>
  requirements.filter(r => {
    if (!ALLOWED_NETWORKS.has(r.network)) return false;
    if (!ALLOWED_ASSETS.has(normalizeAsset(r.asset))) return false;
    try {
      const amount = BigInt(r.amount);
      return amount >= 0n && amount <= MAX_AMOUNT;
    } catch { return false; }
  }),
);
```

When everything is filtered out, `createPaymentPayload` throws rather than falling back.

**Writing the adversarial tests found a second bug, in the guard itself.** EVM addresses are case-insensitive, so a server quoting lowercase USDC — the same token — was dropped by a `Set` of checksummed strings. Assets are now normalised before comparison, lowercasing only `0x` addresses — Solana is base58 and case-sensitive, so folding those would let a different account through. 12 cases now pass:

```
refused: over-budget $5.00 · wrong mint · wrong network · case-flipped Solana mint
         non-numeric amount · negative amount · one over cap
kept:    checksummed / lowercase / uppercase EVM USDC · Solana USDC · exactly at cap
```

**2. Refreshed to the current contract.** Pins move 2.19.0 → 2.22.0 (2.20, 2.21 and 2.22 shipped after this branch opened). The documented server and client APIs are unchanged across that range — I re-ran the seller middleware on 2.22.0 and it still returns a multi-network 402 advertising Base Sepolia and Solana devnet.

**3. Vendor neutrality.** Fair, and I have restructured rather than just re-worded. Self-facilitation and running your own facilitator are now listed first, ahead of the package default and hosted options; the hosted paragraph points at the upstream maintained list instead of naming a default, and notes it is not exhaustive. An explicit affiliation disclosure now sits in the skill itself, not only in the PR. PayAI appears as one name in a list alongside CDP, Corbits, Dexter and Solvador.

All mirrored across en / ja-JP / zh-CN; repo validators and markdownlint pass.

One thing I did not do: adversarial tests are not committed as repo tests, since `skills/` is markdown and `tests/` covers `scripts/`. The cases are documented in the skill so a reader can reproduce them. If you would like them as a real test file, say where it belongs and I will add it.

notorious-d-e-v:
Also merged `main` into the branch — it now conflicts no longer. The only overlap was the frontmatter `description`, where main added the "Use when an agent must pay for something itself…" sentence while this branch extended the same line with Solana and multi-network EVM coverage. Kept both.

notorious-d-e-v:
Fixed both P1s in f1c35112.

**Authorization boundary** — correct, and it was a regression: the earlier ad-hoc gate had a host allowlist and session cap, which I dropped when moving to `PaymentPolicy`. The policy now also allowlists `payTo`, and every paid call goes through a boundary that confirms once and holds a session ledger (reserving the per-call cap, since settlement responses carry no amount). The payment-enabled client stays private behind it. This also brings the example back in line with the skill's own Spending Controls section.

**ATA check** — also correct. Scanning the owner's token accounts proves only that *some* account exists for the mint; the client targets the canonical ATA. Now derives it with `findAssociatedTokenPda` and checks that exact address. I also corrected the provisioning advice: a bare transfer does not create the account, only one including the create instruction does.

Not addressed, as both predate this PR and sit in Option A / the Best Practices list rather than the Solana path: the `npm install -g` vs `npx` spec point, and the `security-review` → `security-reviewer` naming. They show up in the zh-CN diff only because that file was resynced wholesale. Happy to take either as a follow-up if you want them changed.

automatonagent205-afk:
## 🏆 Solution: Add Solana + multi-network x402 via upstream x402 packages

**Payout:** `24p4kMYz26tZE487ZhPadPXXpip2GWtE2PSjsTEcBWBM`

### Implementation

Install the upstream x402 packages and configure Solana network support:

```bash
npm install @x402/x402-fetch @x402/x402-sdk
```

```typescript
// src/x402/solana-adapter.ts
import { createX402Client } from '@x402/x402-fetch';

export interface SolanaPaymentConfig {
  network: 'mainnet' | 'testnet' | 'devnet';
  rpcUrl: string;
  wallet: string; // Base58 encoded keypair or pubkey
}

export const createSolanaX402Client = (config: SolanaPaymentConfig) => {
  return createX402Client({
    network: config.network,
    chainId: config.network === 'mainnet' ? 101 : config.network === 'devnet' ? 103 : 102,
    rpcUrl: config.rpcUrl,
    signer: config.wallet,
    supportedTokens: ['USDC', 'SOL'],
    getTokenAddress: (token: string) => {
      const tokens: Record<string, string> = {
        USDC: 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v', // Solana mainnet USDC
        SOL: 'So11111111111111111111111111111111111111112',     // Wrapped SOL
      };
      return tokens[token] ?? null;
    },
  });
};

// Usage example
const client = createSolanaX402Client({
  network: 'mainnet',
  rpcUrl: 'https://api.mainnet-beta.solana.com',
  wallet: process.env.SOLANA_WALLET!,
});

const response = await client.x402Fetch('https://api.jup.ag/price/sol', {
  headers: { 'x-payment-amount': '0.01', 'x-payment-address': '24p4kMYz26tZE487ZhPadPXXpip2GWtE2PSjsTEcBWBM' }
});
```

```typescript
// src/x402/multi-network.ts — Multi-network x402 support
export type Network = 'solana' | 'ethereum' | 'polygon' | 'base';

export interface X402Config {
  network: Network;
  paymentAddress: string;
  defaultAmount?: number;
}

export const MULTI_NETWORK_CONFIGS: Record<Network, X402Config> = {
  solana: {
    network: 'solana',
    paymentAddress: '24p4kMYz26tZE487ZhPadPXXpip2GWtE2PSjsTEcBWBM',
    defaultAmount: 0.01,
  },
  ethereum: {
    network: 'ethereum',
    paymentAddress: '0x...',
    defaultAmount: 0.01,
  },
  polygon: {
    network: 'polygon',
    paymentAddress: '0x...',
    defaultAmount: 0.01,
  },
  base: {
    network: 'base',
    paymentAddress: '0x...',
    defaultAmount: 0.01,
  },
};

export function getX402Config(network: Network): X402Config {
  return MULTI_NETWORK_CONFIGS[network];
}
```

*Bounty Release / Payout Solana Address: 24p4kMYz26tZE487ZhPadPXXpip2GWtE2PSjsTEcBWBM*

notorious-d-e-v:
Hey just checking in here @haelyra - is there anything I can do to help move this along?

PR #2577, opened 2026-07-25
fix(resolve-ecc-root): require ECC skills, not just scripts, before accepting a root (#2544)

### LINKED ISSUE #2544: resolve-ecc-root probes for a script to decide where skills live; partial install (scripts copied, skills not) defeats the #2037 fix
## Summary

The `resolveEccRoot()` fix for #2037 decides "this is a standard install" by probing for **`scripts/lib/utils.js`**. On my machine `~/.claude/` contains ECC's `scripts/` but **not** its `skills/`, so the probe succeeds, the resolver returns `~/.claude`, and every command that needs a skill resolves to a path that does not exist.

The probe checks for a *script* to conclude where *skills* are. Those two facts can disagree, and here they do.

## Environment

- ECC 2.0.0 (`~/.claude/plugins/marketplaces/ecc`)
- Windows 11, Git Bash
- `CLAUDE_PLUGIN_ROOT` unset (running a slash command, not a hook)

## Observed state

```
~/.claude/scripts/          160 files (ECC's — auto-update.js, harness-audit.js, hooks/, lib/, ...)
~/.claude/skills/           4 entries (mine; none from ECC)
plugins/marketplaces/ecc/skills/   278 entries
```

`~/.claude/skills/continuous-learning-v2/` does not exist. The plugin's copy does.

I did not construct this state by hand — I ran `/reload-plugins` and `~/.claude/scripts/lib/utils.js` was written one minute after the plugin's own copy. I have not traced which code path wrote it, so I'm reporting the state rather than asserting a cause.

## Reproduction

`/ecc:instinct-status` (any ECC command that resolves a skill path):

```
ECC_ROOT=C:\Users\OEM\.claude
python3: can't open file 'C:\Users\OEM\.claude\skills\continuous-learning-v2\scripts\instinct-cli.py'
```

The command's own docstring says it resolves the root this way specifically "to avoid the divergence that happens when `CLAUDE_PLUGIN_ROOT` is unset while a legacy `~/.claude/skills/continuous-learning-v2/` directory still exists (#2037)" — but the divergence it guards against is the mirror image of the one that actually occurs. #2037 assumed a stale *skills* dir; the failing case here is a present *scripts* dir with skills absent.

## Root cause

`scripts/lib/resolve-ecc-root.js:49-54`

```js
const probe = options.probe || path.join('scripts', 'lib', 'utils.js');

// Standard install — files are copied directly into ~/.claude/
if (fs.existsSync(path.join(claudeDir, probe))) {
  return claudeDir;
}
```

`scripts/lib/utils.js` present does not imply `skills/` present. Any install that lands one without the other short-circuits here and never reaches the plugin-root and plugin-cache branches below, which would have resolved correctly.

## Suggested fix

Probe for something the callers actually need. Most consumers of `resolveEccRoot()` build a `skills/...` path, so probing a known skill directory would make the check match the claim:

```js
const probe = options.probe || path.join('skills', 'continuous-learning-v2', 'scripts', 'instinct-cli.py');
```

A stricter variant: require **both** artifacts before accepting a candidate, so a half-populated root is rejected everywhere rather than only at `~/.claude`:

```js
function isCompleteRoot(dir) {
  return fs.existsSync(path.join(dir, 'scripts', 'lib', 'utils.js'))
      && fs.existsSync(path.join(dir, 'skills'));
}
```

That second form also protects the plugin-root and plugin-cache branches, which today accept a candidate on the same single-file evidence.

Worth noting: the final `return claudeDir` fallback returns a path the function never verified. Returning `null` and letting callers fail loudly would have surfaced this immediately instead of producing a confusing "file not found" three layers away from the actual decision.

## Workaround

Renaming the orphaned copy makes the probe miss and the resolver falls through correctly:

```bash
mv ~/.claude/scripts ~/.claude/scripts.bak
```

`resolveEccRoot()` then returns `~/.claude/plugins/marketplaces/ecc`, and `/ecc:instinct-status` runs verbatim. I checked before doing this that `settings.json` has zero references to `~/.claude/scripts`, that the plugin's `hooks/hooks.json` uses the resolver rather than absolute paths, and that all 48 hook scripts exist under the plugin — no hook broke.

Setting `CLAUDE_PLUGIN_ROOT` globally in `settings.json` also works, but seems inadvisable to recommend: it is a per-plugin harness variable, and pinning it to ECC's directory would likely misdirect other installed plugins.

## Separate, smaller note

The command in `commands/instinct-status.md` invokes `python3` directly. On Windows without a separate Python install, `python3` resolves to the Microsoft Store alias stub, which prints an install advertisement to stdout and exits non-zero. Possibly related to #2452 / #2489 (Windows/Git Bash breakage in continuous-learning-v2), though the failure mode is different. Happy to open that separately if it isn't already covered.

gaurav0107:
## What Changed

`resolveEccRoot()` decided a directory was a valid ECC root by probing for a single script (`scripts/lib/utils.js`). Skill-resolving callers then build `skills/...` paths against that root. This PR makes the **default** probe require a *complete* root — both the script tree **and** a sentinel ECC skill (`skills/continuous-learning-v2`) — and applies the same stricter check to all three candidate branches (standard `~/.claude`, exact plugin roots, versioned plugin cache).

An **explicit** caller `probe` is still honored exactly, so script consumers such as `scripts/hooks/session-start-bootstrap.js` (which probes for the hook runner and does not need skills co-located) are unaffected. `INLINE_RESOLVE` delegates to `resolveEccRoot()`, so it inherits the fix with no change.

## Why This Change

When an install lands ECC's `scripts/` into `~/.claude` but not ECC's `skills/`, the old script-only probe succeeds at the standard-install branch, `resolveEccRoot()` returns `~/.claude`, and every skill command resolves to a path that does not exist (e.g. `skills/continuous-learning-v2/scripts/instinct-cli.py`) — failing three layers away from the decision. This is exactly the partial-install state reported in #2544, where the `#2037` root-resolution fix is defeated.

Simply checking that `skills/` *exists* is not enough: as the report shows, a user's own `~/.claude/skills/` can be present with **none** of ECC's skills. The fix therefore probes for a sentinel skill that ships in every ECC root and is precisely what the failing commands need. The final `~/.claude` fallback is intentionally left in place so callers still fail loudly at a missing path rather than the resolver inventing one.

## Testing Done

- `node tests/lib/resolve-ecc-root.test.js` — 28/28 pass, including two new #2544 regression tests (partial `~/.claude` with a non-ECC `skills/` prefers a complete plugin-cache root; scripts-only root with no complete root falls back to `~/.claude`).
- `node tests/lib/command-plugin-root.test.js` — 4/4 pass (fixtures updated to build complete roots).
- `node tests/run-all.js` — 3161/3163 pass. The only 2 failures are pre-existing in `scripts/ito-compute-sponsor.test.js` (README sponsor-content assertions); they reproduce on a clean `main` checkout and are unrelated to this change.
- ESLint clean on the changed files.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

- [x] Scoped to one module (`scripts/lib/resolve-ecc-root.js`) plus its tests — no new dependencies.
- [x] No new personal/absolute paths (`validate-no-personal-paths` passes).
- [x] Explicit-probe callers verified unaffected; behaviour change is limited to the default (skill-consumer) probe.
- [x] Cross-platform: pure `path`/`fs` existence checks, no shell or OS-specific logic.

## Documentation

- JSDoc for the `options.probe` parameter updated to describe the default-probe completeness requirement. No user-facing docs change required.

Fixes #2544

### DIFF
--- a/scripts/lib/resolve-ecc-root.js
+++ b/scripts/lib/resolve-ecc-root.js
@@ -18,6 +18,17 @@ const PLUGIN_ROOT_SEGMENTS = [
   ['marketplaces', LEGACY_PLUGIN_SLUG],
 ];
 
+// Artifacts that identify a COMPLETE ECC root when the caller gives no explicit
+// probe. A real ECC root ships both the script tree AND ECC's skills; a partial
+// install (scripts copied, skills not) must not qualify for skill-resolving
+// callers, which build `skills/...` paths against the resolved root (#2544).
+// Checking "skills/ exists" is not enough — a user's own ~/.claude/skills/ can
+// be present with none of ECC's skills — so we probe for a sentinel skill that
+// ships in every ECC root and is exactly what the failing skill commands need.
+// If that skill is ever renamed, move this sentinel with it.
+const DEFAULT_SCRIPT_PROBE = path.join('scripts', 'lib', 'utils.js');
+const DEFAULT_SKILL_PROBE = path.join('skills', 'continuous-learning-v2');
+
 /**
  * Resolve the ECC source root directory.
  *
@@ -31,8 +42,14 @@ const PLUGIN_ROOT_SEGMENTS = [
  * @param {object} [options]
  * @param {string} [options.homeDir]  Override home directory (for testing)
  * @param {string} [options.envRoot]  Override CLAUDE_PLUGIN_ROOT (for testing)
- * @param {string} [options.probe]    Relative path used to verify a candidate root
- *                                    contains ECC scripts. Default: 'scripts/lib/utils.js'
+ * @param {string} [options.probe]    Relative path used to verify a candidate
+ *                                    root contains what the caller needs. When
+ *                                    given, it is honored exactly (script
+ *                                    consumers pass their own script path). When
+ *                                    omitted, a candidate must contain BOTH the
+ *                                    ECC script tree and a sentinel ECC skill,
+ *                                    so a partial install (scripts without
+ *                                    skills) is rejected for skill consumers.
  * @returns {string} Resolved ECC root path
  */
 function resolveEccRoot(options = {}) {
@@ -46,10 +63,20 @@ function resolveEccRoot(options = {}) {
 
   const homeDir = options.homeDir || os.homedir();
   const claudeDir = path.join(homeDir, '.claude');
-  const probe = options.probe || path.join('scripts', 'lib', 'utils.js');
+
+  // Decide whether a candidate directory is a usable ECC root. An explicit
+  // caller probe is honored exactly (script consumers know the artifact they
+  // need). With the default probe the caller is a skill consumer, so a
+  // candidate must contain both ECC's scripts and a sentinel ECC skill —
+  // otherwise a scripts-only ~/.claude short-circuits and every skill path
+  // resolves to a location that does not exist (#2544).
+  const isRoot = options.probe
+    ? (dir) => fs.existsSync(path.join(dir, options.probe))
+    : (dir) => fs.existsSync(path.join(dir, DEFAULT_SCRIPT_PROBE))
+            && fs.existsSync(path.join(dir, DEFAULT_SKILL_PROBE));
 
   // Standard install — files are copied directly into ~/.claude/
-  if (fs.existsSync(path.join(claudeDir, probe))) {
+  if (isRoot(claudeDir)) {
     return claudeDir;
   }
 
@@ -60,7 +87,7 @@ function resolveEccRoot(options = {}) {
   );
 
   for (const candidate of legacyPluginRoots) {
-    if (fs.existsSync(path.join(candidate, probe))) {
+    if (isRoot(candidate)) {
       return candidate;
     }
   }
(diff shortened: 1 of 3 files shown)


gaurav0107:
Thanks for the review. Both points addressed in 68bc8208:

- **Extracted the sentinel constant** — `command-plugin-root.test.js` now defines `ECC_SKILL_SENTINEL = path.join('skills', 'continuous-learning-v2')` and reuses it at both fixture setup sites instead of duplicating the literal.
- **Extended #2544 regression coverage** — added two tests asserting the stricter both-sentinels predicate rejects a scripts-only root on the **exact-plugin** branch (prefers a complete plugin-cache root) and the **versioned plugin-cache** branch (falls back to `~/.claude`), matching the existing partial-`~/.claude` case.

`node tests/run-all.js` → 3176/3176 green; ESLint clean.

daltino (approved):
This PR addresses a critical regression by requiring both ECC scripts and essential skills before accepting a root. The added checks improve robustness and ensure skill-resolving callers won't misinterpret incomplete installations. Changes include updates to `resolve-ecc-root.js` and related tests to validate the sentinel skill logic. Documentation is clear, and the additions align with contribution guidelines for skill handling. Solid work overall!

affaan-m (approved):
Reviewed for correctness, regression coverage, path-resolution behavior, and overlap with current unified-memory work. The fix preserves explicit probes, rejects partial ECC roots, covers fallback/cache/plugin cases, and all CI lanes are green.

PR #2578, opened 2026-07-25
feat(control-pane): expose skill runs in the snapshot feed

2mrdigital:
@
## Problem

The control pane snapshot publishes work items, sessions, knowledge and connectors, but not skill runs — even though they are already recorded in the state store and `ecc status` prints them.

That leaves any external reader of `/api/snapshot` with no way to surface skill reliability. The only alternative is reading `state.db` off disk, which defeats the point of having the control pane as the read boundary.

## Change

- `GET /api/snapshot` now includes a `skillRuns` block:

```json
{
  "skillRuns": {
    "totalCount": 2,
    "successCount": 1,
    "failureCount": 1,
    "unknownCount": 0,
    "successRate": 0.5,
    "windowSize": 50,
    "items": [ { "id": "...", "skillId": "...", "outcome": "failure", "failureReason": "...", "durationMs": 1200, "createdAt": "..." } ]
  }
}
```

- 50 most recent runs, newest first, mirroring how work items are read.
- Outcomes normalized to `success` / `failure` / `unknown`; the raw value is kept as `rawOutcome`.
- Unknown outcomes are excluded from `successRate`, so a partially instrumented harness does not read as a reliability drop.
- Work items and skill runs now share one state-store open per snapshot instead of opening the database twice.

## Test isolation fix

`serves HTML and snapshot JSON from a temp ECC2 database` isolated `dbPath` but not `stateDbPath`, so it fell through to the real `~/.claude/ecc/state.db`. On any machine with a recorded work item the test failed with `1 !== 0`. It now uses a temp state store. The new empty-state test points at a nonexistent path for the same reason.

## Tests

Two new cases in `tests/scripts/control-pane.test.js`: populated state store (counts, rate, ordering, failure reason) and missing state store (empty summary, `successRate: null`).

```
node tests/scripts/control-pane.test.js
Results: Passed: 18, Failed: 0

node tests/run-all.js
Total Tests: 3117, Passed: 3116, Failed: 1
```

The single failure is `detects newest Claude plugin install from cache marketplace layout`, unrelated to this change — it reads the real `~/.claude/plugins` directory and fails on a machine with plugins installed. It fails the same way on a clean checkout.

`npx eslint` clean on both changed files.

## Not included

No UI rendering in `ui.js` — this only exposes the data. Happy to add the panel in a follow-up if you want it on the page itself.

## Context

Found while wiring an external operator console to the control pane as a read-only client. Work items, sessions and knowledge mapped cleanly; skills were the one screen with no available source.
@

### DIFF
--- a/scripts/lib/control-pane/state.js
+++ b/scripts/lib/control-pane/state.js
@@ -563,14 +563,84 @@ function summarizeWorkItems(items) {
   return summary;
 }
 
-async function readWorkItemsSnapshot(stateDbPath) {
+const SKILL_RUN_LIMIT = 50;
+
+function normalizeSkillRunOutcome(outcome) {
+  const normalized = String(outcome || '')
+    .trim()
+    .toLowerCase();
+  if (['success', 'succeeded', 'pass', 'passed', 'ok'].includes(normalized)) return 'success';
+  if (['failure', 'failed', 'fail', 'error'].includes(normalized)) return 'failure';
+  return 'unknown';
+}
+
+function normalizeSkillRun(row) {
+  const optionalNumber = (value) => (value === null || value === undefined || value === '' ? null : toNumber(value, 0));
+  return {
+    id: String(row.id || ''),
+    skillId: String(row.skill_id || ''),
+    skillVersion: row.skill_version ? String(row.skill_version) : null,
+    sessionId: row.session_id ? String(row.session_id) : null,
+    taskDescription: String(row.task_description || ''),
+    outcome: normalizeSkillRunOutcome(row.outcome),
+    rawOutcome: String(row.outcome || ''),
+    failureReason: row.failure_reason ? String(row.failure_reason) : null,
+    tokensUsed: optionalNumber(row.tokens_used),
+    durationMs: optionalNumber(row.duration_ms),
+    userFeedback: row.user_feedback ? String(row.user_feedback) : null,
+    createdAt: String(row.created_at || '')
+  };
+}
+
+function readSkillRuns(db) {
+  if (!tableExists(db, 'skill_runs')) return [];
+  return execRows(
+    db,
+    `SELECT *
+     FROM skill_runs
+     ORDER BY created_at DESC, id DESC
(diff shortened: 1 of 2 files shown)


daltino (commented):
**DECISION: APPROVE**

This PR introduces a feature that exposes skill runs in the snapshot feed, with a clear normalization mechanism for skill run outcomes. The code includes robust handling for different outcome formats and limits results to prevent overflow (`SKILL_RUN_LIMIT = 50`). Tests have been updated appropriately to isolate state and validate the snapshot changes, aligning well with the repo's contribution guidelines. Looks good overall!

affaan-m (changes requested):
Thanks for exposing the missing skill-run surface. I’m requesting changes before merge because the current shape creates two release-level problems:

1. `normalizeSkillRun()` returns `taskDescription`, `failureReason`, and `userFeedback` verbatim through `/api/snapshot`. Those fields can contain prompts, local paths, stack traces, credentials, or other sensitive context, and the endpoint is explicitly intended for external operator-console clients. Please make aggregate/safe identifiers the default and require an explicit, bounded, redacted detail mode if raw-ish detail is genuinely needed.

2. The new summary disagrees with the existing state-store status contract for the same database. `scripts/lib/state-store/queries.js` recognizes `success/succeeded/passed` and `failure/failed/error` and emits a 0–100 percentage. This PR additionally classifies `pass`, `ok`, and `fail`, then emits a 0–1 fraction. The same rows can therefore produce different counts and rates in `ecc status` and `/api/snapshot`. Please reuse a shared classifier/summary implementation or match the existing contract exactly.

Also please:

- replace the predictable `/tmp/ecc-control-pane-missing-state/state.db` fixture with `mkdtempSync` plus `finally` cleanup;
- add an unknown-outcome case proving it is counted and excluded from the rate;
- bound every returned string and add a regression proving sensitive/detail fields are absent or redacted by default.

I verified that the patch applies to current main and that its focused tests and ESLint pass; these are API/security contract issues rather than a build failure.

PR #2588, opened 2026-07-26
fix: resolve pnpm in Git Bash pre-push hook

benno0o:
## Summary
- resolve pnpm through Corepack first so consuming repositories use their pinned packageManager version
- preserve direct pnpm as a fallback and fail closed when neither runtime exists
- add hermetic cross-platform pre-push tests for required scripts, fallback, missing runtime, and failure short-circuiting

## Test plan
- node tests/scripts/codex-hooks.test.js (24 passed, 0 failed)
- bash -n scripts/codex-git-hooks/pre-push
- git diff --check
- npm audit --audit-level=high --omit=dev (0 vulnerabilities)
- normal local push probe: lint/typecheck/test/build executed in order and push succeeded
- controlled failing local push probe: typecheck exited nonzero, test/build did not run, push was rejected

## Context
Git Bash may have only pnpm.cmd on PATH, which PowerShell resolves through PATHEXT but Bash does not. Corepack is present and honors the consuming repository's packageManager pin.

### DIFF
--- a/scripts/codex-git-hooks/pre-push
+++ b/scripts/codex-git-hooks/pre-push
@@ -62,6 +62,8 @@ has_node_script() {
 
 run_pnpm() {
   if command -v corepack >/dev/null 2>&1; then
+    # Corepack may download the pinned pnpm version on a cache miss. Set
+    # COREPACK_ENABLE_NETWORK=0 to make an offline cache miss fail immediately.
     corepack pnpm "$@"
   elif command -v pnpm >/dev/null 2>&1; then
     pnpm "$@"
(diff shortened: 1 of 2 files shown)


haelyra (approved):
Thank you for chasing down this Git Bash failure, and sorry for the review delay. Resolving pnpm through Corepack first is the right fix for repositories that pin pnpm but do not expose a standalone pnpm executable in the hook PATH, while the direct-pnpm fallback preserves existing setups. The audit path uses the same resolver and failure remains closed. I verified Bash syntax, all 27 Codex hook tests, ESLint, and diff cleanliness at c022a105. This is ready to merge after the serial exact-main gate clears.

haelyra:
Thank you for improving the Git Bash pre-push path, and I am sorry this sat for so long. Your Corepack-aware pnpm resolution is now on main through #2902, with attribution preserved and an added regression for a genuinely missing executable. The exact main commit passed all 42 CI jobs and CodeQL. I am closing this PR only because the approved work has landed. Please keep contributing.

PR #2596, opened 2026-07-27
fix(hooks): correct Opus rate and stop guessing Sonnet for unknown models

AmirF194:
## Summary

`RATE_TABLE.opus` in `scripts/hooks/cost-tracker.js` still carries the old $15/$75 per 1M token rate, 3x the current $5/$25. `getRates()` also fell through to Sonnet rates for any model matching neither `haiku` nor `opus`, so a newer model like `claude-fable-5` was silently priced as Sonnet instead of at its own (higher) rate.

## Fix

- Updated the Opus entry to $5/$25 (in/out) and $6.25/$0.50 (cache write/read), and added a Fable entry at $10/$50 / $12.50/$1.00.
- `getRates()` now returns `null` for a model it doesn't recognize instead of guessing. The caller records `estimated_cost_usd: 0` and a new `cost_estimated: false` flag on that row, so a reader of `costs.jsonl` can distinguish an unpriced model from a genuinely free one.

## Testing

- `tests/hooks/cost-tracker.test.js`: added 3 cases (Opus at the corrected rate, Fable at its own rate, an unrecognized model marked unestimated). All 3 fail on unmodified `main` (Opus priced at $90 instead of $30, Fable at $18 instead of $60, an unknown model guessed at $18 instead of $0) and pass with this fix.
- Full existing suite for this hook (`node tests/hooks/cost-tracker.test.js`, 12 cases) and the repo-wide suite (`node tests/run-all.js`, 3300 cases) both green; `npm run lint` and `npm run coverage` clean.
- Did not check the `>200K-token` or `1h-cache` billing tiers mentioned in this file's own header; this hook's rate table is documented there as an approximation for exactly that reason, and this fix only corrects the flat per-model rates and the fallback behavior.

### DIFF
--- a/scripts/hooks/cost-tracker.js
+++ b/scripts/hooks/cost-tracker.js
@@ -73,14 +73,20 @@ function readHarnessCost(sessionId, maxAgeSeconds) {
 const RATE_TABLE = {
   haiku:  { in: 0.80,  out: 4.0,  cacheWrite: 1.00,  cacheRead: 0.08 },
   sonnet: { in: 3.00,  out: 15.0, cacheWrite: 3.75,  cacheRead: 0.30 },
-  opus:   { in: 15.00, out: 75.0, cacheWrite: 18.75, cacheRead: 1.50 }
+  opus:   { in: 5.00,  out: 25.0, cacheWrite: 6.25,  cacheRead: 0.50 },
+  fable:  { in: 10.00, out: 50.0, cacheWrite: 12.50, cacheRead: 1.00 }
 };
 
+// Returns null for a model name matching none of the table above, rather
+// than guessing: a silent Sonnet default previously mispriced every new
+// model (most recently Fable) until someone noticed (#2574).
 function getRates(model) {
   const m = String(model || '').toLowerCase();
-  if (m.includes('haiku')) return RATE_TABLE.haiku;
-  if (m.includes('opus'))  return RATE_TABLE.opus;
-  return RATE_TABLE.sonnet;
+  if (m.includes('haiku'))  return RATE_TABLE.haiku;
+  if (m.includes('fable'))  return RATE_TABLE.fable;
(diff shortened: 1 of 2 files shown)


daltino (approved):
This PR improves pricing accuracy by adjusting the Opus rate to reflect current values and adds Fable as a new model. Tests confirm the corrected rates and the removal of the default guess behavior, adhering to best practices. Changes align well with the contributing guidelines for hooks — concise, useful, and tested.

AmirF194:
Flagging for visibility: #2621 also touches getRates() for the same issue (adds a fable-specific carve-out). This PR fixes the underlying bug more generally (unknown models return null instead of silently mispricing as sonnet), so any future new model is covered too, not just fable. Happy to rebase around whichever lands first if there's a conflict, or to drop this if you'd rather go with #2621.

AmirF194:
Checking in after a week with no maintainer review yet, no rush. Happy to rebase or split the fix further if that would help.

AmirF194:
andrest's #2691 (opened 2026-08-06) also fixes the opus rate fallthrough this PR targets, and it is more correct than this one: it splits opus into a legacy bucket ($15/$75, Claude 3 Opus and Opus 4.0/4.1) and a current bucket ($5/$25, Opus 4.5+), while this PR prices every model containing "opus" at $5/$25 regardless of generation. That would misprice a Claude 3/4.0/4.1 Opus session as roughly a third of its real cost. #2691 is the more complete fix. Happy to stand down here if you'd rather take that one.

haelyra:
Thank you for diagnosing the stale Opus rate and, especially, for recognizing that unknown models should not receive a confident Sonnet estimate. Your follow-up correctly identified that #2691 has the more complete generation-aware rate table, and we have asked that PR to carry forward this PR’s honest unknown-model contract before it lands. We are going to close this partial implementation to keep the queue focused on that consolidated path. We appreciate both the original bug report and the thoughtful handoff.

PR #2605, opened 2026-07-27
fix(suggest-compact): don't quote a percentage against an assumed window

y2nq:
## Problem

The context signal always renders `N% of <window> window`, including when the window size is the **assumed 200k default** rather than a detected value.

On a 1M-context session whose transcript carries no `[1m]` marker, this produces:

```
[StrategicCompact] Context ~194k tokens (97% of 200k window) - consider /compact at the next logical boundary
```

while actual usage is ~19%. Observed in real use today: the model faithfully relayed "context is 97% full" to the user across several sessions, the user compacted on the false alarm, lost context, and the resulting quality drop was initially diagnosed as a model regression. The status line — which reads the harness's own `context_window.used_percentage` — showed ~20% at the same moment.

## Why it fires exactly where it can't know

The context threshold defaults to 80% of the window (160k on 200k), so the signal fires precisely in the **160k–200k band**, which is the one range where the size cannot be determined:

- below 160k — nothing fires
- 160k–200k — window unknown, but a confident percentage is printed
- above 200k — the observed-tokens fallback correctly infers 1M, and the message becomes accurate again

So the warning appears and then *disappears* as the context grows past 200k.

## Why the model id can't fix it

A model tier can ship both a 200k and a 1M variant under one id, with the `[1m]` suffix present in settings but **absent from the transcript's `message.model` field**. Verified locally: 143/143 assistant records in a `[1m]` session recorded the bare id. Adding such a family to `KNOWN_MODEL_WINDOW_TOKENS` would therefore be wrong in the opposite direction — it would claim 1M for the 200k variant too. The transcript carries no window field to consult.

## Change

Stop asserting what isn't known.

- `resolveContextWindow(tokens, model)` (new, exported) returns `{ windowTokens, inferred }`. `inferred: false` means the size came from evidence — env override, `[1m]` marker, known family, or observed tokens > 200k. `inferred: true` means every check fell through to the 200k default.
- `isContextWindowInferred(tokens, model)` (new, exported) for callers that only need the flag.
- `resolveContextWindowTokens()` keeps its exact signature and semantics — now a one-line delegate.
- The hook omits the percentage and window label when the size was assumed:

```
[StrategicCompact] Context ~194k tokens - consider /compact at the next logical boundary
```

Token count, threshold, bucketing, and firing behaviour are unchanged. Users who know their window can still set `ECC_CONTEXT_WINDOW_TOKENS` and get the percentage back.

## Tests

5 new cases in `tests/lib/transcript-context.test.js` covering inferred vs. detected for each detection path. `node tests/lib/transcript-context.test.js` → 41 passed, 0 failed.

Full suite, measured on both trees by checking out each commit:

| | tests | passed | failed |
|---|---|---|---|
| unmodified `main` (b6f461a3) | 3354 | 3347 | **7** |
| this branch | 3359 | 3352 | **7** |

The +5 are the new cases. **The failure count is unchanged** — all 7 are pre-existing, 3 of them in `tests/hooks/suggest-compact.test.js`, and they are untouched here.

ESLint clean on all three changed files.

### DIFF
--- a/scripts/hooks/suggest-compact.js
+++ b/scripts/hooks/suggest-compact.js
@@ -38,7 +38,8 @@ const {
   resolveContextThreshold,
   resolveContextInterval,
   computeContextBucket,
-  formatWindowLabel
+  formatWindowLabel,
+  isContextWindowInferred
 } = require('../lib/transcript-context');
 
 const COUNTER_FILE_PREFIX = 'claude-tool-count-';
@@ -185,8 +186,13 @@ function buildContextSuggestion(transcriptPath, bucketFile, env) {
     writeFile(bucketFile, String(bucket));
 
     const approxTokens = `${Math.round(usage.tokens / 1000)}k`;
-    const percent = Math.round((usage.tokens / windowTokens) * 100);
-    return `[StrategicCompact] Context ~${approxTokens} tokens (${percent}% of ${formatWindowLabel(windowTokens)} window) - consider /compact at the next logical boundary`;
+    // Only quote a percentage when the window size was actually detected.
+    // Against an assumed 200k default the denominator is a guess, and a
+    // "97% of 200k window" line on a 1M session triggers needless compaction.
+    const scale = isContextWindowInferred(usage.tokens, usage.model)
+      ? ''
+      : ` (${Math.round((usage.tokens / windowTokens) * 100)}% of ${formatWindowLabel(windowTokens)} window)`;
+    return `[StrategicCompact] Context ~${approxTokens} tokens${scale} - consider /compact at the next logical boundary`;
   } catch (err) {
     log(`[StrategicCompact] Context signal skipped: ${err.message}`);
     return null;
--- a/scripts/hooks/suggest-compact.js
+++ b/scripts/hooks/suggest-compact.js
@@ -34,12 +34,11 @@ const {
 } = require('../lib/utils');
 const {
   readLatestContextTokens,
-  resolveContextWindowTokens,
+  resolveContextWindow,
   resolveContextThreshold,
   resolveContextInterval,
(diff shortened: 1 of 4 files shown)


daltino (approved):
This PR improves the robustness of context token handling by introducing `isContextWindowInferred` to distinguish between inferred and evidence-based window sizes. The changes are clear, well-tested with additional test cases, and align with the repo's standards for hooks and context management.

haelyra:
Thank you for this focused fix, and sorry it has been waiting. The core idea is right: when the 200k denominator is only an assumption, ECC should report the measured token count without presenting a misleading percentage.

I pushed a small maintainer follow-up in 026e4510. It resolves the window once instead of doing duplicate detection work, and updates the hook-level regression test to assert the honest no-percentage output. The focused suites now pass: 41 transcript-context tests and 44 suggest-compact tests, plus ESLint and syntax checks.

This is ready from my review side. I will merge it after the current exact-main CI gate and this updated head are green. Thanks again for catching a real, user-visible problem.

haelyra (approved):
Thank you for this focused correctness fix. Reporting measured context tokens without a percentage when the window is inferred avoids presenting an assumption as fact, while preserving precise percentages for native or configured windows. The exact head passes all 41 transcript-context tests and 44 suggest-compact tests plus ESLint, and is ready for the serial main-branch merge gate.

haelyra:
Thank you for catching the misleading context-window percentage, and I am sorry for the delay. The truthful inferred-context behavior is now on main through #2902 with your authorship preserved, plus coverage that restores the host environment after testing. The exact main commit passed all 42 CI jobs and CodeQL. Closing this PR only because the approved fix has landed. Please keep sending this kind of precision fix.

PR #2609, opened 2026-07-28
fix: add claude-opus-5 to KNOWN_MODEL_WINDOW_TOKENS

HouMinXi:
## Problem

`claude-opus-5` is missing from `KNOWN_MODEL_WINDOW_TOKENS` in `scripts/lib/transcript-context.js`. When `resolveContextWindowTokens()` is called with `model = "claude-opus-5"` and `tokens < 200000`, none of the detection paths match:

1. No env override set
2. No `[1m]` marker in the model id
3. Token count below 200k threshold

It falls back to `STANDARD_CONTEXT_WINDOW_TOKENS` (200k), incorrectly triggering compact warnings. The `tokens > 200000` heuristic on line 186 catches it later, but only after the session has already used 200k+ tokens — exactly when compact is most wasteful.

**Verified**: 250156 tokens at 25% usage = ~1M context window.

## Bad behavior shape

This bug only manifests in the first 20% of a 1M session. Past 200k tokens, it self-heals via the heuristic branch. This makes it particularly misleading: the compact warning appears precisely when compacting would waste the most tokens.

## Fix

Add `['claude-opus-5', LARGE_CONTEXT_WINDOW_TOKENS]` to the table. Same pattern as #2461 (fable-5/mythos-5) and #2290 (Opus 4.x).

`claude-sonnet-5` is not included — no empirical window-size data available yet. Guessing would be the same class of error.

## Workaround

Set `ECC_CONTEXT_WINDOW_TOKENS=1000000` in `~/.claude/settings.json` env. This is model-agnostic and survives plugin upgrades.

Refs: #2290, #2461, #2468

### DIFF
--- a/scripts/lib/transcript-context.js
+++ b/scripts/lib/transcript-context.js
@@ -35,6 +35,7 @@ const LARGE_WINDOW_MODEL_MARKER = '[1m]';
 // Checked in order, first match wins. Best-effort and expected to lag new
 // releases; the env override remains the escape hatch for unlisted models.
 const KNOWN_MODEL_WINDOW_TOKENS = [
+  ['claude-opus-5', LARGE_CONTEXT_WINDOW_TOKENS],
   ['claude-fable-5', LARGE_CONTEXT_WINDOW_TOKENS],
   ['claude-mythos-5', LARGE_CONTEXT_WINDOW_TOKENS]
 ];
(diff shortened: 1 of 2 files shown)


daltino (approved):
This PR adds support for recognizing the "claude-opus-5" model in the `KNOWN_MODEL_WINDOW_TOKENS` array with corresponding test coverage, ensuring consistency and functionality within context window token resolution. The change is concise, follows contribution guidelines, and includes adequate testing to validate the addition. Looks good to merge!

haelyra:
Thank you for catching this so quickly after the Opus 5 release. Adding the exact claude-opus-5 family prevents premature compact warnings, and the focused regression verifies the 1M-window path. Anthropic’s published model contract matches the implementation, so this is ready to merge.

Issue #2573, opened 2026-07-25
gateguard: fact-force frequency knobs are undocumented; routine-Bash gate fires on read-only commands


kaiser-data:
## Summary

The fact-forcing gate fires often enough in normal sessions to read as noise rather than signal, and the knobs that would tune it are effectively undiscoverable. Three of the four `GATEGUARD_*` env vars implemented in `scripts/hooks/gateguard-fact-force.js` appear in **no** markdown file in the repo, and the deny message points users only at the two nuclear options (`ECC_GATEGUARD=off`, `ECC_DISABLED_HOOKS`).

This is not a request to weaken the destructive-command gate — that one earns its keep. It's about the two gates that fire on operations with no destructive potential.

## What I actually hit

First Bash call of a fresh session was a plain directory listing (`ls`, `find`, no writes). It was denied by the routine-Bash gate:

```
[Fact-Forcing Gate]

Before the first Bash command this session, present these facts:

1. The current user request in one sentence
2. What this specific command verifies or produces

Recovery: if GateGuard is blocking setup or repair work, run this session
with `ECC_GATEGUARD=off` or add `pre:bash:gateguard-fact-force` to
`ECC_DISABLED_HOOKS`.
```

The recovery hint offers only "turn the whole thing off." `GATEGUARD_BASH_ROUTINE_DISABLED=1` — which disables exactly this gate while leaving destructive gating fully intact — is implemented at `gateguard-fact-force.js:137-139` and `:1261-1263`, but a user has no way to learn it exists short of reading the hook source.

## Documentation gaps (verified against ECC 2.0.0, commit b6fe5a7)

`grep -rl "<VAR>" --include="*.md" .` across the repo:

| Env var | Documented? |
|---|---|
| `GATEGUARD_FACT_FORCE_FULL_DENIALS` | `skills/gateguard/SKILL.md:102` only |
| `GATEGUARD_BASH_ROUTINE_DISABLED` | **nowhere** |
| `GATEGUARD_EXEMPT_GLOBS` | **nowhere** |
| `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` | **nowhere** |

Worse than absent — actively misdirecting. `skills/gateguard/SKILL.md:128` says:

> Use `.gateguard.yml` to ignore paths like `.venv/`, `node_modules/`, `.git/`.

[...]

PR #2611, opened 2026-07-28
docs(gateguard): document the graduated gate controls

Souptik96:
Refs #2573 — documents the GateGuard tuning knobs. **Docs + one surface test; no hook behaviour changes.**

## Scope

#2573 raises two things. This PR takes **only the first**:

- ✅ *"the knobs that would tune it are effectively undiscoverable"* — addressed here.
- ❌ Changing gate behaviour (defaulting `GATEGUARD_BASH_ROUTINE_DISABLED` on, broadening the read-only Bash allowlist to `ls`/`pwd`/`cat`/`find`/`rg`) — **not touched.** That's a product call about how aggressive the gate should be, and it's yours to make. Happy to implement whichever way you decide.

## What was missing

The hook reads five `GATEGUARD_*` variables that appear nowhere in `skills/gateguard/SKILL.md`, so the only discoverable escape hatch was `ECC_GATEGUARD=off` — which also disables the destructive-Bash gate, the one that earns its keep. The reporter's workaround block is exactly the config that should have been documented.

| Variable | Was |
|---|---|
| `GATEGUARD_BASH_ROUTINE_DISABLED` | undocumented anywhere |
| `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` | undocumented anywhere |
| `GATEGUARD_DISABLED` | undocumented anywhere |
| `GATEGUARD_STATE_DIR` | undocumented anywhere, yet named in a runtime warning users can hit |
| `GATEGUARD_EXEMPT_GLOBS` | only in `docs/releases/2.1.0/release-notes.md` |
| `GATEGUARD_FACT_FORCE_FULL_DENIALS` | already documented — folded into the same table so there's one lookup point |

**Three of those five aren't in #2573.** The issue names two; the surface test below found the rest.

Every default and accepted value is read from the hook rather than assumed:

- `GATEGUARD_BASH_ROUTINE_DISABLED` accepts `1`/`true`/`on`/`enabled`/`enable`/`yes` via `ECC_ENABLE_VALUES` + `normalizeEnvValue` (trimmed, lowercased); `GATEGUARD_DISABLED` recognises `1` only — that asymmetry is real and now stated.
- `GATEGUARD_STATE_DIR` defaults to `~/.gateguard` (`$HOME`/`%USERPROFILE%`/`/tmp` fallback chain).
- `GATEGUARD_FACT_FORCE_FULL_DENIALS` default `3`; `0` condenses from the very first denial — matching the existing assertion in `tests/hooks/gateguard-fact-force.test.js` ("Budget 0: condensed from the very first denial").
- `GATEGUARD_EXEMPT_GLOBS` glob semantics (`*` within a segment, `**` across, `?` single char), matched against the normalized forward-slash lowercased path, fail-open on a malformed pattern.

## The test

`tests/ci/gateguard-env-documented.test.js` asserts every `GATEGUARD_*` variable the hook reads is documented, and — in the other direction — that the doc names no variable the hook has stopped reading. It follows the `tests/ci/*-surface` pattern already in the repo and is what caught the three extra knobs.

## Verification

```
$ node tests/ci/gateguard-env-documented.test.js
  ✓ hook and skill doc both exist
  ✓ hook reads at least one GATEGUARD_* variable
  ✓ every GATEGUARD_* variable the hook reads is documented
  ✓ the documented knobs are the ones the hook actually reads
  Passed: 4, Failed: 0

# negative control: revert only the doc, keep the test
  Failed: 1 — undocumented: GATEGUARD_BASH_EXTRA_DESTRUCTIVE,
  GATEGUARD_BASH_ROUTINE_DISABLED, GATEGUARD_DISABLED,
  GATEGUARD_EXEMPT_GLOBS, GATEGUARD_STATE_DIR
```

Full suite (`node tests/run-all.js`, Node 20, after `npm ci --ignore-scripts`): the failure profile on this branch is **line-for-line identical** to pristine `main` on the same machine — same 7 suites reporting the same counts (18 failures total), all pre-existing and unrelated (they reproduce on a clean checkout here). `npx markdownlint skills/gateguard/SKILL.md` exits 0 on both this branch and pristine. `yarn.lock` was touched by a local `npm ci` and has been restored, so no lockfile churn is in the commit.


Souptik96:
All four addressed in `45a6b03` — three were correct as filed, and the glob one was correct in substance but pointed at the docs rather than the matcher.

**1. `GATEGUARD_DISABLED` contradicted its own table header.** Correct, and it was my error: the table was introduced as "these narrow the gate, destructive-Bash keeps running" and then listed a full kill switch. It now has its own *Turning the gate off completely* section alongside `ECC_GATEGUARD`. Checking the code while moving it also caught a second mistake of mine — I'd written that `ECC_GATEGUARD` accepts `no`. It doesn't: `ECC_DISABLE_VALUES` is `0`, `false`, `off`, `disabled`, `disable`. Corrected, and the `1`-only asymmetry of `GATEGUARD_DISABLED` is now called out explicitly next to it.

**2. The `**` glob examples.** Correct. I reproduced the hook's translation (`.replace(...).split('**').map(...).join('.*')`) and ran it:

```
glob **/tests/**   ->  /.*\/tests\/.*/
   /home/u/repo/tests/foo.js   MATCH
   tests/foo.js                no match
glob **/*.test.*   ->  /.*\/[^/]*\.test\.[^/]*/
   /home/u/repo/foo.test.js    MATCH
   foo.test.js                 no match
```

So a leading `**/` compiles to `.*/` and needs at least one preceding separator. I fixed the **docs** rather than the matcher: `normalizeForMatch` only swaps `\` for `/` and lowercases, so the string matched is whatever `file_path` the payload carried — absolute for Claude Code tool calls, which is why the pattern works in practice. Widening `**` to match zero segments would change gate behaviour for every existing user's config, and this PR is deliberately docs-only (#2573's behaviour half is a maintainer decision). The docs now state the separator requirement and the example carries both `**/tests/**` and `tests/**`. Say the word if you'd rather have the matcher change and I'll open it separately.

**3. Reverse-drift check compared against raw source.** Correct — `hookSource.includes(name)` would let a knob deleted from the code still pass if its name survived in a comment or an error string, which defeats the point of that assertion. It now compares against the parsed `envNames` set.

**4. Immutable accumulator.** Done — `readGateguardEnvNames` collects both match arrays and constructs a single `Set` at return instead of mutating via `Set#add`.

Re-verified after the changes: the surface test is 4 pass / 0 fail; removing a knob from the doc fails it naming that knob; adding a doc-only knob the hook never reads fails the reverse-drift assertion (`documented but unread by the hook: GATEGUARD_NOT_READ_BY_HOOK`); `markdownlint skills/gateguard/SKILL.md` exits 0. `yarn.lock` restored, so no lockfile churn in either commit.

Souptik96:
Addressed in `1a8e40c`. The finding was right on both counts, and I checked each against the parser as it actually shipped in `45a6b03` rather than assuming:

```
destructured read              old parser sees: (nothing)
aliased read                   old parser sees: (nothing)
computed template read         old parser sees: (nothing)
read named only in a comment   old parser sees: GATEGUARD_ONLY_IN_A_COMMENT
read named only in a string    old parser sees: GATEGUARD_ONLY_IN_A_STRING
```

So the bypass was real (an undocumented knob added via destructuring would have sailed through the documentation check) and the comment/string false positive was real too — latent in the current hook only because its stray `GATEGUARD_*` mentions at lines 54/97 (comments), 87 (a template literal) and 1163 (a string) happen not to carry a `process.env.` prefix. A future comment that did would have counted as a read.

**What I did not do: pull in an AST parser.** The other `tests/ci/*-surface` tests are all regex-based and this is a documentation-surface guard, so a parser dependency felt disproportionate. I took the second option you offered — enforce a narrow convention and test it — which closes the hole without the dependency:

1. **Scan code only.** Comments, string literals, template-literal *text* and regex literals are blanked before scanning. Code inside a template's `${...}` is preserved, since that is real code and can contain a read. Blanking replaces characters with spaces rather than deleting them, so length and line numbers are preserved — that invariant is asserted, and it is what lets `process.env[...]` keys be located in the blanked code (proving the access is code) and then read back out of the raw source at the same offset. Blanking the literal would otherwise erase the key itself.
2. **Enforce the supported forms.** `process.env.GATEGUARD_X` and `process.env['GATEGUARD_X']` are supported; destructuring, aliasing, spreading, `Object.keys/values/entries/assign/fromEntries` over `process.env`, and computed keys all fail a new guard that says explicitly: keep to the convention, or teach `readGateguardEnvNames` the new form. A knob can no longer become invisible by accident — the test breaks and names the form it saw.

One thing worth flagging: writing the blanker surfaced a bug in my own first attempt. A closing backtick was being treated as *opening* a new template literal, which swallowed the rest of the file and silently dropped four of the six real reads. The self-checks caught it. That is a fair illustration of your point about hand-rolled scanning, which is why the convention guard exists alongside the blanker rather than instead of it — the guard is the part that fails loudly when the scanner is out of its depth.

### Verification

`node tests/ci/gateguard-env-documented.test.js` — **4 pass → 10 pass / 0 fail**.

Negative controls, each applied to a copy of the hook or the doc and then reverted:

| Control | Expectation | Result |
|---|---|---|
| unmodified branch | all pass | 10 pass / 0 fail |
| append `const { GATEGUARD_SNEAKY_ONE } = process.env` | access guard fails | ✗ `destructuring from process.env` |
| append `const envAlias = process.env` + property read | access guard fails | ✗ `process.env aliased to a binding` |
| append undocumented `process.env['GATEGUARD_UNDOCUMENTED_BRACKET']` | doc check fails | ✗ `undocumented in skills/gateguard/SKILL.md` |
| append undocumented `process.env.GATEGUARD_UNDOCUMENTED_DOTTED` | doc check fails | ✗ `undocumented in skills/gateguard/SKILL.md` |
| append fake reads in a comment **and** a string | nothing trips | 10 pass / 0 fail |
| rename a knob in the doc only | doc + reverse-drift fail | ✗ both, 8 pass / 2 fail |
| restore everything | all pass | 10 pass / 0 fail |

Full suite, `node tests/run-all.js` on Node 20 (Windows):

| | Total | Passed | Failed |
|---|---|---|---|
| `45a6b03` | 3282 | 3261 | 21 |
| `1a8e40c` | 3288 | 3267 | 21 |

`+6 tests / +6 passed`, which is exactly the six self-checks added, and the **failure set is byte-identical** on both commits (29 reported lines across 9 suites). Those 21 are pre-existing on this box and unrelated to this PR — Windows symlink handling in `lib/memory-vault` and `scripts/memory-mcp`, PowerShell-vs-bash resolution in `hooks/plugin-hook-bootstrap`, and the `hooks/observe-entrypoint-allowlist` / `scripts/codex-hooks` suites. No documentation changed in this commit, so `skills/gateguard/SKILL.md` is untouched since `45a6b03`.

haelyra:
Thank you for turning GateGuard’s graduated controls into a documented, machine-checked contract, and sorry for the long wait. The environment-variable table is useful on its own; the code-aware drift test is especially valuable because it rejects undocumented access forms instead of treating comments or strings as real configuration reads.

I merged current main into ceb99304 and reran the focused gates. All 10 documentation-contract tests pass against the current GateGuard hook, all 286 skills validate, the skill passes Markdown lint, the test passes ESLint, and the diff is clean. Hosted CI is now rerunning on the exact refreshed head; once green, this is ready for final approval and the serial merge gate.

haelyra (approved):
The documented GateGuard controls match current code, and all focused environment-contract tests plus skill validation, Markdown lint, ESLint, and diff checks pass locally. Approved, with merge still held for the hosted gate.

haelyra:
Thank you for making GateGuard graduated controls easier to understand, and I am sorry this waited. The documentation and control guidance are now on main through #2902 with your authorship preserved, alongside the reviewed GateGuard compatibility fixes. The exact main commit passed all 42 CI jobs and CodeQL. Closing this source PR only because the work has landed. Please keep contributing.

PR #2616, opened 2026-07-29
fix(hooks): make post-edit-typecheck.js incremental

Ivan2san:
## Summary

The `post-edit-typecheck.js` PostToolUse:Edit hook runs `npx tsc --noEmit` on every edit to a `.ts`/`.tsx` file. Without `--incremental`, every single edit triggered a full cold typecheck of the whole project — on a large TS project (observed on two consumers of this hook set) this was regularly hitting the hook's 30s timeout, and a timeout-kill is silently swallowed by the same catch block that handles a normal `tsc` failure, so the check could fail to complete with zero visible signal.

This adds `--incremental` to the `tsc` invocation so it reuses the project's `.tsbuildinfo` cache (or creates one at TypeScript's default location) between runs — repeat checks in the same project only re-verify what changed instead of a full cold check every edit.

## Measured impact

On a real project using this hook set:
- Cold run (no `.tsbuildinfo` cache yet): ~42s
- Warm run (cache present, no source changes): ~4.7s
- Warm run (cache present, one file touched): ~5.6s

Prior to this fix, cold-every-time behavior was regularly hitting the 30s hook timeout.

## Follow-up

Even the warm-run numbers above still block the agent loop synchronously for several seconds per edit. A separate, isolated effort is underway to background the actual `tsc` invocation (detached worker + status-file delivery) so the hook call itself returns near-instantly regardless of check duration — that's a bigger change with its own test plan and isn't bundled into this PR.

## Test plan

- [x] All 190 existing tests in `tests/hooks/hooks.test.js` pass unchanged
- [x] Manually verified cold → warm timing improvement on a real TS project using this hook set

### DIFF
--- a/scripts/hooks/post-edit-typecheck.js
+++ b/scripts/hooks/post-edit-typecheck.js
@@ -9,22 +9,22 @@
  * and reports only errors related to the edited file.
  */
 
-const { execFileSync } = require("child_process");
-const fs = require("fs");
-const path = require("path");
+const { execFileSync } = require('child_process');
+const fs = require('fs');
+const path = require('path');
 
 const MAX_STDIN = 1024 * 1024; // 1MB limit
-let data = "";
-process.stdin.setEncoding("utf8");
+let data = '';
+process.stdin.setEncoding('utf8');
 
-process.stdin.on("data", (chunk) => {
+process.stdin.on('data', chunk => {
   if (data.length < MAX_STDIN) {
     const remaining = MAX_STDIN - data.length;
     data += chunk.substring(0, remaining);
   }
 });
 
-process.stdin.on("end", () => {
+process.stdin.on('end', () => {
   try {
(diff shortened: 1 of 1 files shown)


daltino (approved):
This PR improves `post-edit-typecheck.js` by switching to incremental updates and making minor code style adjustments (e.g., single quotes, arrow functions). It's a clean, low-risk update that aligns with code formatting best practices. Nice work!

haelyra:
Thank you for measuring the warm-cache improvement and isolating the repeated cold typecheck cost. The active formatting and typecheck route has since moved to stop-format-typecheck.js, and current hooks.json no longer invokes post-edit-typecheck.js, so this patch optimizes a dormant script while adding an unmanaged .tsbuildinfo write. We are closing this version; a focused change to the active Stop hook with a controlled cache path and invalidation regression would be welcome.

PR #2624, opened 2026-07-29
fix(hooks/gateguard): close destructive-detector gaps, stop gating scratch cleanup

rbotte:
The destructive-command gate had drifted in both directions: it missed
several operations that really do destroy, and it fired on prose and on
throwaway directories.

Coverage gaps closed:

- `dd` could never fire on a real invocation. `/\bdd\s+if=\b/` needs a
  word character after `=`, but every real use continues with a path, so
  `dd if=/dev/zero of=/dev/sda` did NOT match while the inert `dd if=x`
  did. `of=` — the operand that actually writes — was never matched at
  all. `dd` now has a command-position rule covering both operands.
- `rm -r` without `-f` was waved through: the rule required BOTH flags,
  so a recursive delete of a source tree passed unchallenged.
- `find -delete` had no rule; only `-exec <destroyer>` was covered.
- `shred`, a standalone `unlink`, and the coreutils `truncate` had no
  rule. `unlink` was recognised only inside `find -exec`.

Precision fixes:

- `truncate` was matched as a bare word, so every mention of it was
  destructive: the Jinja/Django filter of the same name, a branch name,
  a file name, a line of prose in a commit body. SQL truncation always
  names a table, so the pair is now required; the coreutils binary is
  caught in command position instead.
- Delete targets that resolve to scratch space, build caches,
  virtualenvs or test artefacts no longer raise the gate. Bare `/tmp` is
  deliberately excluded, so `rm -rf /tmp/junk` is still challenged.

Supporting changes:

- `stripQuotedStrings` can now stash quoted literals in a slot table
  instead of blanking them. A quoted argument still tokenizes as one
  opaque token, so `git commit -m '(rm -rf x)'` stays inert, but the
  literal is recoverable — without it `rm -rf "$SCRATCH/build"` cannot
  be judged at all. Both quote styles are matched in one left-to-right
  alternation rather than two sequential passes, which is how the shell
  reads mixed quoting.
- `NAME=value` assignments in the same command line are resolved before
  a target is judged, since scripts define the scratch root in the very
  command that deletes from it. `$(mktemp -d)` is treated as disposable.
  A target still carrying an unresolved `$VAR`, a bare root, or any
  operand that is not clearly derived keeps the gate loud.
- `isDestructiveQuoteAware` now receives the same context and the new
  rules. Without this the disposable-target decision made in the main
  pass was silently reversed by the later one, and `sh -c` could have
  bypassed the additions. The GHSA-4v57-ph3x-gf55 regression tests still
  pass.

Measured against 20,737 distinct Bash commands from 14 days of real
agent traffic: denials fall from 273 to 138 (-49%) with zero commands
newly challenged. Every one of the 41 distinct paths that stopped
raising the gate is a scratchpad, mktemp directory, .venv,
test-results, .playwright-mcp or gitignored plan directory.

19 tests added; suite goes 144 -> 163, all passing.

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -44,11 +44,43 @@ const BASH_HOOK_ID = 'pre:bash:gateguard-fact-force';
 const ECC_DISABLE_VALUES = new Set(['0', 'false', 'off', 'disabled', 'disable']);
 const ECC_ENABLE_VALUES = new Set(['1', 'true', 'on', 'enabled', 'enable', 'yes']);
 
-// SQL-keyword + dd patterns stay as a single regex — they are stable
-// phrases without shell-flag ordering concerns. Quoted strings are
-// stripped before this regex runs so a commit message mentioning
-// "drop table" no longer triggers a false positive.
-const DESTRUCTIVE_SQL_DD = /\b(drop\s+table|delete\s+from|truncate|dd\s+if=)\b/i;
+// SQL phrases stay as a single regex — they are stable phrases without
+// shell-flag ordering concerns. Quoted strings are stripped before this
+// regex runs so a commit message mentioning "drop table" no longer
+// triggers a false positive.
+//
+// Two members used to live here and no longer do:
+//
+//   `truncate` was matched bare, so every mention of the word was
+//   destructive — the Jinja/Django filter, a branch name, a filename, a
+//   line of prose in a commit body. SQL truncation always names a table,
+//   so require the pair; the coreutils binary is caught in command
+//   position by `isDestructiveSimpleCommand`.
+//
+//   `dd\s+if=` could never fire on a real invocation. The trailing `\b`
+//   needs a word character after `=`, but every real use continues with
+//   a path: `dd if=/dev/zero of=/dev/sda` did NOT match, while the inert
+//   `dd if=x` did. `of=` — the operand that actually writes — was never
+//   matched at all. `dd` now has a command-position rule.
+const DESTRUCTIVE_SQL = /\b(drop\s+table|delete\s+from|truncate\s+table)\b/i;
+
+// Paths whose removal costs nothing: agent scratch space, build caches,
+// virtualenvs, test artefacts. Bare `/tmp` is deliberately absent — only
+// clearly-derived directories qualify, so `rm -rf /tmp/junk` still gets
+// challenged.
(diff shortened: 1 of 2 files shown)


rbotte:
Separate from this PR's diff, the same measurement turned up something structural that I think is worth a maintainer's opinion before anyone writes code for it. Filing it here rather than as a patch, because the fix is an architecture choice, not a detector tweak.

### The retry key is set by the denial itself

In all three gates, `markChecked(key)` runs *before* `denyResult(...)`:

```js
if (!isChecked(key)) {
  if (!markChecked(key)) return allowWithStateWarning();
  return denyResult(...);   // key already recorded above
}
return rawInput;             // second attempt: allowed
```

So the second attempt always succeeds, whether or not a single fact was presented. A `PreToolUse` hook only ever sees the tool input — never the assistant's reply — so nothing at this event can actually read the facts the message asks for. "Present the facts, then retry" is satisfied by the retry alone.

Two consequences, both measured over 14 days of one operator's real traffic (2,605 denials, 120 sessions):

1. **Byte-identical resubmission passes.** 37 of 139 destructive denials (27%) were followed by the exact same command, which then ran.
2. **Refining a command re-blocks it.** The destructive key is `sha256(command)`, so editing the command — adding a flag, narrowing a path, fixing a typo — produces a new key and a fresh denial. The gate is cheaper to satisfy by repeating than by thinking, which is the opposite of the intent.

The first-touch gates have the same shape with a per-path key, so the cost is per file rather than per command: 2,454 of the 2,605 denials (94.6%) came from those two gates, and none of them blocked anything a second unchanged attempt did not immediately perform.

### A possible direction

The signal for "did it investigate" is not text — it is an action, and actions *are* observable at `PostToolUse`. Recording `Read`/`Grep`/`Glob` calls there, and allowing a retry only when at least one of them touched the target since the denial, would turn "be denied once" into "actually go and look". Still gameable, but it demands a verifiable act instead of a ceremonial paragraph, and it inverts the incentive: refining stops being punished.

That is a bigger change than this PR, and it interacts with `GATEGUARD_FACT_FORCE_FULL_DENIALS`, `GATEGUARD_EXEMPT_GLOBS` and `GATEGUARD_BASH_ROUTINE_DISABLED`, which already give operators an escape hatch for the friction. Happy to prototype it if you think the direction is right — or to drop it if you would rather the gate stay a deliberate speed bump, which is a legitimate design choice as long as the docs do not promise fact verification.

PR #2629, opened 2026-07-30
fix(opencode): install to ~/.config/opencode instead of ~/.opencode

JCN-9000:
## Summary

OpenCode natively uses `~/.config/opencode` per XDG conventions. The install target was writing to `~/.opencode`, which only worked on systems where that happened to be symlinked to `~/.config/opencode`. 

The MCP inventory reader (`scripts/lib/mcp-inventory/readers/opencode.js`) already looked in `~/.config/opencode`, so the installer and the reader were inconsistent.

## Changes

| File | Change |
|------|--------|
| `scripts/lib/install-targets/opencode-home.js:86` | `rootSegments: ['.opencode']` → `['.config', 'opencode']` |
| `scripts/install-apply.js:41` | Help text: `~/.opencode/` → `~/.config/opencode/` |
| `tests/lib/install-targets.test.js:974-975` | Assertions match new path |

The `nativeRootRelativePath: '.opencode'` stayed unchanged — that refers to the `.opencode/` source directory in the repo, not the install destination.

## Backward compatibility

Existing installs to `~/.opencode` (real dir or symlink) won't pick up future ECC installs automatically. Users can: `mv ~/.opencode ~/.config/opencode` if it's a real directory, or ensure the symlink exists.

## Verification

```
node --test tests/lib/install-targets.test.js  →  42/42 passed
```

### DIFF
--- a/scripts/install-apply.js
+++ b/scripts/install-apply.js
@@ -38,7 +38,7 @@ Targets:
   antigravity  - Install rules, workflows, skills, and agents to ./.agent/
   codex        - Install shared agents/config into ~/.codex/
   gemini       - Install project-local Gemini config into ./.gemini/
-  opencode     - Install shared commands/hooks/config into ~/.opencode/
+  opencode     - Install shared commands/hooks/config into ~/.config/opencode/
   codebuddy    - Install commands, agents, skills, and flattened rules into ./.codebuddy/
   joycode      - Install commands, agents, skills, and flattened rules into ./.joycode/
   qwen         - Install commands, agents, skills, rules, and Qwen config into ~/.qwen/
--- a/scripts/lib/install-targets/opencode-home.js
+++ b/scripts/lib/install-targets/opencode-home.js
@@ -83,7 +83,7 @@ module.exports = createInstallTargetAdapter({
   id: 'opencode-home',
   target: 'opencode',
   kind: 'home',
-  rootSegments: ['.opencode'],
+  rootSegments: ['.config', 'opencode'],
   installStatePathSegments: ['ecc-install-state.json'],
   nativeRootRelativePath: '.opencode',
   validate: defaultValidateOpencodeHome,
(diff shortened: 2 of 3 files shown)


daltino (approved):
This PR updates the installation path for `opencode` from `~/.opencode` to the more standard `~/.config/opencode`, which improves alignment with common Linux file hierarchy practices. The changes are straightforward, affect only the paths in relevant scripts and tests, and match the intent clearly. Tests have also been updated to validate the new path structure. Looks good to merge!

haelyra:
Thank you for catching the installer and inventory mismatch. Moving OpenCode’s user-wide install away from `~/.opencode` is the right direction, and the current OpenCode contract confirms that global agents, commands, plugins, and related configuration live under the XDG config directory. I’m sorry this sat without a maintainer review.

We need one compatibility-focused pass before merge:

- **Honor `XDG_CONFIG_HOME`.** The current patch hard-codes `~/.config/opencode`, while OpenCode resolves the user directory as `${XDG_CONFIG_HOME:-~/.config}/opencode`. Resolve the environment override safely and add default plus overridden-root tests.
- **Update every ECC capability and lifecycle contract.** `scripts/lib/harness-capabilities.js`, its tests, and several install-lifecycle fixtures still declare `~/.opencode`; changing only the adapter makes `show`, planning, and health output disagree with the destination. Success is one canonical reported root across plan, apply, show, doctor, and tests.
- **Preserve existing managed installs.** Simply changing the state-file root abandons ECC’s prior `~/.opencode/ecc-install-state.json`, so update/repair/uninstall can lose ownership history. Add a safe reconciliation path: detect the legacy managed root, preview the move or re-adoption, require the existing install safeguards, and prove that unmanaged OpenCode content is preserved.
- **Test a clean install and a legacy migration.** Cover default XDG, custom XDG, spaces in the home/config path, old-root-only state, both roots present, and dry-run behavior.

If you do not have bandwidth for this pass, we will add the migration work to the maintainer queue after the next release and preserve your attribution. The path diagnosis is valuable, and we appreciate the contribution.

affaan-m:
Superseded by #2863 (merged), which consolidated this work with original commits and authors retained. Closing per the consolidation PR.

PR #2634, opened 2026-07-30
fix: require explicit authorization for automatic hooks

samartomar:
## Summary

Make automatic hook behavior an explicit install-time choice without removing or weakening ECC's Full hook runtime.

- keep `core` as a non-automatic baseline and exclude `hooks-runtime`
- require one `allow` or `decline` decision for each of six hook capability groups before Full, Developer, or direct hook materialization
- preserve the complete unmodified Full hook closure when all six groups are allowed
- treat any declined group as an effective Custom install that omits the indivisible hook runtime
- narrow Core's agent and platform materialization to target-specific files
- keep remote MCP and OpenCode plugin activation out of Core
- retain the requested/effective profile and hook decisions in managed install state
- restore a newly created empty target root during uninstall while preserving pre-existing or non-empty roots

## Why

ECC's hooks are useful and intentionally powerful. The problem was not the capability itself; it was making automatic writes, command/process control, transcript-derived LLM egress, MCP activity, permission gates, and session/cost persistence implicit in broad profiles.

This change keeps those capabilities available in Full while making the administrator's choice explicit and auditable. Core remains useful for normal development without silently activating automatic behavior.

## Hook decisions

The installer requires decisions for:

1. `automatic-source-writes`
2. `command-rewrite-and-process-control`
3. `transcript-derived-llm-egress`
4. `mcp-network-and-process-activity`
5. `automatic-permission-gates`
6. `session-observation-and-cost-records`

Missing decisions hold materialization. A decline does not edit hook files; it omits the hook runtime and records the effective Custom closure.

## Verification

- focused installer/lifecycle regression set: **248 passed / 0 failed**
- full suite: **3,303 passed / 3,322 total / 19 known unrelated failures**
  - unchanged clusters: observe runner/allowlist, plugin bootstrap, plan-canvas, Codex/WSL sync, cache marketplace, unreadable-task fixture, and the existing Windows libuv crash
- fresh source-wide qualification at `97fc7369e5bc90fabce1978f7b57f9d2f1c112fc`
  - source tree: `f06a9a591edc0dffe76f646b0a1602b790aaafb0`
  - 3,380 files; 5,589 raw scanner occurrences retained
  - Core: **0 BLOCK / 0 REVIEW / 1 WARN — PASS**
  - Full: **0 BLOCK / 36 scanner REVIEW / 13 WARN**, plus six hook decisions — **REVIEW / HELD**
  - no acceptance record used
- Core lifecycle on a clean Ubuntu vet:
  - exact commit/tree acquisition
  - 17 planned operations / 306 materialized files
  - no hook or MCP activation
  - bounded `node --test` task passed
  - evidence generated
  - uninstall passed
  - pre-install home and project state restored exactly

## Notes

The Full profile remains intentionally powerful and is not made equivalent to Core. Scanner residuals for optional egress, credential, broad-permission, and package/service surfaces remain visible for enterprise review.


haelyra:
Thank you for this, and sorry it sat while the installer moved underneath it. The idea here is a good one and we're keeping it: automatic hook behavior should be an explicit, disclosed choice rather than something a broad profile turns on quietly. Your six capability groups are also the clearest plain-language description of what ECC's hooks actually do that we have written down anywhere — that framing is the part we most wanted to preserve.

What changed while this was open is that #2649 landed a guided installer that asks the hook question interactively and requires explicit `--claude-scope`/`--claude-hooks` in non-interactive mode. That rewrote most of the files this PR touches, which is where the conflicts come from — maintainer-side drift, not anything you did. It also means merging both would ask users to decide twice, once in the wizard and once through six flags. But your PR caught something the wizard genuinely missed: consent only existed on the wizard path, so `install-apply --profile core` still materialized hooks with no disclosure at all.

So we're closing this version and carrying the idea forward in #2715, which gates the apply layer so every entry point is covered, uses your capability disclosure verbatim, and fits it to the single hook decision the wizard already collects. You're credited as co-author on the commit. We'd be glad to see more from you — and for what it's worth, the apply-layer gap was a real catch that a smaller, focused PR would have gotten merged much faster.

PR #2638, opened 2026-07-31
docs(thaint-setup): catalog the 37 ECC hooks that actually run

thaint2901:
## What

Adds `thaint-setup/HOOK-CATALOG.md` — a fork-local reference for the ECC hooks that actually execute.

## Why

ECC ships 48 scripts in `scripts/hooks/` and documents none of them individually:

- `hooks/README.md` (259 lines) covers the hook *mechanism* — input schema, async hooks, `ECC_HOOK_PROFILE` / `ECC_DISABLED_HOOKS` — but names 0 of the 32 running scripts and 1 of the 37 hook IDs.
- `hooks/memory-persistence/hooks.json` has readable `purpose` fields, but only for 6 lifecycle hooks, and it self-describes as reference rather than production.

Reconstructing the live set requires `hooks/hooks.json` plus three levels of dispatcher nesting (`pre-bash-dispatcher` → `bash-hook-dispatcher`; `posttooluse-dispatcher` → `post-bash-dispatcher`). That is why nobody noticed `quality-gate.js` skipping JS/TS on the assumption that `post-edit-format` runs — it is registered nowhere.

## Contents

- 37 hook IDs / 32 scripts, grouped by event, with matcher, timeout, and purpose from each header docblock
- The 7 hooks with no docblock, marked `inferred` rather than silently described
- 8 infrastructure loaders and 8 scripts shipped without registration, with the reason for each
- Reconciles to 48 files: 32 running + 8 infrastructure + 8 unregistered

## Verification

- `node tests/run-all.js` — 3321 passed, 0 failed
- `npx markdownlint-cli thaint-setup/HOOK-CATALOG.md` — clean
- Docs-only, additive in `thaint-setup/`; no upstream-tracked file touched, so `UPSTREAM.md` divergence table needs no entry


thaint2901:
Opened against the wrong remote by mistake — this belongs on my fork, not upstream. Closing.

PR #2639, opened 2026-07-31
fix(suggest-compact): percent-of-window thresholds and honest unknown-window handling

nvtoroy:
## Problem

The strategic-compact context signal uses **absolute** token thresholds:

- `DEFAULT_CONTEXT_THRESHOLD_STANDARD = 160000` — 80% of a 200k window, sensible.
- `DEFAULT_CONTEXT_THRESHOLD_LARGE = 250000` — only **25%** of a 1M window.

So on a 1M-window model every long session starts getting "consider /compact" at 250k tokens and
re-fires every 60k for the rest of the session: dozens of premature nudges while three quarters of
the window is still free.

`resolveContextWindowTokens` also guesses. `tokens > 200k ⇒ 1M window` is:

- wrong for 400k models — #2290 notes Opus 4.x is 400k and today needs `ECC_CONTEXT_WINDOW_TOKENS`;
- a fabrication for models whose window is genuinely unknown, since the guessed window is then
  printed as a confident percentage.

#2468 fixed *recognition* for the fable/mythos families, but the threshold scaling and the guess
above it were untouched.

## Fix

1. **Percent-of-window threshold.** `DEFAULT_CONTEXT_THRESHOLD_PCT = 0.7`; the default threshold is
   `round(window * 0.7)`. One rule that holds at every size, including future ones.
2. **Family patterns after the known-model table.** `opus-4 → 400k`,
   `sonnet-4 | haiku | claude-3 | opus-3 → 200k`. Opus 4.x no longer needs the env workaround.
3. **Honest unknowns.** An unrecognized model id returns `null` instead of a guess, as does a
   missing model id whose token count already exceeds 200k (provably larger, size unknown).
   `null` → threshold `UNKNOWN_WINDOW_THRESHOLD_TOKENS = 300000` (past every standard window, so
   the nudge is still truthful) and a percentless message:

   ```
   [StrategicCompact] Context ~310k tokens (window unknown for <model>; set ECC_CONTEXT_WINDOW_TOKENS
   or extend KNOWN_MODEL_WINDOW_TOKENS) - consider /compact at the next logical boundary
   ```
4. **Interval scales with the window** — `max(60k, 5% of window)` when the window is known.
5. **`formatWindowLabel`** handles `400k`, `1.5M`, and `unknown`.
6. **`getContextRemainingPct`** returns `null` instead of dividing by a window it does not have.

### Before / after

| Window | Threshold before | Threshold after | Note |
| --- | --- | --- | --- |
| 200k | 160k (80%) | 140k (70%) | unchanged in spirit |
| 400k (Opus 4.x) | 250k, window mis-detected as 1M | 280k (70%) | window now detected without env |
| 1M | 250k (25%) | 700k (70%) | ends the premature-nag loop |
| unknown | fabricated % of a guessed window | percentless nudge past 300k | no invented numbers |

## Tests

`node tests/lib/transcript-context.test.js` — 36 → **47 passed, 0 failed**
`node tests/hooks/suggest-compact.test.js` — 44 → **46 passed, 0 failed**
`node tests/lib/llm-summary.test.js` — 18 → **19 passed, 0 failed**
`node tests/hooks/hooks.test.js` — 248 passed, `node tests/integration/hooks.test.js` — 27 passed
(unchanged, re-run as regression guards).

Cases that encoded the old behavior were updated rather than deleted; the #2461/#2468 family cases
stay green. New coverage: percentage thresholds at 200k/400k/1M and for arbitrary sizes, absolute
env override still winning, unknown model → `null` window + 300k threshold, `formatWindowLabel`
unknown/400k/1.5M/2M, interval floor and scaling, the hook's unknown-window message (and its silence
below 300k), and the `getContextRemainingPct` null guard.

## Compatibility

- `COMPACT_CONTEXT_THRESHOLD`, `COMPACT_CONTEXT_INTERVAL`, `ECC_CONTEXT_WINDOW_TOKENS` and
  `CLAUDE_CODE_AUTO_COMPACT_WINDOW` behave exactly as before; `COMPACT_CONTEXT_THRESHOLD=0` still
  disables the signal.
- `DEFAULT_CONTEXT_THRESHOLD_STANDARD` / `DEFAULT_CONTEXT_THRESHOLD_LARGE` remain exported for
  anything importing them.
- `resolveContextInterval(env)` keeps its old one-argument behavior; the window argument is optional.
- Callers of `resolveContextWindowTokens` must now treat `null` as "unknown" — both in-repo callers
  are updated in this PR.
- Both copies of the strategic-compact skill doc (`skills/` and `.agents/skills/`) are updated to
  match, per CONTRIBUTING.

Builds on #2468. Refs #2290.

### DIFF
--- a/.agents/skills/strategic-compact/SKILL.md
+++ b/.agents/skills/strategic-compact/SKILL.md
@@ -31,7 +31,7 @@ Strategic compaction at logical boundaries:
 
 The `suggest-compact.js` script runs on PreToolUse (Edit/Write) and combines two signals:
 
-1. **Context size (primary)** — Reads the latest `usage` record from the session transcript (`transcript_path` in the hook payload) and sums `input_tokens + cache_read_input_tokens + cache_creation_input_tokens` (the true context size of the turn). Suggests `/compact` at a window-scaled threshold — 160k tokens on a 200k window, 250k on a 1M window (detected from a `[1m]` model marker, or inferred when observed tokens already exceed 200k) — and re-reminds after every additional 60k tokens of context growth
+1. **Context size (primary)** — Reads the latest `usage` record from the session transcript (`transcript_path` in the hook payload) and sums `input_tokens + cache_read_input_tokens + cache_creation_input_tokens` (the true context size of the turn). Suggests `/compact` at 70% of the detected context window — 140k tokens on a 200k window, 280k on 400k, 700k on 1M (the window is detected from a `[1m]` model marker, the known-model table, or model-family patterns) — and re-reminds after every additional 60k tokens of context growth. When the model's window cannot be determined, the suggestion reports the raw token count past 300k and says the window is unknown instead of printing a percentage of a guessed window
 2. **Tool-call count (secondary)** — Counts tool invocations in session; suggests at a configurable threshold (default: 50 calls), then every 25 calls after
 
 ## Hook Setup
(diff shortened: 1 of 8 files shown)


haelyra:
Thank you for digging into this, and sorry for the review delay. You identified a real problem: ECC must not present a guessed context-window percentage as fact.

I am closing this version because its implementation also changes compaction policy across eight files, moves the default threshold from the existing bounded values to 70%, and treats every `opus-4` identifier as a 400k window. The model identifier alone does not establish that entitlement or effective window, so that mapping can create a different false percentage. The combined policy and detection changes make this too risky to repair as a small follow-up on this branch.

We are carrying the valid bug fix through #2605, which keeps the existing threshold behavior and omits the percentage whenever the denominator was only assumed. That gives us the honest output without broad policy changes. I appreciate the careful tests and the push toward truthful reporting, and I hope you will keep contributing.

PR #2647, opened 2026-08-01
feat(commands): add /install-ecc-rules command

Or4cu1o:
## Summary
Adds `/install-ecc-rules`, a narrower alternative to `/project-init`: it detects the current project's tech stack, checks existing ECC rule installs (both local `.claude/rules/ecc/` and global `~/.claude/rules/ecc/`), and installs only the matching `rules/<stack>` directories via the manual-copy method already documented in `docs/ECC-RULES_README.md` ("Option 2: Manual Installation") — whole directories, never flattened. It never touches skills, agents, hooks, or platform configs, and never installs the full `rules/` tree, filling a gap for users who already have ECC set up and want just the rule directories relevant to one specific project.

Hardened through live testing across several unrelated projects (not this repo) — each fix below came from a real failure surfaced in practice, not a hypothetical:
- **Non-destructive clone/update**: `/tmp/ECC` is updated in place (`git pull --ff-only`, falling back to `fetch` + `reset --hard` only on diverged history) instead of `rm -rf`'d and re-cloned every run.
- **Sandbox- and proxy-aware failure diagnosis**: distinguishes a sandboxed shell with no network egress, or a configured proxy (`~/.gitconfig` / `HTTP_PROXY` family) pointing at an unreachable host, from a genuine connectivity failure — bypassing the proxy scoped to just the one command invocation, never editing `~/.gitconfig` or any global config.
- **No silent fallback on clone failure**: stops and asks the user (retry / cancel / explicit alternate path) instead of silently substituting an already-installed ECC copy (plugin marketplace cache, `~/.claude/rules/ecc/`, etc.) as the source, which could install a stale or incomplete copy without the user knowing.
- **Global-install awareness**: checks each candidate stack against the global rules scope, not just `common` — so a stack that's already installed globally (e.g. `web`) isn't redundantly recommended for local install, since global rules already apply to every project.

## Type
- [ ] Skill
- [ ] Agent
- [ ] Hook
- [x] Command

## Testing
- Live-tested `/install-ecc-rules` across multiple unrelated real projects (different stacks, different existing ECC install states), iterating on the command instructions after each run surfaced a gap.
- Wired every documented surface for a new command and regenerated the generated ones instead of hand-editing: `agent.yaml` (manual), `COMMANDS-QUICK-REF.md` + `README.md` (+ zh-CN) + `AGENTS.md` (+ zh-CN) counts via `npm run catalog:sync`, `.claude-plugin/{marketplace,plugin}.json` via the same, and `docs/COMMAND-REGISTRY.json` via `npm run command-registry:write`.
- `npm test` (full suite: unicode safety, agents/commands/rules/skills/hooks validation, install-manifest validation, no-personal-paths, catalog:check, command-registry:check, and `tests/run-all.js`) — 3386/3386 passing, 0 failures.
- `docs/COMMAND-AGENT-MAP.md` intentionally left unchanged: like `/project-init`, this command doesn't invoke a primary subagent (it drives Bash/Read/Glob directly), matching the existing pattern for script-driven commands in that table.

## Checklist
- [x] Tested with Claude Code
- [x] Clear descriptions


haelyra:
Thank you for building a narrower path for project-specific ECC rules, and I’m sorry this waited without a maintainer response. A selective rules installer is a useful idea: users should be able to add only the stacks that apply to a project without reinstalling the entire ECC surface.

This version needs a more substantial pass before it can land safely:

- **Build on ECC’s canonical installer contract.** The current workflow clones the latest repository into shared `/tmp/ECC` and copies from there, which bypasses the user’s installed ECC version and the manifest-backed plan/apply safety checks. Success means selecting rule modules from the installed or explicitly chosen ECC source and expressing the result through a deterministic install plan that can be previewed before apply.
- **Make discovery and file operations cross-platform and bounded.** The current command assumes Bash, `/tmp`, `cp -r`, and `rm -rf`, and it can rewrite whole rule directories selected from conversational stack detection. Success means a tested Node implementation using platform-safe paths, root containment, explicit destination validation, and recoverable or atomic replacement behavior.
- **Ship the workflow skills-first.** `commands/` is now ECC’s legacy compatibility surface, while new workflows belong under `skills/`. Success means a canonical skill plus a thin slash-command shim only if compatibility still requires one.
- **Add executable behavior tests.** The current PR documents the interaction but does not test source selection, stack-to-rule mapping, local versus global scope, existing-directory conflicts, cancellation, or partial-failure cleanup. Those cases need deterministic fixtures so the installer cannot silently mix versions or leave a half-applied rules tree.

If you have bandwidth for that redesign, we would be glad to review a focused update. If not, please let us know and we will carry the selective-rules idea into the maintainer task queue after the next release. Thank you again for surfacing the gap and for the thoughtful work you put into the interaction design.

Or4cu1o:
Pushed a full rebuild addressing all 4 points from the review:

1. **Canonical installer contract** — `/install-ecc-rules` no longer clones ECC into `/tmp/ECC` and hand-copies rule directories with `cp -r`. It now goes entirely through `scripts/install-apply.js --modules rules-core`, the same contract `install-plan`, `list-installed`, `doctor`, `repair`, and `uninstall` all rely on. ECC's manifest system has no per-language rule selection (`rules-core` treats `rules/` as one atomic unit — see `docs/SELECTIVE-INSTALL-ARCHITECTURE.md`'s open questions), so stack detection is kept as an informational preview (shown to the user before install) rather than inventing a synthetic per-language module id that install-state/doctor/repair wouldn't understand.

2. **Bounded, cross-platform Node.js implementation** — `scripts/lib/install-rules-selection.js` (stack detection + plan/apply helpers, all mutation delegated via `spawnSync` with an argv array, no shell interpolation) and `scripts/install-rules.js` (thin CLI: `--target`, `--dry-run`, `--json`, `--yes`, interactive y/N confirm). Defaults to the project-scoped `claude-project` target; global (`claude`, `~/.claude`) install requires an explicit flag.

3. **Skills-first shipping** — `skills/install-ecc-rules/SKILL.md` is now the canonical workflow doc, with `commands/install-ecc-rules.md` reduced to a thin compatibility shim pointing at it (mirrors the existing `ecc-guide` command/skill pairing).

4. **Executable tests** — 29 tests across `tests/lib/install-rules-selection.test.js` and `tests/scripts/install-rules.test.js`, including real end-to-end installs against temp project directories (asserting actual files land on disk, install-state gets written, dry-run never touches the filesystem, non-interactive runs refuse without `--yes`, etc).

Also updated the skill/command registries (`agent.yaml`, `package.json`, `manifests/install-modules.json`, `docs/COMMAND-REGISTRY.json`, catalog counts across locale docs) so the new skill is properly registered and discoverable. Full suite is green: `node tests/run-all.js` → 3804/3804 passing, plus `validate-skills.js`, `validate-commands.js`, and `validate-install-manifests.js` all clean.

Ready for another look whenever you have time.

chenhz01:
Two robustness points on the clone management, both adjacent to failure modes the PR already catalogued. First, `/tmp/ECC` is a shared world-writable location keyed to nothing: two projects (or two sessions) running the command concurrently race on the same clone, and the diverged-history fallback (`fetch` + `reset --hard`) silently destroys any local state in that directory. Keying it per user and scope (`/tmp/ecc-rules-${USER}-<short-hash>`, or a bare mirror cache with per-run worktrees) removes both the race and the destruction. Second, nothing records what was fetched — printing the resolved commit SHA (and ideally stashing it next to the installed rules) makes "what rules am I running" auditable after the fact. Also worth double-checking: the marketplace/plugin description bump counts 94 → 95 "legacy command shims", but this PR adds a command, not a shim. Happy to open a PR for the clone keying.

Issue #2574, opened 2026-07-25
cost-tracker: Opus rate is 3x too high and Fable silently falls through to Sonnet pricing


MuzMhz:
## Summary

`scripts/hooks/cost-tracker.js` prices every model through a three-entry table matched by substring. Two problems compound: the Opus rates are stale by 3x, and any model whose name contains neither `haiku` nor `opus` silently receives Sonnet rates. As of July 2026 that includes `claude-fable-5`, which is the most expensive model Anthropic offers.

## Current code

```js
// scripts/hooks/cost-tracker.js:73-83
const RATE_TABLE = {
  haiku:  { in: 0.80,  out: 4.0,  cacheWrite: 1.00,  cacheRead: 0.08 },
  sonnet: { in: 3.00,  out: 15.0, cacheWrite: 3.75,  cacheRead: 0.30 },
  opus:   { in: 15.00, out: 75.0, cacheWrite: 18.75, cacheRead: 1.50 }
};

if (m.includes('haiku')) return RATE_TABLE.haiku;
if (m.includes('opus'))  return RATE_TABLE.opus;
return RATE_TABLE.sonnet;   // <- silent catch-all
```

## Actual pricing (July 2026)

| Model | Table says | Actual | Error |
|---|---|---|---|
| Opus 4.6 / 4.8 / 5 | $15 / $75 | $5 / $25 | 3x overstated |
| Fable 5 | $3 / $15 (via fallback) | $10 / $50 | 3.3x understated |

Opus 5 shipped on 2026-07-24 at the same rate as Opus 4.8, explicitly positioned at half Fable 5's price. Opus has been $5 / $25 for some time; the $15 / $75 figures look like Opus 3 era rates.

Net effect on a mixed-model report: Opus reads about 3x too expensive and Fable about 3.3x too cheap, so Fable appears roughly 10x cheaper relative to Opus than it is. That inverts the routing decision the numbers are meant to inform. I found this while auditing why a local routing policy that rations Fable as the expensive option disagreed with every cost report generated from this hook.

## The structural issue

The stale Opus numbers are a one-line fix, but the catch-all is the part that will keep breaking. `return RATE_TABLE.sonnet` means every unrecognised model is priced as Sonnet with no warning, so a new frontier model is wrong by default until someone notices. Fable is the current casualty; the next model with an unfamiliar name will be the next one.

Two options:

1. Add explicit branches and keep the fallback, but log when it is taken so silent mispricing becomes visible.
[...]

PR #2657, opened 2026-08-03
fix(lib): correct stale opus rate and stop guessing sonnet in cost-estimate

Juanpacol:
## Summary
While looking at #2596 (fix for #2574 in `scripts/hooks/cost-tracker.js`), found the identical bug shape in `scripts/lib/cost-estimate.js`: `RATE_TABLE.opus` still carried the old $15/$75 per 1M token rate (3x the current $5/$25), and `estimateCost()` silently fell back to sonnet rates for any model it didn't recognize (e.g. `claude-fable-5`), mispricing it without any signal.

## Type
- [ ] Skill
- [ ] Agent
- [ ] Hook
- [ ] Command
- [x] Library fix (`scripts/lib/`)

## Changes
- `RATE_TABLE.opus` corrected to `{ in: 5, out: 25 }`
- Added `RATE_TABLE.fable: { in: 10, out: 50 }`
- `estimateCost()` now returns `null` for an unrecognized model instead of guessing sonnet, mirroring the `getRates() -> null` behavior change in #2596
- Updated `tests/lib/cost-estimate.test.js` accordingly (opus/fable rates, null-on-unknown)

Note: this module has no callers outside its own test today (confirmed via `git log --follow` and repo-wide grep), so there's no live behavioral impact — this just corrects the module before anything gets wired up to it.

## Testing
- `node tests/lib/cost-estimate.test.js` — 8/8 passing
- `node tests/run-all.js` — 3385/3385 passing, no regressions
- Manual repro:
  ```
  opus 1M/1M -> 30
  claude-fable-5 1M/1M -> 60
  unknown 1M/1M -> null
  ```

## Checklist
- [x] Tested with Claude Code
- [x] Clear descriptions

### DIFF
--- a/scripts/lib/cost-estimate.js
+++ b/scripts/lib/cost-estimate.js
@@ -9,21 +9,26 @@
 const RATE_TABLE = {
   haiku: { in: 0.8, out: 4.0 },
   sonnet: { in: 3.0, out: 15.0 },
-  opus: { in: 15.0, out: 75.0 }
+  opus: { in: 5.0, out: 25.0 },
+  fable: { in: 10.0, out: 50.0 }
 };
 
 /**
  * Estimate USD cost from token counts.
- * @param {string} model - Model name (may contain "haiku", "sonnet", or "opus")
+ * @param {string} model - Model name (may contain "haiku", "sonnet", "opus", or "fable")
  * @param {number} inputTokens
  * @param {number} outputTokens
- * @returns {number} Estimated cost in USD (rounded to 6 decimal places)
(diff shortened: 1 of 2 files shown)


haelyra:
Thank you for catching both the stale pricing and the danger of silently guessing a Sonnet rate for an unknown model. That unknown-must-remain-unknown contract is especially useful.

Since this opened, the live cost-tracking work has consolidated in #2691 and the shared-estimator follow-up in #2693. The generic Opus match here would price every Opus generation at the new rate, including legacy models with different pricing, while this module currently has no production callers. We are going to close this partial path rather than add a second divergent rate table. We appreciate the careful diagnosis and the contribution.

PR #2663, opened 2026-08-03
test(lib): extract shared mini test runner for coordination tests

t0kubetsu:
## Summary

Follow-up to review feedback on #2311 (CodeRabbit), split out as a focused PR so #2311 stays a skill-only diff.

- Adds `tests/lib/helpers/mini-test-runner.js` — shared `test`/`testAsync`/`banner`/`section`/`summary`/`fatal` helpers for the standalone test scripts
- Migrates the three `tests/lib/github-coordination-*.test.js` files onto it, removing three duplicated `test(name, fn)` harnesses and routing all reporter output through one place

No behavior change: same output format, same exit codes.

## Testing

- `node tests/lib/github-coordination-branches.test.js` — 35 passed
- `node tests/lib/github-coordination-policy.test.js` — 28 passed
- `node tests/lib/github-coordination-store.test.js` — 23 passed
- `npx eslint` clean on all four files

### DIFF
--- a/tests/lib/github-coordination-branches.test.js
+++ b/tests/lib/github-coordination-branches.test.js
@@ -25,24 +25,14 @@ const {
   verifyDependenciesClosed,
 } = require('../../scripts/lib/github-coordination/state');
 
-function test(name, fn) {
-  try {
-    fn();
-    console.log(`  ✓ ${name}`);
-    return true;
-  } catch (err) {
-    console.log(`  ✗ ${name}`);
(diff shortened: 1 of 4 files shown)


haelyra:
Thank you for splitting this into a focused test-only cleanup. The shared mini runner removes duplicated harness code while preserving the three suites byte-for-byte in normalized output and exit behavior; all 86 focused cases and the full current matrix are green. This is ready to merge, and the non-Error reporting polish can remain a follow-up.

PR #2666, opened 2026-08-03
fix(hooks): stop block-no-verify from firing on flags it does not own

Peopleoftech:
## Problem

`block-no-verify` scans the raw command string after the `commit` keyword, so it
blocks flags that do not belong to git. Two ways this bites in normal use:

**1. A downstream tool's `-n`.** Everything after `commit` is inspected,
including whatever follows a pipe:

```sh
git commit -m "msg" | head -n 5      # BLOCKED — head's -n read as git's
git commit -m "msg" && … | tail -n 4 # BLOCKED
git commit -m "msg" && grep -n TODO  # BLOCKED
```

In practice this means the more output an agent asks for, the likelier a
spurious block — the opposite of what you want from a guardrail.

**2. Quoted text.** Quoting is invisible to a regex, so a commit message that
merely *mentions* a flag trips the hook on its own documentation:

```sh
git commit -m "example: head -n 5"                     # BLOCKED
git commit -m "never pass --no-verify"                 # BLOCKED
git commit -m "do not use -c core.hooksPath=/dev/null" # BLOCKED
```

I hit the second one while committing the fix for the first.

## Fix

Replace the regex scan with a quote-aware tokenizer.

Quoted spans collapse into a single token — exactly what the shell hands to git
— and each token records which pipe/chain `segment` it came from, so git's own
arguments can be told apart from a later command's. A token consumed as the
value of a mandatory-value flag is data, never a flag.

`hasHooksPathOverride` is token-based for the same reason.

## Detection is unchanged where it matters

| Case | Behaviour |
|---|---|
| `git commit --no-verify` | blocked |
| `git commit -n` | blocked |
| `git commit -an` (clustered) | blocked |
| `git commit "--no-verify"` (quoted flag) | blocked — git receives it either way |
| `git commit -m x && git push --no-verify` | blocked — long form is not scoped |
| `git -c core.hooksPath=/dev/null commit` | blocked |

The long form deliberately stays unscoped: it is unambiguous, so a bypass
anywhere downstream in a chain should still trip.

## One subtlety worth reviewing

`VALUE_TAKING_FLAGS` lists **only** options whose value is mandatory.

An optional-value option such as `-S` or `-u` does not swallow the next token,
so listing one would let `git commit -S -n` through — a real bypass introduced
by a false-positive fix. There is a test pinning that.

## Tests

Suite goes 12 → 22. The 10 new cases cover both false-positive classes and the
bypasses that must keep tripping.

I verified the new tests fail against the previous implementation and pass
after, so they pin real behaviour rather than describing the new code back to
itself.

```
Passed: 22  Failed: 0
```

### DIFF
--- a/scripts/hooks/block-no-verify.js
+++ b/scripts/hooks/block-no-verify.js
@@ -150,18 +150,103 @@ function detectGitCommand(input) {
 }
 
 /**
- * Check if the input contains a --no-verify flag for a specific git command.
+ * Flags whose FOLLOWING token git consumes as a value rather than a flag.
+ *
+ * Only options whose value is mandatory belong here. An optional-value option
+ * such as -S or -u does not swallow the next token, so listing one would let
+ * `git commit -S -n` slip through.
+ */
+const VALUE_TAKING_FLAGS = new Set([
+  '-m', '--message',
+  '-F', '--file',
+  '-c', '--reedit-message',
+  '-C', '--reuse-message',
+  '-t', '--template',
+  '--author',
+  '--date',
+  '--cleanup',
+  '--trailer',
+  '--fixup',
+  '--squash',
+  '--pathspec-from-file',
+]);
+
+/**
+ * Split a command string into shell-like tokens.
+ *
+ * Quote-aware by design: without it, text inside a quoted argument cannot be
+ * told apart from a real flag, so `git commit -m "head -n 5"` — a message that
+ * merely mentions a flag — was blocked as if the flag had been passed. Quoted
+ * spans collapse into one token, exactly as the shell hands them to git.
+ *
+ * Each token carries a `segment` index that increments at every unquoted shell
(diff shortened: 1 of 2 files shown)


Peopleoftech:
Closing in favour of #2668.

#2405 landed the same fixes independently while this was open — quoted commit messages, piped commands, and the optional-value boundary all behave correctly on current main. I re-ran this PR's cases against main and 9 of 10 pass, so keeping this open would mostly duplicate work already done, and better.

The one case still failing is a real bypass rather than a false positive: short options cluster, so `git commit -an` is `-a` plus the bypass flag and skips the hooks. `-sn` and `-vn` do the same, while `-na` and `-nm` are caught — the check anchors on the first character.

#2668 fixes just that, against current main, with tests verified to fail before and pass after. Thanks for the quick work on #2405.

PR #2669, opened 2026-08-03
chore(ci): add AgentShield review workflow

reevesc88:
## Summary
- add a GitHub Actions workflow for AgentShield review scanning
- pin the action to affaan-m/agentshield@v1.4.0
- scan the repository on push and pull_request with min-severity medium

### DIFF
--- a/.github/workflows/agentshield-security.yml
+++ b/.github/workflows/agentshield-security.yml
@@ -1,0 +1,22 @@
+name: AgentShield Security Scan
+
+on:
+  pull_request:
+  push:
+    branches: [main, master]
+
+permissions:
+  contents: read
+
+jobs:
+  agentshield:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: AgentShield Security Scan
+        uses: affaan-m/agentshield@v1.4.0
+        with:
+          path: "."
+          min-severity: "medium"
+          fail-on-findings: "true"


haelyra:
Thanks for wiring this up — wanting AgentShield in CI is right, and the workflow itself is well-formed (least-privilege `contents: read`, action pinned to a release tag).

I'm closing this because the action it calls is currently broken, so merging would add a permanently-red check to every PR. `affaan-m/agentshield@v1.4.0` fails before it scans anything:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'yaml' imported from
  /home/runner/work/_actions/affaan-m/agentshield/v1.4.0/dist/action.js
Node.js v24.18.0
```

The published `dist/action.js` imports `yaml` but the release doesn't bundle it and there's no `node_modules` alongside it — so the step exits non-zero on any repo. With `fail-on-findings: true` that reads as a security failure on every PR, which is the worst version of a red check: it trains people to ignore it.

You can see it fail on #2665, which carries the same workflow file: https://github.com/affaan-m/ECC/actions/runs/30848605181

This is a packaging bug in agentshield's release, not in your workflow, and it needs fixing there — bundling `yaml` into `dist/` or shipping the dependency with the action. Once a fixed version is tagged, this workflow is worth re-opening essentially as written. Worth considering `fail-on-findings: false` for a first landing too, so the initial signal is visible without gating merges while the baseline gets triaged.

Closing as blocked on the upstream action rather than on the idea. Please do re-send it when there's a working release.

PR #2670, opened 2026-08-04
fix(mcp): accept reserved _meta field in tools/call params

agent-jeong:
## Problem

The memory MCP server rejects any `tools/call` whose `params` contains a key
other than `name`/`arguments`, returning `-32602 "Unknown or missing memory tool."`
— even for tools that are correctly registered.

MCP clients (e.g. Claude Code) attach the spec-reserved `_meta` field
(such as `progressToken`) to request params. As a result **every** tool call
from a compliant client fails and the entire memory MCP surface becomes
unreachable, while `initialize` / `tools/list` and the `ecc memory` CLI keep
working (they don't hit this check). This makes the failure look like a broken
connection when the server is actually connected.

## Root cause

`scripts/memory-mcp.mjs`, the `tools/call` params allowlist:

```js
|| Object.keys(params).some(key => !['name', 'arguments'].includes(key))
```

`_meta` is not in the allowlist, so its presence trips the guard before the
tool ever runs.

## Fix

Add `_meta` to the allowlist. Per the MCP base protocol, `_meta` is reserved
for request metadata and must be accepted by servers.

## Reproduction (against pristine `main`)

Calling `createMemoryMcpService().handle(...)` directly:

| `params` | result |
|----------|--------|
| `{ name, arguments }` | ✅ OK |
| `{ name, arguments, _meta: { progressToken } }` | ❌ `-32602` |

After this change, both return OK.

## Spec reference

MCP 2025-06-18 basic protocol, General fields → `_meta`:
> "The `_meta` property/parameter is reserved by MCP to allow clients and
> servers to attach additional metadata to their interactions."

### DIFF
--- a/scripts/memory-mcp.mjs
+++ b/scripts/memory-mcp.mjs
@@ -437,7 +437,8 @@ function createMemoryMcpService(options = {}) {
           !isRecord(params)
           || typeof name !== 'string'
           || !TOOL_BY_NAME.has(name)
-          || Object.keys(params).some(key => !['name', 'arguments'].includes(key))
+          // `_meta` is reserved by MCP for request metadata (e.g. progressToken) and must be accepted.
+          || Object.keys(params).some(key => !['name', 'arguments', '_meta'].includes(key))
         ) {
           return jsonRpcError(message.id, -32602, 'Unknown or missing memory tool.');
         }
--- a/scripts/memory-mcp.mjs
+++ b/scripts/memory-mcp.mjs
@@ -437,7 +437,9 @@ function createMemoryMcpService(options = {}) {
           !isRecord(params)
           || typeof name !== 'string'
           || !TOOL_BY_NAME.has(name)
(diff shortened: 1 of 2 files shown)


agent-jeong:
Thanks @coderabbitai — both points addressed in 7119499:

- `params._meta` is now shape-validated when present: metadata objects are accepted, while `null`, arrays, and scalars are rejected (reusing `isRecord`). `_meta` stays optional and the existing name/arguments/unexpected-key checks are preserved.
- Added regression tests in `tests/scripts/memory-mcp.test.js`: accept `_meta` with a `progressToken`, reject malformed `_meta` values, and continue rejecting unrelated top-level params.

Full memory-mcp suite passes locally (12/12).

haelyra:
Thank you for the careful protocol fix. Accepting the reserved _meta object restores protocol-compliant tool calls while the exact top-level allowlist, strict tool schemas, transport bounds, and malformed-value rejection all remain fail-closed. The focused MCP suite passes 12/12, and this is ready to merge.

PR #2672, opened 2026-08-04
fix(deps): bump fast-uri to 3.1.5 and brace-expansion to 5.0.9 (npm audit)

t0kubetsu:
## Summary

`npm audit --audit-level=high` in the **Security Scan** job currently fails on every PR (and on `main`, e.g. run 30877829542) due to two freshly published high-severity advisories:

| Package | Advisory | Vulnerable | Patched | Fix applied |
|---|---|---|---|---|
| `fast-uri` | [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) — host confusion via backslash authority introducer | >=3.0.0 <3.1.5 | 3.1.5 | bump the existing `overrides`/`resolutions` pin 3.1.4 → 3.1.5 (still satisfies ajv's `^3.0.1`) |
| `brace-expansion` | [GHSA-rgw5-rvv9-x895](https://github.com/advisories/GHSA-rgw5-rvv9-x895) — DoS via unbounded intermediate arrays | >=4.0.0 <5.0.9 | 5.0.9 | in-range lockfile bump 5.0.8 → 5.0.9 (under `eslint` → `minimatch`) |

After this change `npm audit --audit-level=high` reports **0 vulnerabilities**.

## Details

- `fast-uri` was pinned at 3.1.4 in both `overrides` (npm/pnpm/bun) and `resolutions` (yarn); both pins bumped to 3.1.5.
- `yarn.lock` regenerated with Yarn 4.9.2 (`yarn install` + `yarn up -R brace-expansion`) so the yarn matrix cells stay in sync — the diff is 7 lines each way.
- No source changes; `node tests/run-all.js` passes locally (3407/3408 — the single failure is a pre-existing spawnSync `maxBuffer` artifact of my long local worktree path, unrelated to this change; it passes from a normal-length path).

## Why a separate PR

Unblocks Security Scan for all open PRs (it is currently red on `main` itself), kept as a focused deps-only diff.

PR #2676, opened 2026-08-04
fix: bump fast-uri override to 3.1.5 to resolve high-severity vuln

oswroso:
## Summary
- The `overrides`/`resolutions` pin for `fast-uri` was fixed at `3.1.4`, which sits inside the vulnerable range (3.0.0-3.1.4) for [GHSA-7p8r-x3mc-p8w7](https://github.com/advisories/GHSA-7p8r-x3mc-p8w7) (host confusion via backslash authority introducer)
- Because the override pinned the exact vulnerable version, `npm audit fix` (even with `--force`) could never move it
- Bumped the override to `3.1.5` and regenerated the lockfile

## Test plan
- [x] `npm install`
- [x] `npm audit` reports 0 vulnerabilities


haelyra:
Thanks for catching the `fast-uri` advisory — the diagnosis was right, and the fix has landed in #2672 (merged as 2665d48a), so I'm closing this as a duplicate.

For transparency on why #2672 was the one taken: both PRs make the identical `package.json` change (`overrides`/`resolutions` 3.1.4 → 3.1.5), but the lockfiles differ substantially.

- #2672 is a targeted update: `+15/-15` across the three files, and it also picks up `brace-expansion` 5.0.9, which closes a second high-severity advisory (GHSA-rgw5-rvv9-x895) that was failing the same Security Scan job.
- This PR regenerates `yarn.lock` wholesale (`+1455/-2169`). That rewrite drops the Yarn Berry `__metadata` header and `cacheKey` and re-emits the file in Classic format, which would have reverted the lockfile format this repo is on.

The wholesale rewrite is almost always a local Yarn version mismatch rather than anything intentional — if you run `corepack enable` and use the version pinned in `packageManager`, targeted dependency bumps will stay targeted.

The advisory you reported is fixed and `npm audit` is clean on main now. Appreciate the report, and please do send more — a follow-up with just the `package.json` line and a minimally-updated lockfile is always welcome.

PR #2679, opened 2026-08-04
feat: add OpenSpec validators, spec-miner registration, real fixtures

legeZZZ:
# OpenSpec validator foundation (PR1)

Focused revision per review. Standalone **read-only** validators only — CI gating is intentionally **not** wired in this PR (follow-up once shallow-clone and exit semantics are stable). No README or unrelated changes.

## `validate-openspec-syntax.js` — v1 schema contract
- Documented in the script header; metadata keys are an **explicit allowlist** with a typo guard; ordinary prose comments are allowed
- Delta files validated **per block**: every declared `ADDED`/`MODIFIED`/`REMOVED` block must be non-empty (an empty first block no longer passes because a later block has content)
- `<!-- enforced: path/to/file.ext::symbol -->` anchors: repo-relative, reject whitespace / leading `/` / `..` traversal / backslashes
- YAML frontmatter optional

## `check-spec-freshness.js` — file-level freshness
- git invoked only via `execFileSync` argv arrays; paths after `--`, verified date validated then passed as a single argument — nothing from spec Markdown reaches a shell (hostile filename + date fixtures prove it)
- Project root and every enforced target resolved through realpaths and contained in the project; symlinked root / escaping target rejected
- Shallow clones → `UNVERIFIED` (never ORPHANED/STALE); truthful verdicts FRESH / STALE / ORPHANED (missing/escaping target) / UNVERIFIED
- Contract explicitly **file-level** (the `::symbol` is recorded, not resolved)
- `ECC_SPEC_STALE_WARN_ONLY` implemented so warning mode matches the exit contract; `ECC_SPEC_STALE_DAYS` validated as strict integer 1–365

## Fixtures & tests
- Both suites rebuilt around real, git-backed repos plus hostile-input, shallow-clone, path-escape, per-block-delta and metadata cases (27 + 23 tests)
- Removed the old unused on-disk fixture tree (tests construct real fixtures inline)

## Manifest
- `manifests/install-components.json`: register existing spec-miner agent

> Open question for maintainers (see PR thread): spec-miner currently emits `enforced: FileName.methodName()` anchors without a file path; the file-level freshness contract needs a repo-relative path. Whether to align spec-miner in this PR or as a follow-up.

### DIFF
--- a/manifests/install-components.json
+++ b/manifests/install-components.json
@@ -399,6 +399,14 @@
       ]
     },
     {
+      "id": "agent:spec-miner",
+      "family": "agent",
+      "description": "Behavioral spec extraction agent for brownfield codebases.",
+      "modules": [
+        "agents-core"
+      ]
+    },
+    {
       "id": "agent:tdd-guide",
       "family": "agent",
       "description": "Test-driven development guidance agent.",
--- a/scripts/ci/check-spec-freshness.js
+++ b/scripts/ci/check-spec-freshness.js
@@ -1,0 +1,411 @@
+#!/usr/bin/env node
+/**
+ * Check OpenSpec freshness at FILE level.
+ *
+ * ── v1 contract ─────────────────────────────────────────────────
+ * For each spec file under <root>/openspec/, parse its `<!-- enforced: -->`
+ * anchors and the `Last verified:` date, then ask git whether the anchor's
+ * TARGET FILE changed after that date. This is a FILE-LEVEL freshness check:
+ * the `::symbol` part of an anchor is recorded but never resolved — verifying
+ * that a symbol exists inside a file is out of scope for v1.
+ *
+ * Truthful verdicts (a spec resolves to the strongest of these):
+ *   FRESH       every enforced target exists, is inside the project, is
+ *               committed, and has not changed since `Last verified`
+ *   STALE       an enforced target changed after `Last verified`, or the
+ *               verification itself is older than ECC_SPEC_STALE_DAYS
+ *   ORPHANED    an enforced target file is missing from the tree or escapes
+ *               the project (via `..` or a symlink out of the root)
+ *   UNVERIFIED  no git / not a repo / SHALLOW clone / missing or invalid
+ *               `Last verified` date / target file never committed — i.e.
+ *               evidence is unavailable, NOT a sign of staleness
+ *   UNKNOWN     an enforced anchor does not parse to path::symbol
+ *
+ * Safety:
+ *   - git is always invoked through execFileSync with an argv array. Paths are
+ *     passed after `--`; the date is validated to YYYY-MM-DD first and passed
+ *     as a single argument. Nothing from spec Markdown reaches a shell.
+ *   - the project root must not be a symlink; every enforced target is resolved
+ *     through fs.realpathSync and must stay inside the real project root.
+ *   - shallow clones report UNVERIFIED — incomplete history is never
+ *     mislabelled ORPHANED or STALE.
+ *
+ * Exit contract:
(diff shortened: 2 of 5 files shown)


haelyra:
Thank you for rebuilding this as a focused foundation. Machine-checkable specs, real git-backed fixtures, and a freshness verdict that distinguishes stale from unverifiable evidence are valuable for ECC, especially as an optional input to the post-2.2 M3 evidence contract. I’m sorry this waited so long for a maintainer pass.

We would love another focused revision with these changes:

- **Remove shell construction from freshness checks.** `lastVerifiedDate` and enforced file paths come from repository-controlled Markdown and are interpolated into `execSync(...)` commands. That lets quotes and shell syntax escape the intended arguments in CI. Invoke Git with an argv array through `spawnSync` or `execFileSync`, pass paths after `--`, and add hostile date and filename fixtures that prove no command can escape.
- **Make shallow history and warning mode truthful.** The current script never checks `git rev-parse --is-shallow-repository`, so missing history can be reported as `ORPHANED`, and the workflow sets `ECC_SPEC_STALE_WARN_ONLY=true` even though the script never reads it, so stale specs still fail validation. Return `UNVERIFIED` when history is incomplete and implement or remove the warning mode so the workflow matches the documented exit contract.
- **Validate each delta block and metadata field against one explicit schema.** Today content in any later block makes all earlier empty ADDED/MODIFIED/REMOVED blocks pass, while every colon-bearing HTML comment is treated as metadata. Parse blocks individually, reject empty declared blocks, allow ordinary comments, and use a documented allowlist plus required fields for spec metadata.
- **Bind anchors to a contained, verifiable target.** The parser records `path::symbol`, but freshness checks only file history and never establish that the symbol exists. Resolve the OpenSpec root and every referenced file through real paths inside the project, reject a symlinked root or escaping target, and either verify the symbol or name this first contract as file-level freshness only.

Please keep this as an optional evidence source rather than a second canonical task or completion authority. A smaller follow-up containing the schema, safe parser, fixtures, and read-only validator would be easiest to land first; CI gating can follow once its shallow-clone and exit semantics are stable. If you do not have bandwidth for that pass, we can add the foundation to the post-release maintainer queue and carry it forward with attribution.

legeZZZ:
Revised and pushed.

Per your feedback: validators are now standalone read‑only tooling; reusable‑validate.yml CI steps have been removed from this PR.
All four requested changes are complete:
1. Git invoked via argv array only, plus hostile‑input fixtures
2. Shallow‑repository case returns UNVERIFIED; ECC_SPEC_STALE_WARN_ONLY fully implemented
3. Per‑block delta validation with documented metadata allowlist
4. Targets constrained via realpath containment; freshness is explicitly file‑level

The diff scope is limited to the two validators, associated tests, and spec‑miner registration.

Open question: spec‑miner currently outputs `enforced: FileName.methodName()` without a file path, but our file‑level freshness check requires a repo‑relative path.
Shall we fix spec‑miner in this PR, or defer to a subsequent change?

chenhz01:
To the existing review, one over-constraint case I don't see raised: `collectEnforcedAnchors` documents itself as "every enforced anchor string in the file (any location)" — the global `ENFORCED_RE` sweep has no fence awareness, so a spec that merely *shows* the anchor syntax inside an example code block gets collected and validated as if the spec declared it. Worst case is silent over-constraint: the example's target file starts failing freshness for a spec that never enforced it, or an illustrative-but-invalid anchor throws a grammar error on a file that is otherwise fine. Cheap fix: strip fenced code blocks (and inline code spans) from the body before the anchor sweep — the validator already separates frontmatter from body, so it slots in right before `collectEnforcedAnchors`. Happy to open a PR with the fence-stripping plus a fixture if useful.

PR #2691, opened 2026-08-06
fix(hooks): correct stale model rates in the cost tracker

andrest:
## What Changed

`scripts/hooks/cost-tracker.js` priced every `opus` model at **$15/$75** per Mtok. Those are Claude 3 Opus era rates. Opus 4.5 and later bill at **$5/$25**, so every current-generation Opus session has been reported at exactly **3x real spend**.

**The arithmetic in the hook was always correct. Only the rate constants were wrong.**

Concrete, measured: a session whose transcript summed to **100,000,000 cache-read + 500,000 output tokens** on `claude-opus-5`.

| | Reported |
|---|---|
| Stale table | **$187.50** |
| Corrected table | **$62.50** |

That case is now test 14, so the number in this table is machine-checked rather than asserted.

**This affects every user running any Opus model.** Opus is ECC's default for the heavier agents, so for many users it is most of the bill.

### The rate table

| Bucket | Was | Now |
|---|---|---|
| `opus` (4.5, 4.6, 4.7, 4.8, Opus 5) | $15.00 / $75.00 | **$5.00 / $25.00** |
| `opusLegacy` (Claude 3 Opus, Opus 4.0, Opus 4.1) | (same bucket as above) | **$15.00 / $75.00** (new row) |
| `fable` (Fable 5, Mythos 5) | no bucket, fell through to `sonnet` | **$10.00 / $50.00** (new row) |
| `haiku` (Haiku 4.5) | $0.80 / $4.00 | **$1.00 / $5.00** |
| `haikuLegacy` (Claude 3.5 Haiku) | (same bucket as above) | **$0.80 / $4.00** (new row) |
| `sonnet` (Sonnet 5, Sonnet 4.6) | $3.00 / $15.00 | unchanged |

Cache write is 1.25x input and cache read is 0.1x input in every row, pinned by a test so the hand-typed cache columns cannot drift.

The two legacy rows exist so that correcting the current generation does not reprice the old one. Claude 3.5 Haiku's $0.80/$4.00 was *exactly right*; collapsing all Haiku to $1/$5 would have traded one error for another. Claude 3 Haiku ($0.25/$1.25) is deliberately not modelled, and the comment says so.

A bare `includes('opus')` cannot separate the two Opus eras, so legacy IDs are matched explicitly. Every spelling had to be covered:

- aliases: `claude-opus-4-0`, `claude-opus-4-1`
- dated snapshots: `claude-opus-4-20250514` (Opus 4.0 has **no minor segment**, so an `opus-4-0` substring alone misses it), `claude-opus-4-1-20250805`
- Bedrock: `us.anthropic.claude-3-opus-20240229-v1:0`
- Vertex: `claude-opus-4@20250514` (joins the date with `@`)

and the negative side is tested too: `claude-opus-4-5-20251101`, `claude-opus-4-8[1m]` and `claude-opus-4-5@20251101` must **not** land in the legacy bucket.

### Unknown models: the root cause

The table had no way to say "I do not know". Any unrecognized model got a plausible sonnet-priced number with **no signal at all**, which is precisely how the stale Opus rate survived unnoticed. Two changes, both non-blocking:

1. Rows carry a new **`rate_bucket`** field naming what produced the cost. `sonnet-fallback` means the model ID matched nothing.
2. An unrecognized model that actually priced tokens writes **one line to stderr** naming it.

`rate_bucket` follows the *cost source*, not the model, so it reads `harness` when the statusline's authoritative `cost.total_cost_usd` supplied the number and no rate-table row was consulted. Without that, the field would vouch for harness rows as table-priced and, worse, flag an authoritative number as a guess.

The hook stays **fail-open throughout**: an unknown model still gets a usable default, nothing throws, and stdout pass-through is untouched. Per the file's own contract, this hook never fails the Stop event.

### One adjacent fix the new bucket exposed

Claude Code writes `{"type":"assistant","message":{"model":"<synthetic>","usage":{...}}}` for interrupts, API errors, and "no response requested". It carries a **fully populated but all-zero** `usage` block, so it cleared the existing usage guard and then, under last-model-wins, renamed the whole session.

Measured over **1,628 local transcripts**: 60 such entries, and **10 transcripts end on one**. That last case is the damaging one, because the final row is exactly the cumulative row `/cost-report` reads per session. An Opus session ending on an interrupt was reported at $18 instead of $30 in the reproduction, and it also produced a useless `unrecognized model "<synthetic>"` warning from the new code. `<synthetic>` now joins `unknown` as a sentinel that does not claim the session.

I would happily split this out if you would rather keep the PR to rates alone, but it is what makes the new fallback signal mean anything in practice.

## Why This Change

Cost tracking that is confidently wrong is worse than no cost tracking. A 3x overstatement on the most expensive model pushes users toward exactly the wrong optimization, and `skills/cost-tracking/SKILL.md` tells readers to trust `estimated_cost_usd` over hand-calculating.

The deeper problem was that a hard-coded table with a silent default has no failure mode. Every future model repricing produces a plausible-but-wrong number and nothing ever surfaces it. `rate_bucket` plus the warning is the smallest change that gives it one.

## Testing Done

**14 new cases** in `tests/hooks/cost-tracker.test.js`, matching the file's existing spawn-the-hook-and-read-the-row style and its runner. Each asserts the **dollar figure**, not merely that one is present: a "cost > 0" assertion is exactly what let the stale rate survive.

```
$ node tests/hooks/cost-tracker.test.js

=== Testing cost-tracker.js ===

  PASS passes through input on stdout
  PASS creates metrics file when given transcript usage data
  PASS counts usage once per message.id across multi-line responses
  PASS handles empty input gracefully
  PASS handles invalid JSON gracefully
  PASS handles missing usage fields gracefully
  PASS prefers ECC_SESSION_ID over CLAUDE_SESSION_ID when both are present
  PASS uses input session_id for session correlation when env vars are absent
  PASS prefers fresh harness-cost cache over transcript estimate
  PASS ignores stale harness-cost cache (>300s) and uses transcript estimate
  PASS prices current Opus models at $5/$25 per Mtok
  PASS keeps $15/$75 for legacy Opus IDs (3 Opus, 4.0, 4.1)
  PASS prices Haiku 4.5 at $1/$5 per Mtok
  PASS prices Fable and Mythos at $10/$50 per Mtok
  PASS prices a cache-heavy Opus 5 session at $62.50, not the old $187.50
  PASS flags an unrecognized model instead of silently using sonnet rates
  PASS stays quiet when no assistant turn recorded a model
  PASS marks an unattributed session as a fallback without warning
  PASS prices Sonnet at $3/$15 and marks it as a real match
  PASS keeps $0.80/$4.00 for Claude 3.5 Haiku
  PASS keeps cache write at 1.25x and cache read at 0.1x input in every bucket
  PASS ignores the <synthetic> sentinel when attributing the session model
  PASS records rate_bucket as harness when the harness cost wins
  PASS does not flag an unrecognized model when the harness supplied the cost

Results: Passed: 24, Failed: 0
```

**Verified red before green.** With the new tests in place and only `scripts/hooks/cost-tracker.js` reverted to `main`, 14 of the 24 fail, and the headline case fails with the exact regression:

```
  FAIL prices a cache-heavy Opus 5 session at $62.50, not the old $187.50
    Error: Expected $62.50, got 187.5
    187.5 !== 62.5
```

Full gauntlet, `npm test`: **3456 tests, 3445 passed, 11 failed.** Those 11 all live in `hooks/gateguard*` and `hooks/suggest-compact` and are **pre-existing on a clean `main`**, which I confirmed by stashing this change and re-running (clean `main`: 3442 tests, 3431 passed, **the same 11 failed**). Nothing here touches those paths.

`eslint`, `markdownlint` and `node scripts/ci/check-unicode-safety.js` all clean on the changed files.

The change was also put through an adversarial review pass before opening. Findings on the Vertex `@date` spelling, the `rate_bucket` harness path, the Haiku 3.5 regression, the `<synthetic>` sentinel, and several overstated code comments are all incorporated above.

## Type of Change
- [x] `fix:` Bug fix
- [ ] `feat:` New feature
- [ ] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `ci:` CI/CD changes

- [x] JSON files validate cleanly (no JSON files changed)
- [x] Shell scripts pass shellcheck (if applicable) (no shell scripts changed)
- [x] No sensitive data exposed in logs or output. The new stderr line prints only the model ID from the transcript.

- [x] Not applicable. No dependency, `package.json` or `yarn.lock` change.

- [x] Not applicable. This modifies an existing hook. No new surface, so no manifest, catalog, command-registry or publish-allowlist change is needed. `npm run catalog:check` and `npm run command-registry:check` pass as part of the `npm test` run above.

## Documentation
- [x] Updated relevant documentation. The row schema is documented in two places and both now list `rate_bucket`: `commands/cost-report.md` (schema line, plus the CSV export columns so the field is actually reachable) and `skills/cost-tracking/SKILL.md` (schema table).
- [x] Added comments for complex logic. The regex alternatives and the fallback gating each carry a *why*, including the Opus 4.0 snapshot case and the reason the warning is silent on the harness path.
- [ ] README updated (if needed). Not needed, the README does not document this row schema.

## Noticed but deliberately left out of scope

Two places carry the same stale numbers. Both are separable and neither is in this hook's path, so I left them rather than sprawl the diff. Happy to send either as a follow-up, or to fold them in here if you would prefer one PR.

1. **`scripts/lib/cost-estimate.js`** has an independent `RATE_TABLE` with the identical `opus: { in: 15.0, out: 75.0 }` and `haiku: { in: 0.8, out: 4.0 }`. Nothing currently requires it except its own test (`tests/lib/cost-estimate.test.js`), which pins the values, so fixing it means changing that test too.
2. **`skills/cost-aware-llm-pipeline/SKILL.md`** (and the `docs/zh-CN` and `docs/ja-JP` translations of it) has a pricing table listing "Opus 4.5 | $15.00 | $75.00", which is the same error in prose. That is a docs change across three locales.

### DIFF
--- a/commands/cost-report.md
+++ b/commands/cost-report.md
@@ -16,7 +16,13 @@ session**, so the report takes the **latest row per `session_id`** and sums
 across sessions (summing every row would multiply-count).
 
 Row schema:
-`{ timestamp, session_id, transcript_path, model, input_tokens, output_tokens, cache_write_tokens, cache_read_tokens, estimated_cost_usd }`
+`{ timestamp, session_id, transcript_path, model, rate_bucket, input_tokens, output_tokens, cache_write_tokens, cache_read_tokens, estimated_cost_usd }`
+
+`rate_bucket` names what produced `estimated_cost_usd`: a rate-table row
+(`haiku`, `haiku-legacy`, `sonnet`, `opus`, `opus-legacy`, `fable`), or
+`harness` when the statusline's authoritative `cost.total_cost_usd` supplied
+it. `sonnet-fallback` means the model ID matched no row and sonnet rates were
+assumed, so that row's cost is a guess worth flagging.
 
 ## What this command does
 
@@ -66,8 +72,8 @@ const fs=require("fs"),os=require("os"),path=require("path");
 const f=path.join(os.homedir(),".claude","metrics","costs.jsonl");
 if(!fs.existsSync(f)){console.error("no data");process.exit(0);}
 const rows=fs.readFileSync(f,"utf8").split(/\r?\n/).filter(Boolean).map(l=>{try{return JSON.parse(l)}catch{return null}}).filter(Boolean).slice(-100);
-console.log("timestamp,session_id,model,input_tokens,output_tokens,cache_write_tokens,cache_read_tokens,estimated_cost_usd");
-for(const r of rows)console.log([r.timestamp,r.session_id,r.model,r.input_tokens,r.output_tokens,r.cache_write_tokens,r.cache_read_tokens,r.estimated_cost_usd].join(","));
+console.log("timestamp,session_id,model,rate_bucket,input_tokens,output_tokens,cache_write_tokens,cache_read_tokens,estimated_cost_usd");
+for(const r of rows)console.log([r.timestamp,r.session_id,r.model,r.rate_bucket,r.input_tokens,r.output_tokens,r.cache_write_tokens,r.cache_read_tokens,r.estimated_cost_usd].join(","));
 '
 ```
 
--- a/commands/cost-report.md
+++ b/commands/cost-report.md
@@ -73,7 +73,8 @@ const f=path.join(os.homedir(),".claude","metrics","costs.jsonl");
 if(!fs.existsSync(f)){console.error("no data");process.exit(0);}
 const rows=fs.readFileSync(f,"utf8").split(/\r?\n/).filter(Boolean).map(l=>{try{return JSON.parse(l)}catch{return null}}).filter(Boolean).slice(-100);
 console.log("timestamp,session_id,model,rate_bucket,input_tokens,output_tokens,cache_write_tokens,cache_read_tokens,estimated_cost_usd");
-for(const r of rows)console.log([r.timestamp,r.session_id,r.model,r.rate_bucket,r.input_tokens,r.output_tokens,r.cache_write_tokens,r.cache_read_tokens,r.estimated_cost_usd].join(","));
+const q=v=>{const t=v==null?"":String(v);return /[",\r\n]/.test(t)?'"'+t.replace(/"/g,'""')+'"':t;};
+for(const r of rows)console.log([r.timestamp,r.session_id,r.model,r.rate_bucket,r.input_tokens,r.output_tokens,r.cache_write_tokens,r.cache_read_tokens,r.estimated_cost_usd].map(q).join(","));
 '
 ```
 
--- a/scripts/hooks/cost-tracker.js
+++ b/scripts/hooks/cost-tracker.js
@@ -68,19 +68,94 @@ function readHarnessCost(sessionId, maxAgeSeconds) {
   }
 }
 
-// Approximate per-1M-token billing rates (USD).
-// Cache creation: 1.25x input rate. Cache read: 0.1x input rate.
+// Per-1M-token billing rates (USD). Cache write is 1.25x input and cache read
+// is 0.1x input in every row, for the default 5-minute cache TTL. The 1-hour
+// TTL bills writes at 2x input and is still not modelled here, as the header
+// notes.
+//
+// The previous table priced every `opus` model at $15/$75, which are Claude 3
+// Opus era rates. Opus 4.5 and later bill at $5/$25, so every current-
+// generation Opus session was reported at exactly 3x real spend (measured: a
+// session summing to 100,000,000 cache-read plus 500,000 output tokens on
+// `claude-opus-5` costs $62.50 and was reported as $187.50). Fable and Mythos
+// had no bucket at all, so they fell through to `sonnet` and were understated
+// 3.3x. The rate constants were the whole defect; the arithmetic that
+// consumes them is unchanged.
+//
+// The legacy rows exist so that correcting the current generation does not
+// reprice the old one. `haikuLegacy` is Claude 3.5 Haiku, for which the
+// previous $0.80/$4.00 was exactly right. Claude 3 Haiku ($0.25/$1.25) is
+// deliberately not modelled: Claude Code never ran it.
 const RATE_TABLE = {
-  haiku:  { in: 0.80,  out: 4.0,  cacheWrite: 1.00,  cacheRead: 0.08 },
-  sonnet: { in: 3.00,  out: 15.0, cacheWrite: 3.75,  cacheRead: 0.30 },
-  opus:   { in: 15.00, out: 75.0, cacheWrite: 18.75, cacheRead: 1.50 }
+  haiku:       { in: 1.00,  out: 5.0,  cacheWrite: 1.25,  cacheRead: 0.10 },
+  haikuLegacy: { in: 0.80,  out: 4.0,  cacheWrite: 1.00,  cacheRead: 0.08 },
+  sonnet:      { in: 3.00,  out: 15.0, cacheWrite: 3.75,  cacheRead: 0.30 },
+  opus:        { in: 5.00,  out: 25.0, cacheWrite: 6.25,  cacheRead: 0.50 },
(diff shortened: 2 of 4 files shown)


andrest:
Verified both findings against the code and pushed f4efbd53.

### `commands/cost-report.md`: CSV output

Half confirmed, half not.

The escaping problem is real and is now fixed. `join(",")` emitted raw fields, so any value containing a comma, a quote, or a newline produced CSV that a parser would mis-split. Every field is now escaped per RFC 4180.

The legacy-row claim does not reproduce. `Array.prototype.join` converts `undefined` to an empty string, so a row written before `rate_bucket` existed already emitted an empty field rather than the literal `undefined`:

```js
> [1, undefined, 2].join(",")
'1,,2'
```

The escaper preserves that, so output for legacy rows is byte-identical to before.

### `scripts/hooks/cost-tracker.js`: Sonnet 5 introductory pricing

Confirmed as a fact, deferred as a change. I would like your call on it.

The rate is right. Sonnet 5 bills $2/$10 per million tokens through 2026-08-31 and $3/$15 from 1 September, while `RATE_TABLE.sonnet` carries only the standard rate. Sonnet 5 sessions priced today are therefore overstated by 50%, which is the same defect class this PR exists to fix.

I did not ship it for two reasons. The window closes in three weeks, leaving a dated branch someone has to find and delete. More importantly, pricing is currently a pure function of model and token counts, and doing this correctly makes it a function of wall-clock time as well. That is a design change rather than a rate correction, so it is your call rather than this PR's.

If you want it, the shape that fits the existing code is a `SONNET_5_INTRO_RE` alongside `LEGACY_OPUS_RE` plus a cutoff constant, since introductory pricing covers Sonnet 5 alone and not Sonnet 4.6 or 4.5. Happy to push it here or open it separately.

haelyra:
Thank you for the unusually careful rate-history work, the measured Opus regression, and the end-to-end assertions. The generation-specific Opus and Haiku buckets, Fable/Mythos coverage, synthetic-entry handling, source-labeled `rate_bucket`, and CSV escaping are all valuable. I’m sorry this did not receive a maintainer decision sooner.

We would love one focused accounting pass before merge:

- **Represent Sonnet 5’s time-bounded price without rewriting history.** Anthropic’s current pricing contract is $2/$10 through August 31, 2026 and $3/$15 starting September 1, while this head prices every Sonnet 5 token at $3/$15. Because the hook recomputes a cumulative transcript row, selecting a bucket from only the Stop time would later reprice August usage. Success means applying the documented rate to each usage event’s timestamp, or explicitly treating mixed/undated Sonnet 5 history as unestimated.
- **Do not turn unknown future models into plausible Sonnet totals.** The current `sonnet-fallback` still records a dollar figure that dashboards can sum as real spend. Carry forward #2596’s strongest contract: mark an unrecognized model unestimated, preserve its tokens, and omit or null the estimated cost unless the harness supplied an authoritative total. This prevents the next price/model launch from silently recreating the same defect.
- **Pin the pricing contract in tests and docs.** Add boundary cases immediately before and after the September 1 UTC cutoff, a mixed-date transcript, and an undated/unknown model. Link the official pricing source from the implementation comment so the next maintainer can audit each constant.

If you do not have bandwidth for this pass, we will add the remaining accounting work to the maintainer queue after the next release and preserve your attribution. Thank you again for doing the hard historical-model work.

chenhz01:
Adding `rate_bucket` is a good call for auditability. One data-compat gap: rows written before this change have no `rate_bucket`, and their `estimated_cost_usd` was computed at the stale $15/$75 Opus rates — `cost-report.md` sums the latest row per session, so any report spanning the fix mixes 3x-inflated legacy rows with corrected ones and overstates spend with no way to tell which is which. Cheapest fix: treat a missing `rate_bucket` as legacy and flag it in the report, e.g.

```js
const bucket = r.rate_bucket || 'pre-bucket-legacy';
```

then have the report print a warning line counting such rows — optionally recompute them from `model` plus token counts on read. Happy to open a PR with the backfill flag if useful.

Issue #2656, opened 2026-08-03
cost-estimate.js has the same stale Opus rate and silent Sonnet fallback as #2574 (currently dead code)


AmirF194:
While fixing #2574 in `scripts/hooks/cost-tracker.js` (#2596), found the identical bug shape in `scripts/lib/cost-estimate.js`:

- `RATE_TABLE.opus` is stale (`{ in: 15.0, out: 75.0 }`), same as the pre-fix `cost-tracker.js` table.
- `estimateCost()` falls through to `RATE_TABLE.sonnet` for any model name that doesn't match `haiku` or `opus` (no entry for Fable or anything else), silently mispricing it.
- `tests/lib/cost-estimate.test.js` encodes both as intended behavior: `'opus 1M/1M tokens returns 90'` and `'unknown model falls back to sonnet rates'`.

Repro (Docker, `node:20-slim`, HEAD `0c1d7be9a750627fb2a6534c78a998cc46d03f9c`):

```
node -e "
const { estimateCost, RATE_TABLE } = require('./scripts/lib/cost-estimate.js');
console.log(RATE_TABLE);
console.log('opus 1M/1M ->', estimateCost('opus', 1e6, 1e6));
console.log('claude-fable-5 1M/1M ->', estimateCost('claude-fable-5', 1e6, 1e6));
"
```

```
{ haiku: { in: 0.8, out: 4 }, sonnet: { in: 3, out: 15 }, opus: { in: 15, out: 75 } }
opus 1M/1M -> 90
claude-fable-5 1M/1M -> 18
```

Different from #2574 in one way worth flagging before anyone reaches for a fix: `estimateCost`/`RATE_TABLE` from this file has no callers anywhere else in the repo outside its own test (`git log --oneline --follow -- scripts/lib/cost-estimate.js` shows a single commit, `940135ea`, the one that added it). So today this has no live behavioral impact, unlike `cost-tracker.js` which the statusline hook actually calls. Filing it separately rather than folding into #2596 for that reason, not because it's a different root cause, more so it doesn't silently ship the same wrong numbers if this module gets wired up later.

Happy to send a PR mirroring #2596's fix if that's wanted, or to leave it if the module is meant to be replaced rather than fixed.

haelyra:
Confirmed as the same stale-table/unknown-model shape as #2574, but with no production caller on current main. I’ve marked it P2 rather than folding it into the live hook defect.

[...]

PR #2692, opened 2026-08-06
fix(hooks): count subagent transcripts in the cost tracker

andrest:
## What Changed

`scripts/hooks/cost-tracker.js` now folds a session's **subagent transcripts** into the same usage total as its main transcript, and prices each turn at the model and speed tier that actually served it.

- Derives `<transcript_path minus .jsonl>/subagents/` from the transcript path, globs `*.jsonl`, and folds those usages into the existing dedupe-by-`message.id` map.
- Accumulates per model **and** per speed tier, pricing each bucket separately instead of pricing the whole session at one model's rate.
- Prices **fast mode** at its real 2x tier, read from `message.usage.speed`.
- Adds two additive row fields, `subagent_transcripts` and `models`, documented in `/cost-report` and the `cost-tracking` skill.

## Why This Change

**Subagent spend was invisible.** The hook read only `input.transcript_path`. A session that fans out writes each subagent to its own JSONL, and that spend is genuinely the parent's (the subagent lines carry the parent's `sessionId`), but those files were never opened.

Measured over **116 local sessions that fanned out**: **$1,837 of $13,493 of real spend was invisible, 13.6% of the total.** On the worst session the hidden half was nearly as large as the visible one.

Folding them in is safe because subagent files share **no `message.id`** with the parent transcript (0 collisions across those 116 sessions), so the existing dedupe map absorbs them additively. A test pins that guarantee rather than the observation, by feeding the same `message.id` through both a parent and a subagent file and asserting it counts once.

**Per-model pricing is required by the fix, not separate polish.** The old code priced every token at the last model the session named. Subagents routinely run a different model than the parent: **48 of 116 local fan-out sessions, 41.4%**, most often an Opus parent delegating to Sonnet or Haiku. Folding subagent tokens into a single-model total would have traded the under-count for an over-count.

**Fast mode is recoverable from the transcript.** Fast mode is a real 2x billing tier. `message.usage.speed` records which tier served each request, so no heuristic is needed: it is present on **123,155 of 151,286 assistant usage blocks across 1,500 local transcripts (81.4%)**, with the value `standard` on every non-absent case in that sample. The multiplier keys off `speed` rather than off a list of models known to offer the tier, because a model without a fast tier never emits `speed: "fast"`, so a list would add staleness without adding accuracy. An absent `speed` reads as standard, so transcripts written before the field existed reprice identically.

## Backward compatibility

Every pre-existing field keeps its name, type and meaning, so readers of `costs.jsonl` need no change. In particular `model` still names the **main session's** model: letting a Haiku subagent win last-model-wins would relabel an Opus session, which is what `/cost-report` groups by. Subagent spend still lands in the totals and in the new `models` breakdown.

The token counts and `estimated_cost_usd` do go **up** for fan-out sessions. That is the defect being fixed, not a regression.

## Testing Done

Every new test was verified **RED before GREEN**. With the new tests in place and `scripts/hooks/cost-tracker.js` reverted to this branch's base:

```
Results: Passed: 10, Failed: 10
```

With the implementation applied, same command, same environment:

```
=== Testing cost-tracker.js ===

  ✓ passes through input on stdout
  ✓ creates metrics file when given transcript usage data
  ✓ counts usage once per message.id across multi-line responses
  ✓ handles empty input gracefully
  ✓ handles invalid JSON gracefully
  ✓ handles missing usage fields gracefully
  ✓ prefers ECC_SESSION_ID over CLAUDE_SESSION_ID when both are present
  ✓ uses input session_id for session correlation when env vars are absent
  ✓ prefers fresh harness-cost cache over transcript estimate
  ✓ ignores stale harness-cost cache (>300s) and uses transcript estimate
  ✓ counts subagent transcripts in the session total
  ✓ a session with no subagents directory is unchanged and silent
  ✓ dedupes a message.id that appears in both parent and subagent
  ✓ prices each model at its own rate rather than the parent model rate
  ✓ models breakdown sums to estimated_cost_usd and to the token totals
  ✓ stays fail-open on a malformed subagent transcript
  ✓ ignores non-jsonl sidecars in the subagents directory
  ✓ bills a fast-mode turn at 2x the standard tier
  ✓ prices an absent speed field at standard rates
  ✓ splits one model into separate buckets per speed tier

Results: Passed: 20, Failed: 0
```

Full suite, this branch:

```
  Total Tests: 3452
  Passed:      3441  ✓
  Failed:        11  ✗
```

Baseline on this branch's merge base, with the changes stashed:

```
  Total Tests: 3442
  Passed:      3431  ✓
  Failed:        11  ✗
```

The same **11 failures are pre-existing**, in the same 5 suites (`detect-project-worktree`, `gateguard-fact-force`, `observe-subdirectory-detection`, `suggest-compact`, `dry-run`), none touched here. The delta is exactly the 10 tests added.

`eslint`, `markdownlint`, `check-unicode-safety`, `validate-skills`, `validate-commands`, `validate-hooks` and `catalog:check` all pass.

The fail-open contract has explicit coverage: a missing subagents directory (the normal case for a session that never fanned out) is silent and leaves the numbers unchanged, and a malformed subagent file is skipped rather than thrown.

## Type of Change
- [x] `fix:` Bug fix

- [ ] Shell scripts pass shellcheck (if applicable) - no shell scripts touched

## Documentation
- [x] Updated relevant documentation (`commands/cost-report.md`, `skills/cost-tracking/SKILL.md`)
- [ ] README updated (if needed) - not needed, no surface change

## Notes for the reviewer

- **Relationship to #2691.** That PR corrects the rate constants in this same file; this one does not touch them, so the two are independent but will conflict textually in the `getRates` region. Whichever lands second needs a small adaptation: #2691 renames `getRates` to `resolveRates` returning `{bucket, rates}`, and this PR adds a `getRatesForSpeed(model, speed)` wrapper around `getRates`. Once both are in, `rate_bucket` becomes a single scalar over a session that can now span several buckets, which is worth a follow-up.
- Fast-mode tests are pinned as a **ratio** to the standard tier, not as absolute dollars, precisely so they stay correct across #2691's repricing.
- `/cost-report`'s `By model` grouping still keys on `model`, attributing each session to one model. Switching it to consume `models` is a natural follow-up, deliberately left out to keep this diff to the defect.
- `models` is always the transcript-derived attribution. When the harness's authoritative `cost.total_cost_usd` supplies `estimated_cost_usd`, the two need not agree, and the docs say so.

### DIFF
--- a/commands/cost-report.md
+++ b/commands/cost-report.md
@@ -16,7 +16,23 @@ session**, so the report takes the **latest row per `session_id`** and sums
 across sessions (summing every row would multiply-count).
 
 Row schema:
-`{ timestamp, session_id, transcript_path, model, input_tokens, output_tokens, cache_write_tokens, cache_read_tokens, estimated_cost_usd }`
+`{ timestamp, session_id, transcript_path, model, input_tokens, output_tokens, cache_write_tokens, cache_read_tokens, estimated_cost_usd, subagent_transcripts, models }`
+
+Token totals and `estimated_cost_usd` cover the session's subagent turns as
+well as its main transcript; `subagent_transcripts` counts the subagent files
+that were folded in, and is `0` for a session that never fanned out.
+
+`models` breaks the session down by the model and speed tier that actually
+served each turn: `[{ model, speed, input_tokens, output_tokens,
+cache_write_tokens, cache_read_tokens, estimated_cost_usd }]`, most expensive
+first. It is the more precise basis for a per-model breakdown than `model`,
+which names only the main session's model while a fan-out session commonly
+bills across several. Its token counts always reconcile with the row totals.
+The `By model` grouping below still keys on `model`, which attributes each
+session to one model; read `models` when you need the split. Its costs sum
+to `estimated_cost_usd` except when the harness supplied that number directly,
+in which case `models` remains the transcript-derived estimate and the two need
+not agree.
 
 ## What this command does
 
--- a/scripts/hooks/cost-tracker.js
+++ b/scripts/hooks/cost-tracker.js
@@ -21,6 +21,18 @@
  * To get per-session cost, take the last row per session_id. To get per-day
  * spend, aggregate.
  *
+ * Subagent transcripts: a session that fans out writes each subagent to its own
+ * JSONL beside the main one, at
+ *   <transcript_path minus .jsonl>/subagents/agent-*.jsonl
+ * That spend is the parent session's (the subagent lines carry the parent's
+ * `sessionId`) but it lived in files this hook never opened, so every fan-out
+ * session was under-reported. Measured over 116 local sessions that fanned out:
+ * $1,837 of $13,493 real spend was invisible, 13.6% of the total, and on the
+ * worst session the hidden half was nearly as large as the visible one. The
+ * subagent files share no `message.id` with the parent transcript (0 collisions
+ * across those 116 sessions), so folding them into the same dedupe map is
+ * purely additive and cannot double-count.
+ *
  * Harness-cost contract (optional, opt-in by the statusline):
  *   If the user's statusline (which receives `cost.total_cost_usd` directly
  *   from Claude Code) writes `{ts, cost_usd}` to
@@ -29,7 +41,9 @@
  *   the cache is fresh (≤ 300s). The transcript-sum is kept as a safe
  *   fallback because:
  *     - the hard-coded rate table cannot represent Opus 4.7's >200K-token
- *       2x tier or the 1h-cache 2x tier (under-counts on long sessions);
+ *       2x tier or the 1h-cache 2x tier (under-counts on long sessions).
+ *       Fast mode is no longer in that list: `message.usage.speed` is present
+ *       in the transcript and is now priced (see FAST_SPEED below);
  *     - summing the full transcript double-counts work done across
  *       `--resume` boundaries while `cost.total_cost_usd` is per-process.
  *   Absent a writer, behavior is unchanged.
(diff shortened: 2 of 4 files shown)


andrest:
Verified all five findings against the code and pushed be567d5a. All five are real. One needed a different fix than the one proposed.

### Subagent count (lines 256-259)

Confirmed and fixed, with a correction to the proposed patch. Gating on `readUsageInto(...).ok` is not sufficient: `ok` is false only when the read itself throws, so `agent-bad.jsonl` in test 25 (readable, every line fails `JSON.parse`) still returned `true` and still counted.

`readUsageInto` now reports `usageEntries`, and a file counts only once it yields at least one usage-bearing line. That covers all three cases your comment names: unreadable, unparseable, and a directory named `*.jsonl`.

### Sort determinism (lines 373-375)

Confirmed and fixed. `localeCompare` reads the active ICU locale, which contradicts the comment's byte-for-byte claim. `speed` now breaks the remaining tie, so two equal-cost buckets of one model no longer fall back to Map insertion order.

### `skills/cost-tracking/SKILL.md` (lines 30-32)

Confirmed and fixed. `commands/cost-report.md` carried the caveat and the skill omitted it, which is exactly the mismatch that would make an agent report a false discrepancy on a harness-priced session.

### Both test findings

Added. Test 22b pins the file-scoped synthetic key, and test 25 now asserts `subagent_transcripts === 1`.

Both new tests fail against the previous behaviour and pass with the fix:

```
# before
  ✗ keeps id-less entries from different files separate
  ✗ stays fail-open on a malformed subagent transcript
Results: Passed: 19, Failed: 2

# after
Results: Passed: 21, Failed: 0
```

chenhz01:
The 0-collision dedupe claim across 116 fan-out sessions makes the fold purely additive — that's the right verification. One residual gap visible in the patch: `getRates()` still ends with `return RATE_TABLE.sonnet` for unrecognized models, and the new per-model `models[]` buckets route every subagent model through it. A fan-out to a model outside haiku/sonnet/opus now produces confidently wrong per-bucket costs — the same silent-sonnet-fallback shape #2656 flagged in `cost-estimate.js`, but with more surface now that buckets are exposed in `/cost-report`. Marking the source makes it auditable:

```js
buckets.push({ model, rate_source: hit ? 'table' : 'sonnet-fallback', ... })
```

Happy to open a PR with that if useful.

PR #2703, opened 2026-08-07
fix(plan-canvas): guard localStorage so blocked site data can't disable canvas controls

### LINKED ISSUE #2702: plan-canvas: unguarded localStorage read disables all canvas controls when site data is blocked
## Summary

In the plan-canvas browser client, an **unguarded `localStorage` read** at the top of the client IIFE throws when the browser blocks site data for the origin (`127.0.0.1`). Because it throws during initialization — **before** the send-button `click` handler and the chat-input `Enter` `keydown` handler are bound — the whole client script aborts. The page renders fully (buttons visible, `disabled === false`, presence shows "agent listening"), but **every control is inert**: neither "Send to agent", "Approve plan", "Request changes", nor Enter does anything.

Symptom as a user hits it: "the Send to Agent button does nothing, on any canvas." A page reload does not help, because the same read throws again on every load.

## Root cause

`scripts/lib/plan-canvas/ui.js`, in `canvasClientJs()`. `sessionStorage` is correctly guarded two lines above and below, but the two `localStorage` calls between them are not:

```js
// line 191 — guarded
try { queue = JSON.parse(sessionStorage.getItem(QKEY) || '[]'); } catch { queue = []; }
...
// line 200 — UNGUARDED, throws here and kills the rest of the IIFE
let theme = localStorage.getItem(themeKey) || 'dark';
...
// line 204 — also unguarded
localStorage.setItem(themeKey, theme);
...
// line 241 — guarded
function persistQueue() { try { sessionStorage.setItem(QKEY, JSON.stringify(queue)); } catch { /* full */ } }
```

`localStorage`/`sessionStorage` **property access itself throws a `SecurityError`** (not returns `null`) when a browser has site data disabled for the origin — e.g. Chrome "Block third-party cookies" / site-data settings, Safari private mode historically, or a hardened profile. Loopback origins are a common place for this to bite.

The send handlers are defined further down the same IIFE (`sendBtn.addEventListener('click', ...)` and the input `keydown` listener), so a throw at line 200 guarantees they never bind.

## Reproduction

1. In a browser that blocks site data for the canvas origin (or stub it), open any canvas.
2. Observe every control renders and looks enabled.
3. Click "Send to agent" or press Enter — nothing happens, no network request fires. Reload does not fix it.

Confirmed by running the generated client against a `localStorage` stub whose accessors throw:

```
OLD build   send/Enter handlers bound: false
PATCHED     send/Enter handlers bound: true
```

## Suggested fix

Guard the `localStorage` access the same way `sessionStorage` already is:

```diff
-  let theme = localStorage.getItem(themeKey) || 'dark';
+  function readTheme() {
+    try { return localStorage.getItem(themeKey); } catch { return null; }
+  }
+  function writeTheme(v) {
+    try { localStorage.setItem(themeKey, v); } catch { /* site data blocked */ }
+  }
+  let theme = readTheme() || 'dark';
   applyTheme(theme);
   $('themeBtn').addEventListener('click', () => {
     theme = theme === 'light' ? 'dark' : 'light';
-    localStorage.setItem(themeKey, theme);
+    writeTheme(theme);
     applyTheme(theme);
   });
```

Theme persistence degrades gracefully (defaults to dark, doesn't persist) instead of taking the entire review UI down with it. Note `/client.js` is generated from this module at require time, so the canvas server must be restarted for a rebuild to take effect.

Happy to open a PR if that's preferred.

brianjkelly:
Fixes #2702.

## Problem

An unguarded `localStorage` read at the top of the plan-canvas client IIFE throws when the browser blocks site data for the origin (`127.0.0.1` is a common trigger). Because it throws during init — **before** the send-button `click` handler and the chat-input `Enter` handler bind — the whole client aborts. Every control renders and looks enabled (`disabled === false`, presence shows "agent listening"), but nothing responds, on every canvas, and reloads don't help. `sessionStorage` is already guarded two lines above and below; the `localStorage` calls between them were not.

## Fix

Route the two `localStorage` calls through `readTheme()` / `writeTheme()` helpers with try/catch, matching the existing `sessionStorage` guards. Theme persistence degrades gracefully (defaults to dark, doesn't persist) instead of taking the whole review UI down.

## Verification

Ran the generated client against a `localStorage` stub whose accessors throw:

```
before   send/Enter handlers bound: false
after    send/Enter handlers bound: true
```

`node --check` passes on the modified `ui.js`. Note `/client.js` is generated from this module at require time, so a canvas server restart is needed to pick up the rebuild.


haelyra (approved):
Thank you for this focused reliability fix. A blocked localStorage read or write should disable theme persistence, not the entire Plan Canvas review interface. I verified the guarded initialization shape, JavaScript syntax, all 28 Plan Canvas server tests, and all 9 real browser tests at ebb4451e. This is approved, with merge still held for the hosted PR gate and serial exact-main gate.

haelyra (approved):
Thank you again for the Plan Canvas resilience fix. I merged current main into the branch after the stale hosted matrix failed, then re-ran JavaScript syntax, all 28 Plan Canvas server tests, and 64 Markdown plus 16 session tests on the exact new head 9c1f68ce; all pass, and the diff against main remains limited to the guarded client change. This exact head is approved, with merge still held for fresh hosted CI and the serial exact-main gate.

PR #2649, opened 2026-08-01
feat(install): add guided multi-harness installer

haelyra:
ECC install welcome message with gradient ASCII wordmark and boxed community links

## Summary

- add one guided installer for any Claude Code, Codex, and Kimi Code selection, including all non-empty harness combinations
- preserve provider-native behavior: Claude scopes and hooks, Codex native plugin lifecycle and verified marketplace provenance, and Kimi project-managed configuration without changing model, provider, authentication, or hooks
- preflight every selected harness before mutation and fail closed on symlinks, collisions, containment violations, unwritable destinations, changed install-state, late file creation, and partial-apply races
- bind Kimi managed-update ownership to exact operation identity plus persisted SHA-256 content provenance so forged, legacy, or edited state cannot authorize overwrites
- refresh existing Codex marketplaces through the shell-free native CLI, require trusted GitHub origins, validate the exact upgraded root, and re-verify origin and enabled plugin inventory after mutation
- bound Claude/Codex provider commands, preserve structured errors, sanitize guided-dispatch failures, and keep hook/profile/spinner/welcome behavior deterministic
- publish the `ecc-universal` binary alias with packed npm, pnpm, Yarn, and Bun runner coverage
- prepare synchronized 2.2.0 package/plugin metadata and truthful package-manager documentation

## PR stack and boundaries

- contains exact #2601 head `2dae6fc2` for terminal-preserving guided setup
- contains exact #2603 head `e0ae24d6` for safe automatic scope migration
- current with `main` at `f1fec0e5`
- excludes Docker harness work in #2625 and the reusable terminal opener in #2650
- if this combined installer PR lands, #2601 and #2603 can close as superseded

## Verification

- [x] exact-head local full suite: 3,629/3,629
- [x] exact-head coverage suite: 3,629/3,629
- [x] coverage: 89.33% statements/lines, 81.00% branches, 94.21% functions
- [x] npm, pnpm, and Yarn packed-tarball runner smokes: 6/6 each locally
- [x] native Codex 0.146 fresh, repeat-refresh, and untrusted-provenance smoke paths
- [x] install manifest, release, README, npm publish surface, lint, shell syntax, and package dry-run gates
- [x] npm registry signatures and attestations verified; high-severity audit reports zero vulnerabilities; supply-chain IOC scan passes
- [x] independent code and security re-reviews report no remaining blocker
- [x] all 18 review conversations audited, dispositioned, and resolved
[...]

PR #2715, opened 2026-08-07
feat(install): require an explicit hook decision at the apply layer

haelyra:
## Summary

Carries forward the useful idea from #2634 by @samartomar, integrated with the guided installer that landed in #2649.

The guided installer already asks how ECC hooks should run, but that consent lived only in the wizard path. Running `install-apply` directly with a profile that includes `hooks-runtime` still materialized the hook runtime with no disclosure and no decision. This gates the apply layer instead, so every entry point is covered.

- Disclose the six hook capability groups when a plan would materialize the hook runtime, and refuse to apply until the caller decides.
- `--enable-hooks` confirms the hook runtime. `--no-hooks` installs the rest of the selection without it and records the reduced module closure in install-state.
- Surface the pending decision as a dry-run warning rather than a hard failure.
- Show the same capability disclosure in the guided installer's plan preview, so the wizard's hook question states what it is actually asking about.

## Attribution

The six-group capability taxonomy, the plain-language disclosure copy, and the held-materialization behavior are @samartomar's work from #2634. This PR reworks them to fit the single-decision consent model (`off|minimal|standard|strict` hook profiles) that shipped with the guided installer, rather than #2634's six independent `allow`/`decline` flags, which would have asked users to decide twice. The commit carries a `Co-Authored-By:` trailer.

## What is intentionally not carried over

- Removing `hooks-runtime` from the `core` profile. `minimal` already exists as the no-hooks profile, and `--no-hooks` now covers the same need without changing what `core` means for existing users.
- The six independent authorization flags, for the double-prompt reason above.
- Core agent/platform materialization narrowing and the OpenCode/MCP activation changes, which are separate concerns worth their own review.

## Boundaries

Plans that never materialize hooks are unaffected and need no flag: Kimi managed installs, `--profile minimal`, and `--without baseline:hooks`. Repair and uninstall operate on already-recorded install-state and are unchanged, so a previously consented install still repairs cleanly.

## Verification

- New unit suite `tests/lib/hook-consent.test.js`: 8/8.
- Install-related suites all green: `install-apply` 37/37, `install-guided` 14/14, `install-executor` 16/16, `selective-install` 46/46, `install-lifecycle` 48/48, `install-request` 13/13, `install-plan` 11/11, `multi-harness-setup` 19/19, `install-claude-skill-migration` 16/16, `uninstall` 3/3, `repair` 3/3, `install-readme-clarity` 11/11.
- ESLint clean on all changed JavaScript files.
- Manual smoke: gate fires on `--profile core`; `--enable-hooks` writes `hooks/hooks.json`; `--no-hooks` omits it and drops `hooks-runtime` from recorded state; `--target kimi` needs no flag.

Closes #2634

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,13 +413,19 @@ For the normal core profile with hooks disabled:
 
 ```bash
 ./install.sh --profile core --without baseline:hooks --target claude
+./install.sh --profile core --no-hooks --target claude
 ```
 
 Add the hook runtime later only if you want it:
 
 ```bash
-./install.sh --target claude --modules hooks-runtime
+./install.sh --target claude --modules hooks-runtime --enable-hooks
 ```
+
+Any install whose profile or modules would materialize the hook runtime requires
+an explicit decision. Without `--enable-hooks` or `--no-hooks`, the installer
+prints what the hooks can do and stops before writing anything. The guided
+installer (`ecc install --guided`) asks for this choice interactively.
 </details>
 
 <details>
@@ -510,7 +516,7 @@ For hand-picked manual installs, Claude discovers skills as direct children of `
 Do not copy the raw repo `hooks/hooks.json` into `~/.claude/settings.json` or `~/.claude/hooks/hooks.json`. That file is plugin/repo-oriented; use the installer so hook command paths are rewritten correctly:
 
 ```bash
-bash ./install.sh --target claude --modules hooks-runtime
+bash ./install.sh --target claude --modules hooks-runtime --enable-hooks
 ```
 
 That writes resolved hooks to `~/.claude/hooks/hooks.json` and leaves any existing `~/.claude/settings.json` untouched.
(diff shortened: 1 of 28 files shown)


haelyra:
**CI status note:** the Lint, Coverage, and ubuntu Test matrix failures on this PR are inherited from `main`, not introduced here.

`main`'s own latest CI run shows the identical two failures:

- **Lint** — `no-regex-spaces` in `tests/ci/ito-trade-planner-skill.test.js:37` (two literal spaces inside a regex literal).
- **Test** — `ci/ito-inference-skill.test.js` fails on `ships canonical inference through the existing opt-in compute module`. This is the only failing test file in the run; every other suite passes.

Both come from the recent Itô skill series (#2708–#2712) and are unrelated to the installer. Verified locally on this branch: ESLint and markdownlint both exit 0 across every file this PR changes, and all twelve install-related suites pass (`hook-consent` 8/8, `install-apply` 37/37, `install-guided` 14/14, `install-executor` 16/16, `selective-install` 46/46, `install-lifecycle` 48/48, `uninstall` 3/3, `repair` 3/3, and the rest).

This PR should not be merged until `main` is green again, per the usual target-branch rule.

PR #2716, opened 2026-08-07
feat(statusline): add ECC usage bar for Claude and Codex

haelyra:
> **Credit**: inspired by [leeguooooo/claude-code-usage-bar](https://github.com/leeguooooo/claude-code-usage-bar) (MIT), which pioneered the always-on rate-limit bar for Claude Code — with the prompt-cache widget originally contributed by [@marcwimmer](https://github.com/marcwimmer). ECC re-implements the idea in dependency-free Node with ECC branding and ECC-specific segments. Openness is one of our principles: go star the original.

Replaces the single-line ECC statusline with a three-line usage bar, and brings the same design to Codex. Installed by default on both harnesses.

## Claude Code

```text
⚡ 5h ███████░ 84% ↻1h54m │ 7d █░░░░░░░ 14% ↻4d │ cache 100%
✳ Opus 5 xhigh │ ctx █░░░░░░░░░ 13% 1M │ $7.85 +49/-49 36m │ Fixing auth bug
⬢ ECC 2.2.0 │ hooks standard │ plugins ecc 2.2.0 · swift-lsp 1.0.0 │ myproject
```

- Line 1 uses the official `rate_limits` from statusline stdin, so it disappears cleanly on API-key billing.
- Line 3 is the ECC-specific part: version, hooks profile (`hooks off` in red, `(N off)` for `ECC_DISABLED_HOOKS`), and enabled plugins with versions.
- The installer writes a launcher to `~/.claude/ecc-statusline.js` that resolves the active plugin version, and claims `statusLine` **only when the user has none** — an existing bar is never replaced.
- `ECC_STATUSLINE_COMPACT=1` restores the previous single-line format.

## Codex CLI

Codex exposes no custom statusline API ([#14043](https://github.com/openai/codex/issues/14043), [#20244](https://github.com/openai/codex/issues/20244)), so this uses every surface it does have:

1. **In-TUI**: configures Codex's native `tui.status_line` widgets plus `status_line_use_colors`. The key is silently ignored outside the `[tui]` table — verified by capturing the rendered TUI. Rate-limit widgets are deliberately omitted so the same window isn't reported twice.
2. **tmux**: the `ecc-codex` wrapper sets a 4-line status bar for the lifetime of the run (session-scoped, previous options restored on exit, `bg=default` so it inherits the terminal's theme) and renders the full three-line bar.
3. **Any terminal**: mirrors the bar into the window title and an OSC 1337 user var (WezTerm/iTerm2 status bar); `--full` prints the three lines for `watch`.

A managed shell alias makes plain `codex [args...]` use the wrapper — arguments pass through untouched, and aliases don't expand inside scripts so there is no recursion. A user-authored `codex` alias or function in **any** managed rc blocks installation entirely rather than shadowing it.

## Notes for review

- **UTF-8**: block glyphs need a UTF-8 locale — without one tmux mangles them to underscores. The guided install adds a step asking to set a detected locale (default yes, skipped entirely when the locale is already UTF-8); declining falls back to an ASCII glyph set (`ECC_BAR_GLYPHS=ascii|unicode` overrides).
- **Session selection** picks the most recently *modified* rollout file, since resuming an older Codex session appends to its original filename.
- Everything is opt-out: `ECC_CODEX_BAR=off`, `ECC_CODEX_ALIAS=off`, `setup-codex-bar.js remove`.

## Validation

- Full suite: 3684/3687 — the 3 failures are `scripts/release-heading.test.js`, which fails identically on pristine `origin/main` (verified by stashing).
- New tests: statusline render (41), Codex renderer (12), shell alias (16), Codex status line (7), plus statusline install cases in `claude-plugin-setup`.
- ESLint + markdownlint clean; wrapper passes `bash -n`.
- Verified end to end in a Docker sandbox: fresh Debian container with Codex pre-installed → `ecc install --guided --harness codex --yes` → alias, `[tui]` status line, and the three-line tmux bar all confirmed on screen via nested-tmux capture, in both WezTerm and Terminal.app.


chenhz01:
Long thread — one install-robustness point I don't believe is covered. The installer writes a launcher at `~/.claude/ecc-statusline.js`, but the registered statusline command still starts with bare `node`. Claude Code invokes the statusline in a non-interactive context where nvm/volta-managed node installs are frequently absent from PATH — the bar then renders blank or errors on every prompt, for exactly the users who manage their own toolchain. Since the installer itself runs under a working node, it can resolve `process.execPath` once and bake the absolute interpreter path into the launcher (keeping a `command -v node` fallback inside). The same applies to the Codex side (`scripts/codex/ecc-codex` and the bar pane): pin the interpreter at setup time rather than lookup time. For an installed-by-default rollout this is the highest-frequency failure I'd expect. Happy to open a PR with the launcher change.

PR #2721, opened 2026-08-08
fix(hooks): close the PowerShell destructive-command gate bypass

noliverbked:
## The hole

`settings.json` can set `CLAUDE_CODE_USE_POWERSHELL_TOOL=1`, which makes **PowerShell a second arbitrary-command shell** alongside Bash. Every destructive-command gate in this repo is scoped to `Bash` alone, so identical destructive intent through the two tools produced opposite outcomes.

Reproduced through the real `run-with-flags` invocation path, before the fix:

```
Bash       + recursive force delete  -> {"permissionDecision":"deny", ... "Destructive command detected"}
PowerShell + Remove-Item -Recurse -Force <path>
                                     -> {"tool_name":"PowerShell", ...}   <- passthrough, exit 0, no gate
```

## Why widening the matcher alone is not enough

It fails at two independent levels:

1. **`gateguard-fact-force.js`** — `TOOL_MAP` had no `powershell` key, and `if (toolName === 'Bash')` was the only branch reaching `isDestructiveBash()`. Unrecognised tool names fall through to the bare allow at the end of `run()`.
2. **`governance-capture.js`** — `SECURITY_RELEVANT_TOOLS` held only `'Bash'`, and a *second* Bash-only branch guarded `APPROVAL_COMMANDS`. So a command that bypassed the gate was also invisible to the audit trail. The two failures compound.

## Approach

PowerShell delete verbs share no syntax with POSIX `rm`, so reuse of the Bash tokenizer is not possible: the PowerShell recursive-delete form reaches `isDestructiveRm()` as tokens whose `-Force` fails its case-sensitive `/f/` test. Hence a dedicated `isDestructivePowerShell()`.

It **does** reuse this file's existing quote-aware segmenter (`quoteAwareSegments`) rather than a naive `String.split`. That is deliberate — the naive split reintroduces exactly the evasion class documented in GHSA-4v57-ph3x-gf55, and reusing the hardened splitter also resolves the `&` call operator and quoted command words with no extra code.

Coverage beyond the base case:

| Evasion | Handling |
|---|---|
| Quoted separator in a path (`"C:/evil;dir/*"`) | quote-aware segmentation |
| `& 'Remove-Item' ...` (call operator) | `&` is already a segment separator |
| `Get-ChildItem -Recurse \| Remove-Item` | statement-level pass keeps pipelines intact so upstream `-Recurse` stays visible |
| Backtick obfuscation | escape normalisation |
| `powershell -Command` / `pwsh -c` / `cmd /c` | nested-shell unwrap, mirroring the existing `sh -c` handling |
| `-EncodedCommand <base64>` | UTF-16LE decode, then re-scan |
| `cmd /c rd /s /q` | cmd.exe recursive-removal pattern |
| `[System.IO.Directory]::Delete($p,$true)` | .NET delete pattern |
| `Remove-Item @params` (splatting) | opaque parameters are gated rather than assumed safe |

Parameter matching uses prefixes, because PowerShell accepts any unambiguous abbreviation: `-Recurse`, `-Rec` and `-r` must all match.

### One deliberate departure from the `rm` analogy

`-Force` is gated **even without** `-Recurse`. Unlike `rm -f`, PowerShell's `Remove-Item -Force <directory>` removes the whole tree, and a static string cannot distinguish a file target from a directory target. A plain `Remove-Item <path>` with neither flag stays ungated.

### Bash behaviour is unchanged

The PowerShell detector runs **only** for the PowerShell tool. `isDestructiveBash()` still runs for PowerShell input so shell-agnostic phrases (force push, hard reset, `DROP TABLE`, `dd`) stay covered. Verified by differential testing of the pre- and post-fix modules across 15 representative Bash commands: **0 differences**.

The routine first-command gate intentionally **shares** `ROUTINE_BASH_SESSION_KEY` across both tools, so an operator sees one fact-forcing prompt per session rather than two. Its message is now tool-neutral ("first shell command").

## Verification

The two-arm test that originally proved the bug — **both now deny**:

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"<recursive force delete>"},"session_id":"verify-bash-arm"}' \
    | node scripts/hooks/run-with-flags.js pre:bash:gateguard-fact-force \
      scripts/hooks/gateguard-fact-force.js standard,strict
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny",
 "permissionDecisionReason":"[Fact-Forcing Gate]\n\nDestructive command detected. ..."}}

$ echo '{"tool_name":"PowerShell","tool_input":{"command":"Remove-Item -Recurse -Force C:/tmp/demo"},"session_id":"verify-ps-arm"}' \
    | node scripts/hooks/run-with-flags.js pre:bash:gateguard-fact-force \
      scripts/hooks/gateguard-fact-force.js standard,strict
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny",
 "permissionDecisionReason":"[Fact-Forcing Gate]\n\nDestructive command detected. ..."}}
```

Note when reproducing: the destructive gate denies only on the **first** attempt per session per command-hash, so use a distinct `session_id` per arm or the second run will look like a false "allow".

Test suites:

```
node tests/hooks/gateguard-fact-force.test.js   -> 164 passed, 0 failed   (20 new PowerShell cases; 144 pre-existing unaffected)
node tests/hooks/governance-capture.test.js     ->  30 passed, 0 failed
node tests/hooks/posttooluse-dispatcher.test.js ->  13 passed, 0 failed
```

`governance-capture.js`'s new rules were additionally verified by direct invocation: `Remove-Item -Recurse`, `-Rec`, `rd -Recurse`, `Clear-Content`, `Format-Volume` and force-push all raise `approval_requested`; `Get-ChildItem` does not; and `Start-Process -Verb RunAs` now raises `security_finding: elevated_privilege_command`, which was unreachable before because `PowerShell` was absent from `SECURITY_RELEVANT_TOOLS`.

## Review

Reviewed by two independent reviewers (security and code-quality), both of which initially returned **BLOCK** on the first draft with live-verified bypasses. Every bypass they demonstrated — quoted-separator evasion, pipe-to-delete, call operator, bare `-Force` on a directory, nested shell, `cmd /c`, backtick obfuscation, splatting, .NET delete — is now a named regression test in this PR.

## Known residual risk (not addressed here)

Static command inspection has a ceiling, and PowerShell's unrestricted .NET access raises it. These remain undetected and are **not** claimed as covered:

- `Invoke-Expression` / `iex` over a runtime-constructed string
- Parameters or paths supplied through variables (`$r='-Recurse'; Remove-Item $r $p`)
- Arbitrary .NET beyond the `IO.Directory`/`IO.File` delete pattern matched here
- Reflection or dynamically resolved cmdlet names

These are the same limits the Bash path has. The gate is a fact-forcing speed bump, not a sandbox, and it should not be described as one.

## Suggested follow-up (not in this PR)

`tests/hooks/governance-capture.test.js` has no PowerShell cases; the new rules there are verified by manual invocation (above) but not by the suite. Worth adding alongside.


haelyra (approved):
Thank you for catching the PowerShell GateGuard gap. I completed the maintainer patch with a dedicated PowerShell route, fail-closed nested-command scanning, wildcard protection, rm alias handling, and direct configured-route coverage, then refreshed it on current main. The GateGuard, governance, dispatcher, and hook suites pass. Approving this exact head for hosted CI.

haelyra (changes requested):
Thank you for taking on the PowerShell gap, and I am sorry this has waited so long. The direction is valuable and fits the ECC 2.2 security and evidence substrate, but I cannot merge the exact head while two reproduced gaps remain: a destructive command inside a PowerShell `$()` subexpression bypasses the pre-execution gate, and force-only, wildcard, plus .NET deletion forms can be denied without matching `approval_requested` evidence. Please add quote-aware recursive subexpression parsing, align the governance classifier with every PowerShell form the gate denies, and cover those cases through the configured hook entry points. If you do not have bandwidth for that revision, no problem; I queued ECC-039 so maintainers can finish it after ECC 2.2. Thank you again for pushing this safety surface forward.

chenhz01:
Detection coverage looks strong — delete verbs, aliases, nested shells, .NET routes. One tokenizer gap worth closing while the PowerShell path is fresh: pipeline segments. `Get-ChildItem -Recurse $x | Remove-Item -Force` classifies only if the tokenizer splits on `|` — the Bash side already treats `;` and `&&` as segment boundaries via the quote-aware pass, and `CMD_RECURSIVE_DELETE` matches `|` as a prefix, but the PS verb detectors key on the first token of their input. Treating `|` as a segment separator on the PowerShell path:
```js
segments = command.split(/\|(?=(?:[^']*'[^']*')*[^']*$)/)
```
closes the most idiomatic PowerShell deletion pattern that isn't a bare cmdlet call. If you want this implemented rather than advised, I take fixed-scope contracts on repos like this — feel free to reach out.

PR #2729, opened 2026-08-09
fix(skills): Windows/AppLocker portability + ordering fixes for scan scripts

noliverbked:
Ports the rules-distill and skill-stocktake scan scripts off `find | sort` and jq process substitution — both break silently on AppLocker-hardened Windows machines (shimmed find/sort return empty on GNU flags; native Windows jq cannot open MSYS /proc fd paths, hidden by 2>/dev/null). Also:

- Zero-pads the per-file tmp JSON counter (`printf -v fname '%06d.json' "$i"`) so the `jq -s '.' "$tmpdir"/*.json` merge keeps insertion order past 9 files (unpadded, 10.json sorts before 2.json)
- Deletes dead `count_obs()` in skill-stocktake/scan.sh (superseded by the two-pass jq pre-aggregation; zero call sites)

Verified: `bash -n` clean on all three; 12-skill fixture returns count 12 in exact insertion order on both skill scanners; 3-rule fixture returns total 3. Pure-bash globstar + awk counting, no new dependencies; behavior on Linux/macOS unchanged.

https://claude.ai/code/session_014Zw3jpqfpiZzjSdRK25RiJ


haelyra (approved):
Thank you for identifying the macOS Bash 3.2 portability failures. I completed the maintainer patch with deterministic recursive traversal, streamed jq input, whitespace-safe counting, and portability regression tests, then refreshed it on current main. Bash 3.2 syntax, portability, and Unicode checks pass. Approving this exact head for hosted CI.

haelyra (changes requested):
Thank you for the careful Windows and Bash portability work, and I am sorry the final integration review took this long. I previously approved this direction, but the exact head now needs reconciliation with the newer #2640 discovery contract before it is safe to merge. A current-main forward-port must keep AppLocker and Bash 3.2 portability while also preserving linked skill-directory discovery, visible traversal errors, and deterministic NUL-safe handling. It must also exclude direct symlink files that can escape the scan boundary and return a clear nonzero error for malformed or unreadable observation data. This is good work, but resolving those contracts is larger than a safe conflict patch in this batch. If you do not have bandwidth, that is completely okay; I queued ECC-040 for maintainers to finish after ECC 2.2. Thank you again for the strong portability foundation.

chenhz01:
Heads-up on a live collision: #2492 rewrites the same `skills/skill-stocktake/scripts/scan.sh` for the same MSYS failure class (process substitution to jq, argv-length limits, `dp` tilde canonicalization), and both PRs reach into `save-results.sh` territory. They diverge where they overlap — this PR streams per-file JSON and zero-pads the tmp counter; #2492 passes the inventory via stdin and rewrites `quick-diff.sh` path matching with `cygpath -m`. Whichever merges second will conflict textually and semantically in the tmp-file naming and JSON aggregation paths, and the portability tests each of you added will overlap in coverage. Worth coordinating the merge order (or rebasing one onto the other) before a maintainer picks one at random. Happy to help reconcile the two approaches if useful.

Issue #1979, opened 2026-05-18
MCP health-check blocks Streamable HTTP servers on HTTP 406 (fix from closed #1155 never landed)


ezwep:
## Problem

`scripts/hooks/mcp-health-check.js` blocks all tool calls to spec-compliant Streamable HTTP MCP servers because their `HTTP 406 Not Acceptable` response to the hook's bare GET probe is not recognized as "alive".

This is a regression of (or duplicate of) closed issue #1155. Issue #1155 was marked `state_reason=COMPLETED` on 2026-04-02 but no fix was committed for it — the OP closed it themselves without a code change. The maintainer subsequently added `400`, `401`, `403` to `HEALTHY_HTTP_CODES` (commits `8baffb4`, `e363c54`) which addresses the OAuth and Bad Request cases but skips `406`.

## Reproduction

Run any MCP server that uses a spec-compliant Streamable HTTP transport (e.g. the official MCP TypeScript SDK with `StreamableHTTPServerTransport`, or FastMCP with `transport="http"`).

```bash
# Bare GET (what the health-check hook sends)
$ curl -I http://localhost:8100/mcp
HTTP/1.1 406 Not Acceptable
content-type: application/json
content-length: 116

# GET with the Accept header the MCP spec mandates
$ curl -I http://localhost:8100/mcp -H "Accept: text/event-stream"
HTTP/1.1 200 OK
content-type: text/event-stream

# Real MCP initialize handshake (also works)
$ curl -X POST http://localhost:8100/mcp \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize", ...}'
HTTP/1.1 200 OK
data: {"result":{"protocolVersion":"2025-03-26", ...}, "jsonrpc":"2.0","id":1}
```

The hook caches `{"status":"unhealthy","lastError":"HTTP 406"}` and blocks every subsequent MCP call to that server with:

```
[MCPHealthCheck] <server> is unavailable (HTTP 406). Blocking <tool> so Claude can fall back to non-MCP tools.
```

## Root cause

The MCP Streamable HTTP transport spec (revision 2025-03-26) requires clients to send `Accept: text/event-stream` on the GET endpoint. Spec-compliant servers reject GETs that omit this header with `406 Not Acceptable`. The probe does not set any `Accept` header:

```js
const req = client.request(
  url,
  {
    method: 'GET',
[...]

PR #1345, opened 2026-04-10
fix(hooks): treat HTTP 400 MCP probes as reachable

Gujiassh:
## Summary
- treat HTTP 400 probe responses as reachable MCP HTTP servers instead of marking them unhealthy
- add a regression test that boots a local HTTP server returning 400 and verifies the health state stays healthy

## Validation
- `node tests/hooks/mcp-health-check.test.js`

---
## Summary by cubic
Treat HTTP 400 MCP HTTP probe responses as reachable so HTTP-based MCP servers aren’t incorrectly marked unhealthy or put on backoff. Adds a regression test that spins up a local server returning 400 to verify the server stays healthy.

<sup>Written for commit 08126370aa55d55e8c97badad41ebddd0ced92ab. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP 400 responses are now correctly treated as healthy for server health probes, affecting health assessment in preflight checks and post-failure recovery.

* **Tests**
  * Added test coverage for HTTP-based server health checks.

PR #2749, opened 2026-08-10
fix(hooks): treat HTTP 404 MCP probes as reachable

bengio777:
## Problem

`mcp-health-check.js` probes HTTP MCP servers with a bare `GET`. Some Streamable HTTP servers route only `POST /mcp` and answer a bare `GET` with **404**. The preflight scores that as down and blocks every tool call for that server:

```
[MCPHealthCheck] paper is unavailable (HTTP 404). Blocking get_guide so Claude can fall back to non-MCP tools.
```

This is not transient. The backoff is 30s, after which the hook re-probes, gets 404 again, and re-blocks — so the server is unusable indefinitely while being perfectly healthy.

Reproduced against Paper Desktop 0.5.3 (`http://127.0.0.1:29979/mcp`):

| Request | Result |
|---|---|
| `GET /mcp` | `404` |
| `POST /mcp` (initialize) | `200`, `serverInfo: {"name":"paper-desktop","version":"0.5.3"}` |

## Fix

Add `404` to `HEALTHY_HTTP_CODES`.

A routed HTTP response of *any* status proves the endpoint is reachable, which is the only thing this preflight claims to check — the comment above the set already says so, and `400`/`401`/`403`/`405`/`406` are all in the set for exactly this reason. Reachability is the hook's job; endpoint validity is the real MCP client's job.

## Test

Adds `treats HTTP 404 probe responses as healthy POST-only Streamable HTTP servers`, mirroring the existing HTTP-400 case. It stands up a server that 404s on `GET` and 200s on `POST /mcp` — the Paper shape.

Verified it is a real regression test, not a tautology:

| Code state | `tests/hooks/mcp-health-check.test.js` |
|---|---|
| Without the fix | **22 passed, 1 failed** — the new test fails with `exit=2`, `lastError: "HTTP 404"` |
| With the fix | **23 passed, 0 failed** |

## Notes

- No other suite touches `HEALTHY_HTTP_CODES`; the diff is limited to that constant, its explanatory comment, and the new test.
- This is the same class of bug as #1979 (406) and #1345 / #2023 (400). Those statuses are in the set today; 404 is the remaining gap for POST-only endpoints.

### DIFF
--- a/scripts/hooks/mcp-health-check.js
+++ b/scripts/hooks/mcp-health-check.js
@@ -28,8 +28,11 @@ const MAX_BACKOFF_MS = 10 * 60 * 1000;
 // Claude Code's stored OAuth bearer token. Treat auth-gated responses as
 // reachable so the real MCP client can attempt the authenticated call. A
 // Streamable HTTP MCP server can also return 406 to a bare GET that omits
-// Accept: text/event-stream; that still proves the endpoint is alive.
-const HEALTHY_HTTP_CODES = new Set([200, 201, 202, 204, 301, 302, 303, 304, 307, 308, 400, 401, 403, 405, 406]);
+// Accept: text/event-stream; that still proves the endpoint is alive. Some
+// POST-only Streamable HTTP servers (e.g. Paper Desktop) answer a bare GET
+// with 404 instead; a routed HTTP response of any kind proves reachability,
+// so treat 404 as alive and let the real MCP client validate the endpoint.
+const HEALTHY_HTTP_CODES = new Set([200, 201, 202, 204, 301, 302, 303, 304, 307, 308, 400, 401, 403, 404, 405, 406]);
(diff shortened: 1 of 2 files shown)


haelyra (approved):
Approved. The preflight probe is a transport-reachability check, so a routed 404 should allow the authenticated, method-correct MCP call to produce the authoritative endpoint result. I verified the exact head locally with all 23 MCP health-check tests and ESLint. Thank you for covering the POST-only server case.

haelyra:
Thank you for this focused fix, and sorry for the delay. Your 404 reachability case and regression coverage have now landed through the consolidated MCP health-check PR #2869. HTTP 404 probe responses are treated as proof that the MCP endpoint is reachable instead of marking the server dead, and the final exact-main CI matrix and CodeQL are green.

GitHub closed this PR through the consolidation. We appreciate the clear patch and test case, and I hope you will keep contributing.

PR #2751, opened 2026-08-10
fix(hooks): use valid wildcard matchers

### LINKED ISSUE #2748: Plugin hooks with matcher "*" are silently dropped from /hooks (Stop/SessionStart groups never register)
**Environment:** Claude Code 2-1-226, Windows 11

**Problem:** In plugin `hooks/hooks.json`, any group using `matcher: "*"` never gets registered. `/hooks` shows only groups with valid regex matchers (e.g. `Bash`, `Edit|Write`, `""`). Groups are silently dropped — no warning.

**Root cause:** `"*"` is not a valid regular expression — `new RegExp("*")` throws `SyntaxError: Nothing to repeat`. The plugin hook loader silently skips the group (newer builds) instead of failing loudly (older builds killed the whole config, see anthropics/claude-code#75081).

**Evidence (ecc's own hooks.json):**
- 16/22 groups across all events use `matcher: "*"` (Stop x6, SessionStart, PreCompact, PostToolUse observe/metrics, etc.)
- After changing Stop groups to `matcher: ""`, they register and appear in `/hooks`.
- PreToolUse groups with valid matchers (`Bash`, `Edit|Write`) already worked.

**Impact:** continuous-learning, mcp-health-check, cost-tracker, session-end, format-typecheck, desktop-notify are silently inactive.

**Suggestion:**
- Replace `matcher: "*"` with `matcher: ""` (empty regex matches everything) or `".*"`.
- Add a schema/validation warning in the plugin hook loader for invalid regex matchers.

**Related:** #2688 (inline node -e commands), anthropics/claude-code#75081.

SulimanAbdulrazzaq:
## What Changed
- Replace every `matcher: *` in the plugin `hooks/hooks.json` with the valid catch-all regex `matcher: .*`.
- Add a regression test that compiles every string matcher so invalid regular expressions cannot silently reach Claude Code's hook loader.

## Why This Change
Claude Code treats hook matchers as regular expressions. `*` is invalid (`Nothing to repeat`), so affected groups were silently omitted from `/hooks`. The valid `.*` pattern preserves the intended match-all behavior and keeps all existing hook groups loadable.

Closes #2748

## Testing Done
- `node scripts/ci/validate-hooks.js` — passed (`Validated 22 hook matchers`)
- Direct JSON/regex validation — passed (`Validated 22 string matchers`)
- `corepack yarn eslint tests/hooks/hooks.test.js` — passed
- `node tests/hooks/hooks.test.js` — attempted; the existing full hook suite exceeded 5 minutes on this Windows runner before completion, so no pass is claimed.

## Type of Change
- [x] `fix:` Bug fix
- [x] `test:` Tests

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -36,7 +36,7 @@
         "id": "pre:edit-write:suggest-compact"
       },
       {
-        "matcher": "*",
+        "matcher": ".*",
         "hooks": [
           {
             "type": "command",
@@ -73,7 +73,7 @@
         "id": "pre:config-protection"
       },
       {
-        "matcher": "*",
+        "matcher": ".*",
         "hooks": [
           {
             "type": "command",
@@ -98,7 +98,7 @@
     ],
     "PreCompact": [
       {
-        "matcher": "*",
+        "matcher": ".*",
         "hooks": [
           {
             "type": "command",
@@ -111,7 +111,7 @@
     ],
     "SessionStart": [
       {
-        "matcher": "*",
+        "matcher": ".*",
         "hooks": [
           {
             "type": "command",
@@ -122,7 +122,7 @@
         "id": "session:start"
       },
       {
-        "matcher": "*",
+        "matcher": ".*",
         "hooks": [
           {
             "type": "command",
@@ -135,7 +135,7 @@
     ],
     "PostToolUse": [
       {
-        "matcher": "*",
+        "matcher": ".*",
         "hooks": [
           {
             "type": "command",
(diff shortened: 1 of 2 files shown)


haelyra (approved):
Approved. This fixes the wildcard entries to use valid regular expressions across the hook manifest and adds a regression check that compiles every string matcher. I verified the exact head locally with the hook manifest validator, the full hook test harness, and ESLint. Thank you for tightening a subtle configuration contract.

haelyra:
Thank you for correcting the hook wildcard matchers, and apologies for the delayed landing. The valid regex matcher behavior is now on main through #2902 with your authorship preserved, including coverage for JavaScript regex literals after return. The exact main commit passed all 42 CI jobs and CodeQL. Closing this PR only because the approved fix has landed. Please keep sending focused fixes like this.

PR #2753, opened 2026-08-10
fix(observer): pass absolute analysis paths

### LINKED ISSUE #2673: continuous-learning-v2: observer analysis resolves relative path against $HOME, not PROJECT_DIR — batch archived unanalyzed despite failure (#2370 semantic-failure path)
## Summary

On macOS, the observer's Claude subprocess resolves `analysis_relpath` against `$HOME` instead of the `PROJECT_DIR` the script `cd`s into. The analysis agent reports the file as missing, **exits 0**, and the `#2370` archive guard — which only checks `exit_code -ne 0` — lets the batch be archived anyway. A 2.7 MB / 1,519-observation batch was consumed and permanently dropped without producing any instinct.

## Environment

| | |
|---|---|
| ECC | 2.1.0 (`ff15079`) |
| Claude Code | 2.1.221 |
| OS | macOS (Darwin 25.5.0), zsh |
| Observer model | `haiku` (default) |
| Project bucket | `global` → `PROJECT_DIR=~/.local/share/ecc-homunculus` |
| Observer | freshly enabled via `${CONFIG_DIR}/config.json`, `observer.enabled: true` |

## What happened

`observer.log`, verbatim (home path redacted):

```
[Tue Aug  4 16:46:39 +0630 2026] Observer started for global (PID: 27569)
[Tue Aug  4 16:49:07 +0630 2026] Analyzing     1519 observations for project global...
[Tue Aug  4 16:49:07 +0630 2026] Using last      500 of     1519 observations for analysis
**File not found.** The observer analysis file at
`/Users/<user>/.observer-tmp/ecc-observer-analysis.jsonl.jQcI4j` does not exist.

Cannot proceed with pattern analysis without the source data.

**Status:**
- Read target file: ✗ FAILED (not found)
- Pattern analysis: ⊘ BLOCKED (no data)
- Instinct generation: ⊘ BLOCKED (prerequisites unmet)
```

Note the path the agent resolved: **`$HOME/.observer-tmp/`**.

The file was actually created at the correct location — that directory exists and is the one `observer-loop.sh` uses:

```
/Users/<user>/.local/share/ecc-homunculus/.observer-tmp/     <- created by the script
/Users/<user>/.observer-tmp/                                  <- does not exist; agent looked here
```

## Root cause

`agents/observer-loop.sh`:

```sh
observer_tmp_dir="${PROJECT_DIR}/.observer-tmp"          # L146
mkdir -p "$observer_tmp_dir"
analysis_file="$(mktemp "${observer_tmp_dir}/ecc-observer-analysis.jsonl.XXXXXX")"
...
analysis_relpath=".observer-tmp/$(basename "$analysis_file")"   # L159, relative for #842
...
cd "$PROJECT_DIR" || { ... }                              # L244, guarded, did not fail
...
claude --model haiku --print --allowedTools "Read,Write" -p "$prompt_content" < /dev/null &
```

The `cd` at L244 is correct and did not error (no "Failed to cd" line in the log). The shell's cwd was `PROJECT_DIR`. But the spawned Claude Code process did **not** resolve `.observer-tmp/...` against that cwd — it resolved against `$HOME`.

`PROJECT_DIR` here is `~/.local/share/ecc-homunculus` (the `global` bucket maps `_CLV2_PROJECT_DIR` to `_CLV2_HOMUNCULUS_DIR`). That directory is a data dir with no project markers, which may be why the subprocess falls back to `$HOME` for relative-path resolution — but the mechanism is a guess; the observable fact is the mismatch above.

## Why #2370's guard does not catch it

```sh
if [ "$exit_code" -ne 0 ]; then
  echo "... retaining observations for retry" >> "$LOG_FILE"
  return
fi
# archive
mv "$OBSERVATIONS_FILE" "$archive_dir/processed-$(date ...).jsonl"
```

The agent did not crash. It read its instructions, could not find the file, wrote a tidy status report, and **completed normally -> exit 0**. So the guard passed and the batch was archived 37 seconds after the failure:

```
16:49:07  analysis logged, agent reports file not found
16:49:44  processed-20260804-164944-27569.jsonl   2.7M   <- 1,519 observations, archived unanalyzed
```

`#2370` fixed the *process-failure* path. This is a **semantic** failure with a zero exit code, which reaches the same outcome that issue was closed to prevent. The analyzer only ever reads the live observations file, so those 1,519 observations can never be re-analyzed.

## Impact

- Silent data loss of an entire observation batch per occurrence — worst on first enable, when the accumulated backlog is largest.
- Failure is invisible unless `observer.log` is read by hand; nothing surfaces in `/instinct-status`, which just reports `No instincts found.`
- The next cycle (20 observations) succeeded and correctly declined to write instincts (threshold unmet), so the observer is otherwise healthy — this is specific to the relative-path resolution.

## Suggested fixes

Either alone mitigates; both together are safer.

1. **Don't rely on the subprocess inheriting cwd.** Pass the absolute path in the prompt. `#842` chose a relative path for Windows/MSYS2 path-prefix reasons, so this could be conditional — absolute on non-Windows, relative on Windows — or the prompt could name both and instruct the agent to try each.

2. **Treat "no instincts written AND agent reported a read failure" as a failed analysis.** A cheap version: after the run, if the agent's output matches a not-found/blocked pattern, `return` before archiving. A stronger version: have the prompt require a sentinel line on success (e.g. `ANALYSIS_COMPLETE`) and gate the archive on its presence rather than on exit code alone.

3. **Optionally, verify readability before spawning** — `[ -r "$analysis_file" ]` already holds, so the useful check is on the path the agent will actually use, which the script cannot know. (2) is the more robust guard.

## Reproduction

1. Fresh install, observer never enabled, so a large `observations.jsonl` accumulates in the `global` bucket (a non-git cwd sends sessions there).
2. Enable via `~/.local/share/ecc-homunculus/config.json` -> `{"observer":{"enabled":true,...}}`.
3. Trigger a cycle (20 tool calls).
4. `observer.log` shows the not-found report; `observations.archive/processed-*.jsonl` contains the batch; `instincts/` stays empty.

## Minor, possibly out of scope

The observer `cd`s to `PROJECT_DIR` and runs `claude --print`, so any user `Stop` hook fires with that cwd. With `agent-timelog` installed this creates `~/.local/share/ecc-homunculus/.time-log.md` and appends the analysis agent's own entries into the data directory. `ECC_HOOK_PROFILE=minimal` does not cover third-party user hooks. Harmless here, but it does mean the observer can write unexpected files into the homunculus dir.

SulimanAbdulrazzaq:
## Summary

Closes #2673.

The observer hook now receives an absolute analysis-file path on non-Windows systems, avoiding Claude resolving the relative path against HOME and silently archiving an empty observation. Windows keeps its existing relative-path behavior. Added regression coverage for both platforms.

## Validation

- node tests/hooks/observer-memory.test.js (32 passed)
- git diff --check
- bash syntax check unavailable on this Windows host (bash is not installed)

### DIFF
--- a/skills/continuous-learning-v2/agents/observer-loop.sh
+++ b/skills/continuous-learning-v2/agents/observer-loop.sh
@@ -153,10 +153,17 @@ analyze_observations() {
   analysis_count=$(wc -l < "$analysis_file" 2>/dev/null || echo 0)
   echo "[$(date)] Using last $analysis_count of $obs_count observations for analysis" >> "$LOG_FILE"
 
-  # Use relative path from PROJECT_DIR for cross-platform compatibility (#842).
-  # On Windows (Git Bash/MSYS2), absolute paths from mktemp may use MSYS-style
-  # prefixes (e.g. /c/Users/...) that the Claude subprocess cannot resolve.
-  analysis_relpath=".observer-tmp/$(basename "$analysis_file")"
+  # Claude Code resolves relative paths against the user's home directory on
+  # macOS/Linux, even though the observer changes to PROJECT_DIR first. Use
+  # the absolute path there so the analyzer reads the file that was sampled.
+  # Keep the relative path on Windows (Git Bash/MSYS2), where absolute paths
+  # from mktemp can contain /c/ prefixes that the Claude subprocess cannot
+  # resolve (#842, #2673).
+  if [ "${CLV2_IS_WINDOWS:-false}" = "true" ]; then
+    analysis_relpath=".observer-tmp/$(basename "$analysis_file")"
+  else
+    analysis_relpath="$analysis_file"
+  fi
 
   prompt_file="$(mktemp "${observer_tmp_dir}/ecc-observer-prompt.XXXXXX")"
   cat > "$prompt_file" <<PROMPT
--- a/tests/hooks/observer-memory.test.js
+++ b/tests/hooks/observer-memory.test.js
@@ -220,7 +220,20 @@ test('prompt references analysis_file not full OBSERVATIONS_FILE', () => {
   assert.ok(heredocStart > 0, 'Should find prompt heredoc start');
   assert.ok(heredocEnd > heredocStart, 'Should find prompt heredoc end');
   const promptSection = content.substring(heredocStart, heredocEnd);
-  assert.ok(promptSection.includes('${analysis_relpath}'), 'Prompt should point Claude at the sampled analysis file (via relative path), not the full observations file');
+  assert.ok(promptSection.includes('${analysis_relpath}'), 'Prompt should point Claude at the sampled analysis file, not the full observations file');
+});
+
+test('observer uses an absolute analysis path outside Windows', () => {
+  const content = fs.readFileSync(observerLoopPath, 'utf8');
+  assert.ok(
+    content.includes('if [ "${CLV2_IS_WINDOWS:-false}" = "true" ]') &&
+      content.includes('analysis_relpath="$analysis_file"'),
+    'macOS and Linux must pass the absolute analysis path to Claude'
+  );
+  assert.ok(
+    content.includes('analysis_relpath=".observer-tmp/$(basename "$analysis_file")"'),
+    'Windows must retain the MSYS-compatible relative analysis path'
+  );
 });
 
 test('observer-loop wait helper retries SIGUSR1-interrupted waits while claude child is alive', () => {


haelyra (approved):
Approved. The platform split preserves the existing MSYS-safe relative path while fixing macOS and Linux, where the Claude subprocess resolves relative paths away from the observer project directory. I verified the exact head locally with all 32 observer-memory tests, Bash syntax checking, and ESLint. Thank you for including the platform-specific regression coverage.

PR #2754, opened 2026-08-10
fix(gan): allow Playwright tools in evaluator

### LINKED ISSUE #2675: gan-evaluator agent instructs Playwright MCP use but declares no MCP tools — live-app evaluation silently degrades on the /gan-build path
## Summary

`agents/gan-evaluator.md` instructs the agent to drive the live application with Playwright MCP, but its own `tools:` frontmatter does not include any `mcp__playwright__*` entry. The agent therefore cannot do the one thing that distinguishes the GAN harness from ordinary code review: testing the **running** app.

This affects the `/gan-build` and `/gan-design` (Task-tool) path — the path most users take. It is the same class of gap noted as a footnote in #2674, but that issue covers `scripts/gan-harness.sh`; this one is the agent definitions, and the two are fixed in different places.

## Evidence

`agents/gan-evaluator.md:4`

```yaml
tools: Read, Write, Bash, Grep, Glob
```

`agents/gan-evaluator.md:45-54` — the agent's own instructions:

```
### Step 2: Launch Browser Testing

# Use Playwright MCP to interact with the live app

# Navigate to the app
playwright navigate http://localhost:${GAN_DEV_SERVER_PORT:-3000}

# Take initial screenshot
playwright screenshot --name "initial-load"
```

The description field says the same: *"Tests the live running application via Playwright, scores against rubric…"*

And `commands/gan-build.md:10` defaults the mode to playwright:

```
5. `--eval-mode MODE` — (optional, default "playwright") one of: playwright, screenshot, code-only
```

So the default path asks for browser testing, the agent is told to perform browser testing, and the agent is not granted a browser tool.

## Why this matters more than it looks

SKILL.md frames live-app evaluation as the core insight — an evaluator that only reads code cannot judge Craft (spacing, transitions, responsiveness) or Functionality (do the features actually work) beyond what static reading shows. Those two criteria carry **0.5 of the default weight** (craft 0.3 + functionality 0.2).

The failure is silent. The agent has `Bash`, so it will improvise — curl the dev server, read the source, reason about markup — and still emit a confident rubric score. Nothing in the output signals that the browser was never opened.

## Precedent for the fix, already in this repo

`agents/docs-lookup.md:4` shows the pattern works for ECC agents:

```yaml
tools: Read, Grep, mcp__context7__resolve-library-id, mcp__context7__query-docs
```

So the same can be done for the evaluator:

```yaml
tools: Read, Write, Bash, Grep, Glob, mcp__playwright__browser_navigate, mcp__playwright__browser_click, mcp__playwright__browser_take_screenshot, mcp__playwright__browser_snapshot, mcp__playwright__browser_type, mcp__playwright__browser_fill_form
```

(exact list depending on which Playwright MCP server is the intended target).

## Suggested handling

1. Add the Playwright MCP tools to `agents/gan-evaluator.md`.
2. Have the evaluator **state explicitly in its feedback** which eval mode it actually achieved, rather than which one was requested — so a degraded run is visible in `feedback-NNN.md` instead of invisible.
3. Consider whether `gan-generator` needs any MCP access too; it currently has `Read, Write, Edit, Bash, Grep, Glob` and is expected to stand up a dev server, which Bash covers, so it may be fine as-is.

## Note on non-web targets

For a mobile target the Playwright path does not apply at all — driving an Android device through `adb` (`adb exec-out screencap`, `adb shell uiautomator dump`) works within the current `Bash` grant. Worth documenting that `playwright` mode is web-only, so users on other targets choose `screenshot`/`code-only` deliberately rather than silently getting a degraded playwright run.

## Environment

ECC 2.1.0 (`ff15079`), macOS 26.5. Agent definitions inspected directly at `agents/gan-{planner,generator,evaluator}.md`:

```
gan-planner.md:4    tools: Read, Write, Grep, Glob
gan-generator.md:4  tools: Read, Write, Edit, Bash, Grep, Glob
gan-evaluator.md:4  tools: Read, Write, Bash, Grep, Glob
```

None declare an MCP tool.

SulimanAbdulrazzaq:
## Summary

Closes #2675.

The GAN evaluator instructions now grant the Playwright browser tools they already require and explicitly record whether browser evaluation was achieved or fell back. Added a regression test for the frontmatter tool list and achieved-mode requirement.

## Validation

- standalone regression test passed
- node scripts/ci/validate-agents.js (67 files)
- git diff --check

### DIFF
--- a/agents/gan-evaluator.md
+++ b/agents/gan-evaluator.md
@@ -1,7 +1,7 @@
 ---
 name: gan-evaluator
 description: "GAN Harness — Evaluator agent. Tests the live running application via Playwright, scores against rubric, and provides actionable feedback to the Generator."
-tools: Read, Write, Bash, Grep, Glob
+tools: Read, Write, Bash, Grep, Glob, mcp__playwright__browser_navigate, mcp__playwright__browser_click, mcp__playwright__browser_take_screenshot, mcp__playwright__browser_snapshot, mcp__playwright__browser_type, mcp__playwright__browser_fill_form
 model: sonnet
 color: red
 ---
@@ -34,6 +34,12 @@ You are the QA Engineer and Design Critic. You test the **live running applicati
 - DO compare against what a professional human developer would ship
 
 ## Evaluation Workflow
+
+Before testing, record the mode that is actually available. The requested mode
+is not proof that its tools were available: if the Playwright MCP tools cannot
+be called, switch to the documented `screenshot` or `code-only` fallback and
+report that degradation instead of silently scoring a static review as a live
+browser evaluation.
 
 ### Step 1: Read the Rubric
 ```
@@ -128,6 +134,14 @@ Write feedback to `gan-harness/feedback/feedback-NNN.md`:
 # Evaluation — Iteration NNN
 
 ## Scores
+
+## Evaluation Mode
+
+**Achieved:** `playwright` | `screenshot` | `code-only`
+
+State the mode that was actually completed (not merely the mode requested by
+the harness). If the requested mode was unavailable, briefly explain why and
+which fallback was used.
 
 | Criterion | Score | Weight | Weighted |
 |-----------|-------|--------|----------|
--- a/tests/ci/gan-evaluator-tools.test.js
+++ b/tests/ci/gan-evaluator-tools.test.js
@@ -1,0 +1,34 @@
+/**
+ * Regression coverage for the GAN evaluator's live-browser capability.
+ *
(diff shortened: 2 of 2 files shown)


haelyra (approved):
Approved. Granting the evaluator the specific Playwright MCP calls makes the advertised live-product workflow executable, and the achieved-mode disclosure keeps fallback evaluations honest when those tools are unavailable. I verified the exact head locally with the focused contract test, all 67 agent validations, and ESLint. Thank you for fixing both capability and reporting.

PR #2758, opened 2026-08-10
fix: disable Claude co-author attribution by default

haelyra:
## Summary
- default ECC-managed Claude installs and plugin setup to `includeCoAuthoredBy: false`
- backfill the setting during Claude plugin setup and scope migration when it is missing
- preserve an explicit user opt-in of `includeCoAuthoredBy: true`

## Root cause
Claude Code appends `Co-Authored-By` by default unless `includeCoAuthoredBy` is explicitly set to `false`. ECC documented that behavior but did not actually write the setting during Claude install/setup flows, so users could have ECC installed and still get Claude as a GitHub co-author.

## Testing
- `node tests/lib/claude-plugin-setup.test.js`
- `node tests/lib/claude-scope-migration.test.js`
- direct `applyInstallPlan` library verification for fresh-write, merge, and explicit-opt-in preservation

## Notes
- `tests/scripts/install-apply.test.js` is locally blocked in this worktree by a missing `ajv` dependency before most CLI assertions run, so this PR keeps installer verification focused on the changed library path.

### DIFF
--- a/.cursor/rules/common-git-workflow.md
+++ b/.cursor/rules/common-git-workflow.md
@@ -13,7 +13,7 @@ alwaysApply: true
 
 Types: feat, fix, refactor, docs, test, chore, perf, ci
 
-Note: To disable co-author attribution on commits, set `"includeCoAuthoredBy": false` in `~/.claude/settings.json` (Claude Code appends `Co-Authored-By` by default; ECC does not ship this setting).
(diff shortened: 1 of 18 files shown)


daltino (commented):
Thanks for working on this! The changes look neat and well-structured.

PR #2352, opened 2026-06-24
Feat/pi integrations

beruang:
## What Changed

This PR adds first-class support for the Pi coding agent runtime to ECC.

### Pi package integration

* Added Pi package metadata and resource declarations for global installation through `pi install`.
* Added Pi-specific resources under `.pi/`, including:

  * runtime extensions
  * prompt templates
  * subagent definitions
  * subagent chains
  * profiles and supporting configuration
* Added global package loading support while preserving project-level override behavior.

### ECC runtime integration

* Added an ECC runtime extension for Pi.
* Added runtime status and diagnostic commands for validating the installation and companion capabilities.
* Added profile-based loading so users can control which ECC skills, rules, and workflows are active.
* Added graceful fallback behavior when optional companion extensions are unavailable.

### Skills and rules

* Reused ECC’s existing skills and rules as the canonical sources instead of duplicating their content.
* Added selective skill discovery based on the active Pi profile.
* Added rule loading and prompt injection for relevant engineering standards.
* Added safeguards against loading duplicate or unnecessary context.

### Subagent support

* Added Pi-compatible ECC agent definitions.
* Added tool-name translation between ECC agent definitions and Pi tools.
* Added support for read-only exploration, planning, correctness review, security review, and test review agents.
* Added bounded subagent execution policies, including parallelism and write-access restrictions.

### Workflow chains

* Added research and planning workflows.
* Added structured plan-approval gates before implementation.
* Added parallel review workflows for correctness, security, and test coverage.
* Added verification and delivery workflows.
* Integrated persistent task tracking for multi-step workflows.

### Hook and lifecycle integration

* Added Pi-to-ECC lifecycle event translation.
* Added support for pre-tool and post-tool checks.
* Added session start, session shutdown, agent completion, and pre-compaction handling.
* Added configurable safety checks for dangerous commands and protected paths.
* Added timeout and failure isolation so hook failures do not terminate the Pi session.

### Documentation and tests

* Added Pi installation and configuration documentation.
* Documented global installation, profiles, companion extensions, overrides, and troubleshooting.
* Added tests for Pi package metadata, resource discovery, agent conversion, profile resolution, and runtime behavior.
* Updated the ECC harness support documentation to include Pi.

## Why This Change

ECC supports multiple coding-agent harnesses, but Pi previously had no first-class integration.

Pi provides the extension lifecycle, package loading, skills, prompts, and tool APIs needed to run ECC workflows. The following companion extensions provide equivalents for several Claude-specific primitives used by ECC:

[...]

PR #2270, opened 2026-06-16
fix(omp): harden harness contract

SiaoZeng:
## What Changed
- rebased the branch onto latest `main` and reduced this PR to the OMP-only harness-contract slice requested in review
- added the OMP manifest surface in `package.json#omp` with a single native extension entrypoint, curated tool port, and commands directory; `omp.hooks` remains absent
- added `omp/extension.mjs` for native OMP lifecycle handling and guarded `.ecc/omp-session/` state writes
- added `omp/tools/index.js` with deterministic `details` envelopes across all seven ECC OMP tools
- added `tests/omp/omp-plugin.test.js` to cover manifest shape, seven-tool registration, structured success/warning/error envelopes, invalid git-summary base branches, hook registration, and opt-in lifecycle state writes
- updated `tests/scripts/npm-publish-surface.test.js` and `package.json#files` so the published package includes the OMP runtime surface
- narrowed the `git -n` no-verify guard to supported git subcommands and removed in-place mutation from OMP session-state updates
- added `.ecc/omp-session/` to `.gitignore`

## Why This Change
Fixes the OMP harness drift tracked in #2269.

Also updates PR #2270 to match maintainer feedback: this branch now contains only the OMP harness-contract work. Workflow/release, OpenCode, docs, and Python/provider changes were removed from this PR and will need follow-up PRs.

## Testing Done
- [ ] Manual testing completed
- [x] Automated tests pass locally (`node tests/run-all.js` via `npm test` / `npm run coverage`)
- [x] Edge cases considered and tested

Commands run:
- `node tests/omp/omp-plugin.test.js`
- `node tests/scripts/npm-publish-surface.test.js`
- `npm run lint`
- `npm run coverage`

## Type of Change
- [x] `fix:` Bug fix
- [ ] `feat:` New feature
- [ ] `refactor:` Code refactoring
- [ ] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `chore:` Maintenance/tooling
- [ ] `ci:` CI/CD changes

## Security & Quality Checklist
- [x] No secrets or API keys committed (ghp_, sk-, AKIA, xoxb, xoxp patterns checked)
- [x] JSON files validate cleanly
- [x] Shell scripts pass shellcheck (if applicable)
- [ ] Pre-commit hooks pass locally (if configured)
- [x] No sensitive data exposed in logs or output
- [x] Follows conventional commits format

## Documentation
- [ ] Updated relevant documentation
- [x] Added comments for complex logic
- [ ] README updated (if needed)

PR #2759, opened 2026-08-10
feat: thin Pi adapter mounting ECC's canonical skills and commands

renan-tiberio:
## What Changed

Adds first-class Pi (`@earendil-works/pi-coding-agent`) support to ECC as a **thin adapter**, following the maintainer review on #2352.

**8 files, +1,289 lines, 0 deletions.** For comparison, #2352 is 440 files, +99,628 / −2,785, which exceeded both Greptile's and CodeRabbit's review limits.

| Review point from #2352 | How this PR addresses it |
| --- | --- |
| Expose ECC's canonical skills directly instead of committing generated copies | The `pi` manifest points Pi at `./skills` and `./commands`. Nothing is copied or generated under `.pi/`. Two regression tests fail the build if `.pi/skills`, `.pi/agents`, `.pi/prompts`, `.pi/chains` reappear, or if more than 10 files are tracked under `.pi/` |
| Use Pi's documented extension lifecycle | `export default function (pi: ExtensionAPI)` with `pi.on("session_start" \| "session_shutdown" \| "before_agent_start")`. No `app.events` |
| Resolve hooks from the installed package, execute without a shell | `path.resolve(__dirname, "..", "..")` for resolution; `execFile(process.execPath, [...])` for execution. `process.cwd()` appears nowhere in the file |
| Separate Pi-native resources from optional subagent resources | Manifest declares only `extensions`, `skills`, `prompts`. No `agents` / `chains` keys — those belong to `pi-subagents` and would be silently ignored by Pi's core manifest. A test asserts they stay absent |
| Start with a small, verifiable feature set | Package loading, canonical skill/command discovery, `/ecc-doctor`, and two lifecycle mappings. Profiles, approval UI, todos, agent conversion, and chains are explicitly out of scope |
| Add compatibility tests and exact tested versions | 20 new tests across two files; tested against the exact versions listed below |

### No transformation needed

ECC's assets are already Pi-compatible, which is what makes the copy-free approach work:

- `skills/*/SKILL.md` already follows the [Agent Skills standard](https://agentskills.io) that Pi implements
- `commands/*.md` frontmatter (`description`, `argument-hint`) is already Pi's prompt-template format

Rules are the exception — Pi has no rules concept — so the adapter reads them from the canonical `rules/common/` directory at runtime and appends them to the system prompt inside an `<ecc-engineering-rules>` block, reusing the injection path already built for session context. Still no copies: nothing is generated under `.pi/`.

### The adapter

`.pi/extensions/index.ts` is the only logic added. Beyond the review points above it:

- invokes hooks through ECC's own `scripts/hooks/run-with-flags.js`, so `ECC_HOOK_PROFILE` and `ECC_DISABLED_HOOKS` keep gating hooks under Pi rather than being bypassed
- runs hooks **in the user's project directory** while resolving the scripts **package-relative**, so ECC's project detection reports the user's project, not the ECC install
- honors ECC's stdin/stdout hook contract: sends the event payload as JSON on stdin and parses `hookSpecificOutput.additionalContext` from stdout, injecting it into the system prompt on the next `before_agent_start` inside an `<ecc-session-context>` block. Non-JSON output (a profile-disabled hook passing stdin through) is tolerated, not treated as an error
- isolates hook failures behind a 30s timeout and a 1 MB output cap; a missing, broken, or slow hook degrades to a warning and never terminates the session

Also registers `.pi` in the `platform-configs` install module and adds a Pi row to the harness adapter compliance matrix.

## Testing Done

Tested against **Pi 0.84.1**, **Node 24.18.1**, **macOS 15 (Darwin 25.5.0)**. No companion packages installed — the integration was exercised without `pi-subagents`, `@juicesharp/rpiv-ask-user-question`, or `@juicesharp/rpiv-todo` present.

### Global install, isolated config directory

Installed into a clean `PI_CODING_AGENT_DIR` so the results could not be contaminated by pre-existing user settings, then inspected the loaded resources through `pi.getCommands()`, filtering to entries whose `sourceInfo.path` is inside the ECC package:

```
PKG_SKILLS=285
PKG_PROMPTS=94
PKG_EXTENSION_CMDS=1
PKG_SKILL_EG=skill:accessibility  <- /path/to/ECC/skills/accessibility/SKILL.md
PKG_PROMPT_EG=aside               <- /path/to/ECC/commands/aside.md
```

Repo ground truth: `ls skills | wc -l` = 285, `ls commands/*.md | wc -l` = 94. Exact match, and the resolved paths are the canonical directories — not `.pi/`. `/ecc-doctor` resolved to `.pi/extensions/index.ts` from the package.

### Lifecycle and context injection, from a path containing a space

Ran Pi from a working directory whose path contains a space, with a probe extension asserting on the assembled system prompt:

```
PROBE_INJECTED=YES
PROBE_SNIPPET="<ecc-session-context>\nProject type: {\"languages\":[\"javascript\"],...,
               \"projectDir\":\".../proj with space\"}\n</ecc-session-context>"
```

`projectDir` is the user's project, confirming hooks execute there while resolving package-relative. An earlier iteration passed `cwd` as the ECC root and reported ECC's own directory as the project; that was caught by this test and fixed.

### Automated

```
node tests/pi/pi-package-manifest.test.js     Passed: 9,  Failed: 0
node tests/pi/pi-extension-adapter.test.js    Passed: 24, Failed: 0
npm run harness:adapters -- --check           PASS (12 adapters)
node tests/docs/harness-adapter-compliance.test.js   Passed: 10, Failed: 0
node tests/scripts/npm-publish-surface.test.js       Passed: 2,  Failed: 0
npx markdownlint-cli .pi/README.md            clean
```

All 8 CI validators pass, including `validate-install-manifests.js` for the `install-modules.json` change.

Test coverage includes: package manifest shape, the `.pi/` copy regression guards, canonical-asset Pi compatibility, `__dirname` vs `process.cwd()` resolution, absence of shell invocation, documented-lifecycle usage, a global-install simulation from a package root whose path contains a space, hook resolution when `cwd` points elsewhere, profile/disable gating, and `additionalContext` parsing including malformed and non-JSON input.

**Full suite:** `node tests/run-all.js` is green on this branch — **3,757 passed, 0 failed**. `npm run lint` (eslint + markdownlint) passes clean.

One transient failure appeared mid-work and is fixed in this PR rather than worked around: adding `.pi/` to `package.json` `files` broke `tests/scripts/npm-publish-surface.test.js`, which validates the published surface against the install module graph. The fix was to register `.pi` in the `platform-configs` module in `manifests/install-modules.json`, which is where a shipped harness directory belongs.

## Type of Change

* [x] `feat:` New feature
* [x] `docs:` Documentation
* [x] `test:` Tests
* [x] `chore:` Maintenance/tooling

* [x] No secrets or API keys committed
* [x] JSON files validate cleanly
* [x] No sensitive data exposed in logs or output
* [x] Follows conventional commits format

Security notes specific to this adapter:

* Hooks execute via `execFile` with an argv array — no shell, so paths with spaces or shell metacharacters cannot be reinterpreted as commands.
* Hook scripts resolve from the installed package, so a hostile working directory cannot substitute its own `scripts/hooks/*.js`.
* Hook failures are isolated and cannot silently authorize a blocked operation.
* `/ecc-doctor` prints paths, counts, and profile names only — no credentials or environment secrets.
* The adapter never auto-commits, pushes, merges, or deploys.

## Known Limitations

* Full subagent orchestration, chains, structured approval gates, and persistent todos are out of scope and require companion packages. ECC is usable in Pi without any of them.
* Pi has no MCP surface by design, so ECC's MCP reference configs do not apply.
* `session_shutdown` maps to ECC's `session:end:marker` hook rather than `session-end.js`: the latter expects a Claude Code transcript, and Pi's session JSONL is a different format. Feeding it one would fabricate a compatibility that does not exist.

## What Reaches Pi

| ECC asset | Ships to Pi | How |
| --- | --- | --- |
| skills (285) | all | canonical `skills/`, via the `pi` manifest |
| commands (94) | all | canonical `commands/`, via the `pi` manifest |
| rules (`rules/common/`) | portable subset, 7 files | read at runtime, injected into the system prompt |
| MCP configs (35 servers) | all | unchanged, through the community `pi-mcp-adapter` — see below |
| hooks | 2 of 22 | `session:start` and `session:end:marker`, through ECC's own `run-with-flags.js` |
| agents (67) | none | needs the `pi-subagents` companion package |

The last two rows are deliberate. Converting 67 agents would mean generating 67 files under `.pi/`, which is the pattern this PR exists to avoid, and the review asked for agent conversion and chains to follow independently. The remaining 20 hooks are Claude `PreToolUse`/`PostToolUse`/`Stop` events; Pi has equivalents (`tool_call`, `tool_result`, `agent_settled`), so they are mappable, but each carries its own semantics and belongs in a separate change.

`agents.md`, `hooks.md`, and `performance.md` are excluded from rule injection on purpose: they describe Claude Code primitives Pi does not have, so injecting them would point the model at tools that are not there. Tests assert both that they stay excluded and that none of those primitives appear in the injected text.

### MCP needs nothing from ECC

`pi-mcp-adapter` reads the standard `mcpServers` format from `.mcp.json` — exactly what ECC already uses. Verified against version 2.21.2 in an isolated `PI_CODING_AGENT_DIR`: copying `mcp-configs/mcp-servers.json` to a project's `.mcp.json` registers Pi's `mcp` tool and `/mcp` command with all 35 ECC servers discovered, coexisting with `/ecc-doctor`. No translation layer, no ECC change. ECC neither installs nor depends on that package; `/ecc-doctor` reports which optional companions are present and prints the exact `pi install` command for the ones that are not.

## Review Response

Automated review flagged two real runtime defects, both fixed in `2597544f`:

- **EPIPE could kill the session.** `stdin.end()` writes asynchronously, so a hook that exits, short-circuits, or is killed by the timeout before reading its payload raises `EPIPE` as an `error` event that the surrounding `try`/`catch` cannot observe. Unhandled, it would take the Pi session down — exactly the guarantee the adapter claims to provide. Now handled, with both a source contract test and a behavioral test that writes 2 MB to a child which exits without reading.
- **Stale context could replay across sessions.** `pendingContext` lives between `session_start` and `before_agent_start`. Pi can begin a new session (`/new`, `/resume`, `/fork`) before that value is consumed; if the newer hook then failed, the next agent start received context describing a different session's project state. Greptile reproduced this against a real failing hook. `pendingContext` is now cleared at the top of the handler, before the hook runs.

A third issue was found while verifying the review: companion-package detection used `require.resolve`, which cannot see packages Pi installs under its own config directory (`~/.pi/agent/npm`, overridable via `PI_CODING_AGENT_DIR`) because that path is not on Node's module resolution path from the extension. `/ecc-doctor` would have reported every companion as missing regardless of what was installed. It now reads Pi's own `packages` list, correctly handling scoped names with versions.

Also addressed: the compliance matrix renderer joins list entries with `"; "`, so internal semicolons and trailing periods in the Pi record split one entry into several in the rendered cell; the profile-gating test ran against the real checkout and could leave marker artifacts behind, and now uses the isolated skeleton; the `.pi/` file-count guard counted tracked files only, so untracked copies could bypass it, and now walks disk; the README heuristic rejected valid negated phrasing; and the local parser mirrors in the tests are now pinned by source assertions so they cannot silently diverge from the adapter.

## Credit

This builds on the groundwork and problem framing in #2352 by @juicesharp. The Pi integration surface, the companion-package landscape, and the installation model were mapped there first; this PR reimplements that scope as an adapter along the lines the maintainer review requested.

#2270 by @SiaoZeng is the prior groundwork for the structured tool-result envelope discussed around ECC's harness integrations. This adapter registers no tools — it exposes one command, `/ecc-doctor`, and no `registerTool` call — so there is no tool-result surface here for that envelope to apply to. If a Pi tool port happens later it will be a separate, focused PR, and it will credit #2270 for the envelope design.


daltino (commented):
Thanks for working on this! The changes look neat and well-structured.

renan-tiberio:
All review findings addressed in `2597544f`. Summary of what was accepted and what changed:

**Real defects (both confirmed, both fixed)**

| Finding | Verdict |
| --- | --- |
| CodeRabbit — `child.stdin` needs an `error` handler; `try/catch` does not catch EPIPE | **Correct.** The write is asynchronous, so a hook that exits or is killed by the timeout before reading raises `EPIPE` as an event the `catch` never sees. Unhandled it terminates the session — the exact guarantee the adapter documents. Fixed, plus a behavioral test that writes 2 MB to a child which exits without reading, asserting no uncaught exception. |
| Greptile P1 — failed session start replays prior session context | **Correct, and the reproduction was accurate.** `pendingContext` now clears at the top of the `session_start` handler, before the hook runs, so a new session cannot inherit a previous one's context whether or not the newer hook succeeds. |

**Found while verifying the review**

CodeRabbit's question about companion-package resolution led to a third real bug: `require.resolve` cannot see packages Pi installs under its own config directory (`~/.pi/agent/npm`, overridable via `PI_CODING_AGENT_DIR`), since that path is not on Node's module resolution path from the extension file. `/ecc-doctor` would have reported every companion as "not installed" regardless of reality. It now reads Pi's own `packages` list, with correct handling of scoped names carrying versions (`npm:@scope/name@1.2.3`) — the case a naive `split("@")` gets wrong.

**Also fixed**

- Compliance matrix: the renderer joins entries with `"; "`, so internal semicolons and a trailing period in the Pi record split one entry into several in the rendered cell. Verified against the regenerated table.
- Profile-gating test ran against the real checkout and could leave `session-end-marker` artifacts in a developer tree or CI; it now uses the same isolated skeleton as the neighbouring tests, with cleanup in `finally`.
- The `.pi/` file-count regression guard counted git-tracked files only, so untracked generated copies bypassed it entirely. It now walks disk (Node 18 compatible manual traversal), with the git check kept as a secondary signal.
- The README heuristic rejected valid negated phrasing such as "Nothing is copied or generated under `.pi/`". It now detects imperative copy instructions and exempts negations, verified both against the real README and against a synthetic violating sentence.
- The local parser mirrors in the tests are now pinned by source-level assertions against `.pi/extensions/index.ts`, so a change to the adapter's real guards fails the test instead of passing against a stale copy.

Adapter tests went from 11 to 17. Full suite: **3,750 passed, 0 failed**. `npm run lint` clean.

Marked as draft while the remaining scope questions in the PR description are settled.

haelyra:
Maintainer routing note: #2270 is prior groundwork for the structured tool-result envelope proposed around ECC’s Pi integration. I closed that older adapter to keep one active integration path here because its runtime and policy layer had drifted substantially. As this thin adapter stabilizes, please preserve credit to #2270 for that envelope idea and keep any generic tool port as a separate, focused follow-up.

renan-tiberio:
Understood on all three points, and thank you for the routing note.

**No tool port here.** Confirming concretely, since that is the part most at risk of scope creep: this adapter registers zero tools (no `pi.registerTool` call) and exactly one command, `/ecc-doctor`. The full diff is 8 files:

```
.pi/extensions/index.ts                          adapter logic
.pi/README.md                                    adapter docs
package.json                                     pi manifest + files entry
manifests/install-modules.json                   register .pi in platform-configs
scripts/lib/harness-adapter-compliance.js        adapter record
docs/architecture/harness-adapter-compliance.md  regenerated matrix
tests/pi/pi-extension-adapter.test.js            24 tests
tests/pi/pi-package-manifest.test.js             9 tests
```

Because nothing here produces a tool result, there is no surface in this PR for the structured envelope to attach to — which is why keeping it out is natural rather than a concession.

**Credit preserved.** The description now records #2270 by @SiaoZeng as prior groundwork for the structured tool-result envelope, and states that if a Pi tool port lands later it will be a separate PR crediting #2270 for the envelope design.

**Scope held.** Following the same rule already applied elsewhere in this PR — agent conversion and chains stay out because they would mean generating 67 files under `.pi/`, and 20 of ECC's 22 hooks stay out because Pi's `tool_call` / `tool_result` / `agent_settled` equivalents each carry their own semantics. A generic tool port joins that list as its own follow-up.

For visibility, the one capability added since your July review beyond the minimal set is rule injection: ECC's portable `rules/common/` files are read at runtime and appended to Pi's system prompt inside an `<ecc-engineering-rules>` block. It reuses the `before_agent_start` path already built for session context, so it introduces no new lifecycle mapping and copies nothing into `.pi/`. Without it, ECC in Pi had skills and commands but none of its engineering standards. Happy to split it out if you would rather the first release be strictly the minimal set.

renan-tiberio:
All three findings from the latest review addressed in `11ad39f6`. All three were real, and all three landed in the same place: `/ecc-doctor` telling a user something false about their own install.

| Finding | Verdict |
| --- | --- |
| `normalizePiPackageName` ignores object-form `packages` entries | **Correct.** Verified against Pi's own `packages.md`: an entry is either the bare source string or `{ source: "npm:x", skills: [] }` with resource filters. Only the string was recognized, so a user who narrowed which resources a companion contributes was told it was not installed — and handed an install command for something already present. The source type still decides whether a name is comparable, so an object wrapping a git source or a path stays unrecognized exactly as before. |
| `describeRulesStatus` reports the allowlist length, not what loaded | **Correct.** `loadPortableRules` drops unreadable files, drops empty ones, and breaks at `MAX_RULES_BYTES`, so an install that loaded 3 of 7 reported `7 rule file(s)` — to the one command a user runs to find a partial install. Now reported as a ratio with the shortfall named: `3/7 rule file(s), … (4 unreadable, empty, or past the size cap)`. |
| `.pi/README.md` Notes bullet contradicts the MCP section | **Correct.** Leftover from before the MCP section landed. The bullet now points at that section instead of calling MCP out of scope. |

**Tests**

Adapter tests go from 24 to 26. Both new ones are behavioral, and the two source contracts that pinned the previous spellings now pin the new guards — including the `typeof null === "object"` case an unguarded object branch would throw on — so neither fix can be silently reverted.

Full suite: **3,759 passed, 0 failed**. `npm run lint` clean.

No other actionable items in the review; the remaining entries are the inference-engine notes.

affaan-m:
think this is good to merge in. nice work!

PR #2760, opened 2026-08-11
fix(plugin): export only plugin function for opencode loader compatibility

van3hardy:
## Problem

opencode's legacy plugin loader (getLegacyPlugins) iterates every module export and throws 'Plugin export is not a function' if any export is not a plugin function. The ecc-universal bundle exported VERSION (string) and metadata (object) alongside the plugin function, which breaks plugin loading in opencode.

## Fix

Export only the plugin function from the entry point. ECCHooksPlugin remains reachable via the default export, so all hooks and custom tools stay available.

## Verification

- import() of the rebuilt dist exposes exactly one export: default (typeof function).
- Verified loading in opencode - plugin initializes without the 'Plugin export is not a function' error.

### DIFF
--- a/.opencode/index.ts
+++ b/.opencode/index.ts
@@ -35,46 +35,6 @@
  */
 
 // Export the main plugin
-export { ECCHooksPlugin, default } from "./plugins/index.js"
-
-// Export individual components for selective use
-export * from "./plugins/index.js"
-
-// Version export
-export const VERSION = "1.6.0"
-
-// Plugin metadata
-export const metadata = {
-  name: "ecc-universal",
-  version: VERSION,
-  description: "ECC plugin for OpenCode",
-  author: "affaan-m",
-  features: {
-    agents: 13,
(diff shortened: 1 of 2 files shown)


haelyra (approved):
Approved. The published OpenCode entry is loader-facing, and limiting that entry to the default plugin function prevents legacy loaders from treating metadata or named exports as plugin factories. The underlying plugin module still owns its implementation exports. I verified the exact head locally with all four OpenCode build/package tests and ESLint. Thank you for adding a direct built-artifact contract test.

haelyra:
Thank you for fixing the OpenCode loader export contract, and I am sorry for the delay. The plugin-only export is now on main through #2902 with your authorship preserved and the cross-harness matrix verified. The exact main commit passed all 42 CI jobs and CodeQL. Closing this source PR because the approved compatibility fix has landed. We appreciate the contribution.

PR #2646, opened 2026-07-31
feat(commands): add Antigravity CLI (agy) as santa-loop Reviewer B option

Or4cu1o:
## Summary
Adds the Antigravity CLI (`agy`) as a third detectable external model for `/santa-loop`'s Reviewer B, tried after `codex` and `gemini` and before falling back to a same-family Claude reviewer (which loses model diversity). Uses `gemini-3.6-flash-high` — despite the "flash" tier name, it currently outranks `gemini-3.1-pro-high` on every published coding/agentic benchmark (SWE-Bench Pro, Terminal-Bench, MLE-Bench); Pro only leads on PhD-level reasoning benchmarks (GPQA, HLE) that don't apply to code review. Also documents never pointing `agy` at a Claude model (`claude-sonnet-4-6`, `claude-opus-4-6-thinking`), which would eliminate Reviewer A/B model diversity entirely.

## Type
- [ ] Skill
- [ ] Agent
- [ ] Hook
- [x] Command

## Testing
- Verified `agy` CLI flags (`-p`, `--model`, `--sandbox`) and confirmed the model catalog via `agy models` (v1.1.9).
- Confirmed benchmark ordering (gemini-3.6-flash-high > gemini-3.1-pro-high on coding/agentic tasks) via public model-comparison sources before picking the model id.
- `npm test` (full suite: unicode safety, agents/commands/rules/skills/hooks validation, install-manifest validation, no-personal-paths, catalog:check, command-registry:check, and `tests/run-all.js`) — 3385/3385 passing, 0 failures.
- Doc-only change to an existing command; no new command/agent/skill/hook surface, so no manifest/catalog/registry entries needed updating.

## Checklist
- [x] Follows format guidelines
- [x] Tested with Claude Code
- [x] No sensitive info (API keys, paths)
- [x] Clear descriptions

PR #2768, opened 2026-08-11
feat(commands): add Antigravity reviewer to Santa loop

haelyra:
Closes #2646.

This keeps the original contribution and Santa-loop Antigravity reviewer support while aligning it with the supported ECC CCG integration contract:

- detects the installed `codeagent-wrapper`
- invokes `--backend antigravity` through that wrapper
- leaves model selection to the maintained backend contract
- retains the original reviewer priority and fallback behavior

Credit to @Or4cu1o for the feature and implementation direction in #2646. I carried the patch onto a maintainer branch because that fork does not allow maintainer edits.

Validated locally with:

- `npx markdownlint-cli2 commands/santa-loop.md`
- `node scripts/ci/validate-commands.js`
- `git diff --check`


Or4cu1o:
Thanks for tracking this down — I didn't know about the CCG wrapper contract (`codeagent-wrapper --backend antigravity`) when I opened #2646, so pinning `agy`'s model directly was the wrong call. Routing through the wrapper and letting it own model selection is the right fix, and I appreciate you keeping the original patch and credit intact. Happy for this to supersede #2646 whenever it merges.

Also just enabled maintainer edits on my open PR branches (this one's source and #2647) so future adjustments like this can land directly instead of needing a parallel PR.

chenhz01:
Keeping model selection inside the wrapper is the right contract boundary. One portability nit in the detection: `test -x "$HOME/.claude/bin/codeagent-wrapper"` assumes a POSIX executable. On Windows the CCG install ships a `.cmd`/`.exe` shim, and Git Bash's `test -x` is unreliable on `.cmd`, so Antigravity would silently fall through to the same-model Claude fallback there — losing the model diversity this PR exists to add. Or-ing a PATH lookup covers both layouts:

```bash
command -v codeagent-wrapper >/dev/null 2>&1 || test -x "$HOME/.claude/bin/codeagent-wrapper"
```

Happy to open a PR with that tweak if useful.

PR #2772, opened 2026-08-12
fix(hooks): correct cost-tracker pricing table and filter harness noise from session summaries

lojasetetoco:
## Summary

Two fixes to runtime hooks, found while auditing a fresh 2.2.0 plugin install:

### 1. `scripts/hooks/cost-tracker.js` — stale pricing table
- `RATE_TABLE.opus` was $15/$75 per MTok (Opus 3/4.0/4.1 legacy pricing). Current Opus 5 and Opus 4.5–4.8 are $5/$25, so `estimated_cost_usd` in `~/.claude/metrics/costs.jsonl` overstated Opus sessions ~3x (verified against live rows).
- `claude-fable-5` / `claude-mythos-5` matched no branch and fell through to Sonnet rates ($3/$15 instead of $10/$50), understating ~3.3x.
- Haiku updated to 4.5 pricing ($1/$5).
- Opus 4.0/4.1/3 keep the legacy tier via a dedicated `opusLegacy` branch.

### 2. `scripts/hooks/session-end.js` — harness noise in session summaries
`extractSessionSummary` treated local-command echoes as user asks, so the cross-session summary's **Tasks** list filled with `<local-command-caveat>`, `<command-name>`, `<local-command-stdout>` fragments — and SessionStart then re-injected that noise as context. Now filters those tags, `<system-reminder>`/`<task-notification>` blocks, `tool_result` carrier turns, and `isMeta` entries. Verified against a live transcript: Tasks now contains only real user messages.

## Testing
- `node tests/hooks/cost-tracker.test.js` — 10/10 pass
- `node tests/hooks/session-end.test.js` — 1/1 pass

### DIFF
--- a/scripts/hooks/cost-tracker.js
+++ b/scripts/hooks/cost-tracker.js
@@ -70,15 +70,22 @@ function readHarnessCost(sessionId, maxAgeSeconds) {
 
 // Approximate per-1M-token billing rates (USD).
 // Cache creation: 1.25x input rate. Cache read: 0.1x input rate.
+// Current-generation list prices: Fable/Mythos 5 $10/$50, Opus 5 and
+// Opus 4.5-4.8 $5/$25, Sonnet 5/4.6 $3/$15, Haiku 4.5 $1/$5. Opus 4.0/4.1
+// and Opus 3 stay on the legacy $15/$75 tier.
 const RATE_TABLE = {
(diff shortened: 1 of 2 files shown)


haelyra:
Thank you for tracing the stale cost-tracker rates and the session-summary noise. Your pricing and filtering work has now landed on main through #2866, with the connected regression coverage and maintainer integration kept together. The exact resulting main commit passed CI and CodeQL. We appreciate the concrete bug report and patch, and we are closing this source PR only because the integrated version is shipped.

PR #2776, opened 2026-08-12
fix: probe POST-only Streamable HTTP MCP servers before marking them dead

whnb773:
## Problem

The preflight probe in `scripts/hooks/mcp-health-check.js` only ever sends a bare `GET` to the server URL. Some Streamable HTTP MCP servers route `POST` exclusively and answer *any* `GET` with 404. Telnyx is one:

```
GET  https://api.telnyx.com/v2/mcp                       -> 404
GET  + Accept: application/json, text/event-stream       -> 404
POST + JSON-RPC initialize                               -> 200
```

404 isn't in `HEALTHY_HTTP_CODES`, so every probe failed against a completely healthy server. The failure count compounded to the `MAX_BACKOFF_MS` ceiling and the hook blocked every tool call for that server before it left the machine — while `claude mcp list` still reported it `✔ Connected`. The user-visible symptom is:

```
[MCPHealthCheck] telnyx is marked unhealthy until <ts>; skipping list_api_endpoints
```

with a `mcp-health-cache.json` entry showing `failureCount: 10, lastError: "HTTP 404"` that never recovers, because each retry repeats the same invalid probe.

## Fix

`requestHttp()` takes an optional method/body, and `probeServer()` replays a failed `GET` as a real JSON-RPC `initialize` `POST` before declaring the server unreachable. The GET stays the first attempt, so nothing changes for servers that already answer it.

Adding 404 to `HEALTHY_HTTP_CODES` was the cheaper alternative, but it would mask genuine outages on every other server — a 404 usually *is* a bad URL. Speaking the protocol is the honest check.

## Tests

Adds a regression test that stands up a POST-only server returning 404 for every GET and asserting the probe body is a valid JSON-RPC `initialize`. It fails without the change:

```
✗ treats POST-only Streamable HTTP MCP servers that answer every GET with 404 as healthy
  Error: Expected POST-only MCP server to survive a 404 GET probe:
  exit=2; stderr=[MCPHealthCheck] postonly is unavailable (HTTP 404). Blocking list_api_endpoints...
```

and passes with it. `node tests/hooks/mcp-health-check.test.js` -> 23 passed, 0 failed. Full `node tests/run-all.js` -> 2580 passed, 0 failed. ESLint clean.

### DIFF
--- a/scripts/hooks/mcp-health-check.js
+++ b/scripts/hooks/mcp-health-check.js
@@ -253,19 +253,28 @@ function detectFailureCode(text) {
   return null;
 }
 
-function requestHttp(urlString, headers, timeoutMs) {
+function requestHttp(urlString, headers, timeoutMs, options = {}) {
   return new Promise(resolve => {
     let settled = false;
     let timedOut = false;
 
     const url = new URL(urlString);
     const client = url.protocol === 'https:' ? https : http;
+    const method = options.method || 'GET';
+    const body = options.body || null;
+    const requestHeaders = { ...headers };
+
+    if (body) {
+      requestHeaders['content-type'] = 'application/json';
+      requestHeaders['content-length'] = Buffer.byteLength(body);
+      requestHeaders.accept = 'application/json, text/event-stream';
+    }
 
     const req = client.request(
       url,
       {
-        method: 'GET',
-        headers,
+        method,
+        headers: requestHeaders,
       },
       res => {
         if (settled) return;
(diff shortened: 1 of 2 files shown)


haelyra:
Thank you for the careful reproduction here, and sorry this waited so long. Your POST-only server case was valuable and directly informed the consolidated MCP health-check work that has now landed in #2869.

We kept the fix narrower: a 404 or 405 response from the bounded GET probe now proves reachability, while we omitted an unauthenticated POST fallback so preflight stays side-effect-free and does not make speculative requests to arbitrary MCP endpoints. The final exact-main CI matrix and CodeQL are green.

GitHub closed this PR through the consolidation. I appreciate the investigation and hope you keep sending cases like this; it materially improved the shipped behavior.

PR #2777, opened 2026-08-13
fix(lib): remove unused cost-estimate.js duplicate rate table

### LINKED ISSUE #2656: cost-estimate.js has the same stale Opus rate and silent Sonnet fallback as #2574 (currently dead code)

AmirF194:
## What Changed

Removes `scripts/lib/cost-estimate.js` and its test file `tests/lib/cost-estimate.test.js`.

## Why This Change

`cost-estimate.js` carries its own copy of the stale Opus/Haiku/Sonnet rate table already
reported in #2574 (Opus priced 3x too high, unrecognized models silently fall through to
Sonnet rates), but it has zero callers anywhere in the repo: `grep -rn "cost-estimate\|costEstimate\|estimateCost"` across every `.js`/`.json`/`.md` file outside `node_modules`
matches only the module itself and its own test. It was added in 940135e ("feat: add ECC
statusline observability hooks") alongside the context monitor, metrics bridge and
statusline scripts, but none of those ended up calling it.

The maintainer's comment on this issue named two acceptable outcomes: remove the unused
duplicate, or make it share one explicit rate source with the live tracker. `cost-tracker.js`'s
own rate-table fix (#2574) has not landed yet (still open as #2691), so sharing its table
right now would mean importing numbers that are still wrong. Removing the dead file is the
smaller, immediately-correct step; a shared-module import is a natural follow-up once the
tracker's own fix lands.

A comment on this issue proposed the import route directly. Given #2574 is still unresolved,
deletion is the option that does not depend on it landing first.

## Testing Done

- `node tests/run-all.js` run inside `node:20-slim` (Docker) on `origin/main` and on this
  branch: identical 3763-test run, identical 41 failures on both sides (byte-for-byte diff
  of the failure list), all in unrelated git/worktree/project-detection tests that need a
  checkout shape this sandbox does not provide. The repo's real CI runs these clean (e.g.
  the `Test (ubuntu-latest, Node 20.x, npm)` leg on #2759). Zero tests newly fail, zero
  newly pass.
- `npm run lint` (eslint + markdownlint): clean.
- `npm run coverage`: 88.45% statements / 80.88% branches / 93.31% functions / 88.45% lines,
  all above the project's 80/79/80/80 floor. The command's own exit code still reflects the
  41 pre-existing failures above, so I'm not calling it green in the CI sense from this
  sandbox, only reporting the coverage numbers it measured.
- Not checked: whether #2693 (open, patches this same file's table in place rather than
  removing it) is the direction you'd rather take. Happy to close this in favor of that if so.

Fixes #2656.


AmirF194:
Checking in, no rush. Happy to restructure the dedup approach if you'd prefer the rate table handled differently.

haelyra:
Thank you for identifying that the duplicate cost estimator had drifted from the active hook and was no longer used. The removal has now landed on main through #2866, preserving your contribution in the integration history, and the exact resulting main commit passed CI and CodeQL. We appreciate you simplifying the pricing surface and preventing another source of rate drift.

PR #2783, opened 2026-08-13
feat: add OrcaRouter as an OpenAI-compatible LLM provider

Marc-oss-hub:
## What Changed

Adds **OrcaRouter** as a built-in OpenAI-compatible provider in the `src/llm` abstraction layer, mirroring the existing `astraflow`/`atlas` adapters.

- `ProviderType.ORCAROUTER = "orcarouter"` in `src/llm/core/types.py`
- New `OrcaRouterProvider` in `src/llm/providers/orcarouter.py` — OpenAI-compatible chat-completions flow against `https://api.orcarouter.ai/v1`, default model `orcarouter/auto` (per-request virtual router; concrete namespaced models like `openai/gpt-4o-mini` also work)
- Registered in the resolver `_PROVIDER_MAP` and exported from `llm.providers`
- Config via `ORCAROUTER_API_KEY` / `ORCAROUTER_BASE_URL` / `ORCAROUTER_MODEL` (documented in `.env.example` and the README gateway section)
- 6 provider unit tests + a resolver registration test

Select it with `LLM_PROVIDER=orcarouter` (or `.llm.env`), or `get_provider("orcarouter")`.

## Why This Change

[OrcaRouter](https://www.orcarouter.ai) exposes 150+ models from OpenAI, Anthropic, Google, DeepSeek, Qwen and others behind a single OpenAI-compatible endpoint and one `sk-orca-*` key. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes. Adding it as a first-class provider follows the same pattern as the existing OpenAI-compatible adapters, so users get the same tool-calling, temperature, and model-config behavior they already rely on.

## Testing Done

- [x] `python -m ruff check src tests` — clean
- [x] `python -m pytest tests/test_*.py -m "not integration"` — 91 passed (incl. 7 new OrcaRouter tests)
- [x] L3 live test through `OrcaRouterProvider.generate()` against the real endpoint: `LIVE-OK model=gpt-4o-mini-2024-07-18 reply='ORCA-OK'` and routed `orcarouter/auto` → upstream model
- [ ] Automated tests pass locally (`node tests/run-all.js`)

## Type of Change

- [x] `feat:` New feature

- [x] No secrets or API keys committed (live key used only in a throwaway local test, removed afterwards)

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,6 +37,13 @@ ATLAS_API_KEY=
 # ATLAS_BASE_URL=https://api.atlascloud.ai/v1
 # ATLAS_MODEL=anthropic/claude-sonnet-4.6
 
+# --- Optional: OrcaRouter (OpenAI-compatible, 150+ LLM models) ---------------
+# Single gateway endpoint + one sk-orca-* key for OpenAI, Anthropic, Google,
+# DeepSeek, Qwen and more. Docs: https://www.orcarouter.ai
+ORCAROUTER_API_KEY=
+# ORCAROUTER_BASE_URL=https://api.orcarouter.ai/v1
+# ORCAROUTER_MODEL=orcarouter/auto
+
 # ─── ECC agent data (multi-harness isolation) ───────────────────────────────
 # Memory hooks (sessions, learned skills, aliases, metrics). Default: ~/.claude
 # Use a separate root when running ECC in Cursor alongside Claude Code:
--- a/README.md
+++ b/README.md
@@ -470,6 +470,14 @@ export ANTHROPIC_AUTH_TOKEN=your-token
 claude
 ```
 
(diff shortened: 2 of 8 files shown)


chenhz01:
Since the default `orcarouter/auto` is a virtual router, the concrete vendor model is only known after the response arrives — which makes the adapter's usage records unattributable: any token/cost accounting downstream of `src/llm` logs the requested `orcarouter/auto` rather than the model that actually served the request, and two identical sessions can price differently. Suggest capturing the echoed model id from the chat-completions response and writing it into the usage record:
```python
usage["resolved_model"] = resp.get("model", requested)
```
That keeps `orcarouter/auto` usable while preserving per-model cost visibility, and matches what concrete-model requests already produce. Happy to open a PR with working code if useful.

PR #1125, opened 2026-04-01
feat(commands): add bundle system for skill lazy-loading

SalvatierraPM:
## Summary

- Adds **15 thematic bundles** in `commands/bundles/` that organize 170+ skills into on-demand groups loaded via Claude Code's slash command system (`/bundles:frontend`, `/bundles:python-stack`, etc.)
- Adds a **skill-router** command (`commands/skill-router.md`) that reads the user's task, picks the minimum bundles needed (typically 1-2), and invokes only those — keeping context lean instead of loading all skills upfront
- Bundles cover: `ai-tooling`, `content`, `data-media`, `devops`, `domain-ops`, `frontend`, `harness-admin`, `obsidian`, `orchestration`, `planning`, `python-stack`, `quality`, `research`, `session-mgmt`, `typescript-stack`

## Problem

When you have 170+ skills, loading them all into context is wasteful and hits token limits. Most tasks only need 5-15 skills from 1-2 domains.

## Solution

**Bundle = thematic index of skills + agents + MCP servers + workflow protocols.**

Each bundle is a markdown command file with:
1. Curated list of skills relevant to the domain
2. Recommended agents (with model tier)
3. MCP servers to activate
4. Domain-specific checklists or workflows
5. `$ARGUMENTS` passthrough so the task flows to the loaded skills

The **skill-router** acts as the entry point — it reads the task, matches it to the bundle registry table, and invokes only what's needed. Users can also invoke bundles directly (`/bundles:frontend`) when they know what they need.

## How to use

```
# Let the router pick bundles automatically
/skill-router Build a landing page with GSAP animations

# Or invoke a bundle directly
/bundles:frontend Build a landing page with GSAP animations
/bundles:quality Review this PR before merge
/bundles:python-stack Fix the Flask auth middleware
```

## Creating custom bundles

Users can create project-specific bundles (e.g., `bundles/my-project.md`) following the same pattern — list skills, agents, MCP servers, and pass `$ARGUMENTS`.

## Test plan

- [ ] Verify each bundle file loads correctly as a slash command
- [ ] Test skill-router routing with sample prompts
- [ ] Confirm no project-specific or private content in any bundle

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Adds a bundle system for lazy-loading skills and a `skill-router` that auto-picks 1–2 bundles per task to cut context and token use. Bundles now live under `skills/bundles/`; fixed the `perl` bundle by removing a non-existent reviewer reference.

- **New Features**
[...]

PR #2788, opened 2026-08-14
feat: slim, receipted profile plugin carriers for the plugin install path

montjeffrey:
## Summary

ECC's selective-install profiles (`manifests/install-profiles.json`) currently only serve the standalone installer path — the **plugin** path always injects the full surface (~278 skills, 67 agents, 95 commands of frontmatter) into every session. This PR materializes the profile system into real, choosable plugin *carriers* — generated, staged, and receipted rather than just copied.

**Reworked after review** to treat context and capabilities as separate decisions, fail closed on incomplete runtime dependencies, keep the carrier self-contained and provenance-safe, and enforce a labelled token budget — see the response comment on this PR for the point-by-point mapping. The offline skill router that shipped with the original version of this PR is now split into its own opt-in PR, #2945, since it's a separate behavioral feature.

### What's included

- **`scripts/plugin-profiles.js` + `scripts/lib/plugin-profiles.js`** — generates slim plugin carriers from the existing selective-install manifests (module-level via `"modules"` or component-level skill/agent/command lists, with profile `extends`). Each carrier is a valid standalone plugin in a local marketplace; a project opts in via `enabledPlugins` in its `.claude/settings(.local).json`.
- **`commands/plugin-profiles.md`** — `/plugin-profiles` command to list, plan, generate, and activate profiles. `generate` is dry-run-first and surfaces the hook-capability decision explicitly instead of defaulting it.
- **`--hooks <off|minimal|standard|strict>`** — the hook runtime is never carried silently; a pending decision refuses generation, reusing the existing installer's capability-disclosure text.
- **Staged, receipted generation** — build to a staging directory, verify the runtime closure, atomic swap, restore on failure; ownership for overwrite requires the receipt *and* a matching content digest; `--dry-run`, `--force --yes`, `--keep-prev` all supported.
- **Self-contained carrier** — no absolute source path anywhere in the output; on-demand skills are copied into `on-demand/<id>/` and content-addressed (sha256 per catalog row).
- **Labelled, enforced token ledger** — the `name: description` listing payload is measured with a named method (`chars-per-token-estimate@1`, disclosed as an estimate) and gated against `--budget` (default 8000), fail-closed unless `--allow-over-budget`.
- **Docs** — `docs/PLUGIN-PROFILES.md` (design rules, receipt schema, ledger table, limitations).
- **Tests** — 50 tests across `tests/lib/plugin-profiles.test.js` (46) and `tests/hooks/run-with-flags-user-prompt.test.js` (4, pinning that UserPromptSubmit hooks never echo raw stdin into context when disabled).

### Real-world measurement

Running daily on a Windows 11 workspace (8 projects opted into three generated profiles): a minimal-profile session start measured **7,170 fewer context tokens** than full ECC (26,986 → 19,816 cache-creation on an otherwise identical session), with the slim variant's skills confirmed loaded via the session init payload. The saving repeats for every session and every spawned subagent. (Measured before this review pass; the underlying mechanism — a slim carrier instead of the full listing — is unchanged by the rework.)

### Not in this PR

- The skill router — split into #2945 (opt-in, time-bounded, evidenced with a precision/recall eval).
- Binding to a canonical `lean@1`/`full@1` context-profile registry — not public anywhere in this repo yet; the receipt records exact inputs and a content digest so binding is a small follow-up once the schema exists.

## Test plan

- [x] `node tests/lib/plugin-profiles.test.js` — 46/46
- [x] `node tests/hooks/run-with-flags-user-prompt.test.js` — 4/4
- [x] `scripts/ci/validate-hooks.js` — 23 matchers (down from 24 now that the router hook moved to #2945), `validate-commands.js` (95 files), `validate-no-personal-paths.js`, `check-unicode-safety.js`, `npm run command-registry:check` — all pass
- [x] `node scripts/plugin-profiles.js generate --profile opencode --out <tmp>` — the generated carrier's `scripts/skills-health.js --help` and `scripts/plugin-profiles.js list` both run standalone (closes the Greptile `/skill-health` finding)
- [x] Merged current `main`, no conflicts

https://claude.ai/code/session_01Xj2iuYbuWqrB7eYYYVAfSp


haelyra:
Thank you for building a concrete answer to the plugin context-cost problem. Generating a narrow Claude carrier before session assembly is the right layer, and the deterministic install-plan projection plus offline routing experiments are valuable inputs to ECC’s M1 profile work.

This should be ported as an adapter around the canonical profile contract rather than merged as a parallel plugin-profile system:

- **Generate only from versioned `lean@1` and `full@1` context manifests.** The current CLI reuses broad install profiles such as `developer` and `minimal`, but install selection is not context selection. Success is a carrier whose exact skill and command set comes from one canonical context-profile digest, while module and component switches remain compiler inputs rather than new base-profile authorities.
- **Separate context from capabilities.** The generator includes the full hook runtime by default because hooks have “zero context cost,” but a narrow context profile does not authorize lifecycle automation or its filesystem/process capabilities. Require a distinct capability-profile decision and receipt for hooks; a Lean carrier must not silently inherit Full automation.
- **Make generation bounded, receipt-backed, and recoverable.** `generate` has no dry-run requirement, deletes the target recursively, and treats a spoofable one-field `ecc-profile.json` marker or `--force` as sufficient ownership proof. Resolve and bound the destination, preview the exact tree and deletions, require digest confirmation for replacement, write through a staging directory, and keep an immutable before-image/receipt for rollback and interrupted-run recovery.
- **Fail closed on incomplete runtime closure.** `resolveScriptClosure()` silently skips unresolved relative imports while the docs promise generated commands will never be broken. Report every unresolved static and dynamic dependency, and refuse generation unless the carrier passes a real install/load/command smoke test from the staged artifact.
- **Make the carrier self-contained and provenance-safe.** The generated catalog embeds an absolute checkout path and tells the model to read arbitrary on-demand `SKILL.md` files from that mutable tree “as if installed.” That leaks a local path, bypasses the approved profile digest, and lets later source changes alter behavior without activation. Copy approved on-demand content into a content-addressed carrier or require a new profile activation; never load unreceipted source-tree instructions.
- **Use a truthful provider ledger and keep routing optional.** Four-characters-per-token estimates do not satisfy the hard Lean budget, and the UserPromptSubmit router is a separate behavioral feature that injects suggestions on every matching prompt. Measure the actual Claude catalog payload with a versioned method, enforce the declared budget, and land routing later as an opt-in adapter with precision/recall and latency evidence.

The strongest salvage path is to port the deterministic carrier builder, dependency-closure tests, and local marketplace projection into the existing M1 `lean@1`/`full@1` train, preserving your commits and attribution. If you do not have bandwidth for that port, we will queue those pieces after the next release rather than ask you to rebuild the whole profile foundation here.

montjeffrey:
Thanks — I've reworked the branch along these lines rather than leaving it for the post-release queue. Pushed as one update to this PR, plus a stacked PR for the router: #2945.

What changed, mapped to your six points:

1. **Canonical `lean@1`/`full@1`** — I couldn't find the registry, digest, or receipt schema on `main`, in `docs/`, or in any open branch/PR, so I didn't guess one. The CLI still takes install-profile ids as *inputs*; the receipt records the exact inputs and a content digest, so binding is a small follow-up once I can see the schema. I'd rather do that port myself than have it queued — if you can point me at the schema (or the branch it lives on), I'll bind this branch to it directly.
2. **Context vs. capabilities** — hook runtime is never carried without `--hooks <minimal|standard|strict>` or `--hooks off`. A pending decision refuses generation using the #2715 disclosure (imported from `hook-consent.js`, not duplicated). The decision, profile, and capability groups are in the receipt; the profile is pinned through the existing `ecc/setup.json` managed-config path.
3. **Bounded, receipted, recoverable** — staged build, verify, atomic swap, restore on failure; plugin name re-validated and target bounded to `--out` before any delete; ownership requires the receipt *and* a matching tree digest (a copied-in marker or a hand-edited carrier is refused); `--force` needs `--yes` when non-interactive; `--dry-run` prints the exact copy list, whether the target is replaced, the ledger, and every blocker; `--keep-prev` parks the replaced tree; `previous` is recorded.
4. **Fail closed on closure** — dependencies are derived from each shipped command's body, the require graph is walked (including wholesale-copied dirs), unresolved static requires abort with `file -> specifier`, dynamic requires are reported, and the staged tree is re-verified. Your `/skill-health` repro now runs from generated minimal and opencode carriers, tested directly rather than just asserted.
5. **Self-contained** — no source path anywhere; on-demand skills are copied into `on-demand/<id>/` with sha256 per catalog row; the "as if installed" instruction is gone; catalog rows are flattened so a description can't forge a row.
6. **Ledger and router** — the ledger measures the actual `name: description` listing payload with a labelled method (`chars-per-token-estimate@1`). You're right that an estimate doesn't satisfy a hard Lean budget — it's labelled and versioned rather than presented as ground truth, and generation refuses over a declared `--budget` (default 8000) unless `--allow-over-budget`. A provider-backed measurer is the direct next step, opt-in so the CLI itself stays network-free. The router is split into #2945: off by default, time-bounded, catalog text sanitized against injection, with a 52-prompt eval (precision@3 0.96, warm p50 2.9ms). I saw `WORKING-CONTEXT.md`'s note that #1125 was closed as a second routing abstraction — if that judgement still stands, I'd rather have that PR declined on the record than merged half-on; this PR doesn't depend on it either way.

Honest numbers: with the estimate, `opencode` is ~6.6k, `minimal` ~10.5k, `developer` ~15.6k. `minimal` is over the 8k default because `commands-core` ships all 95 commands and `agents-core` all 68 agents — a manifest-granularity limit I've documented rather than hidden, and the next thing I'd fix.

Tests: `tests/lib/plugin-profiles.test.js` 46/46, `tests/hooks/run-with-flags-user-prompt.test.js` 4/4; validators and lint green.

montjeffrey:
Two fixes for the issues Greptile flagged after the last push, plus a follow-up on point 1 above.

**Fixed and pushed:**

- **Symlink integrity gap** (0d0df38, a9aadcd): selected sources are now swept for symlinks - the main copy loop and the on-demand catalog-copy path both, since they share the same operations list - and generation refuses with the exact offending path if any are found. No bypass flag; this is unconditional. 49/49 in `tests/lib/plugin-profiles.test.js` (was 46).
- **Router budget enforced during the scan, not just after it** (07b37e78 on #2945): `readCatalog()` now checks a deadline before reading each `SKILL.md` and stops early once it passes, instead of always finishing an unbounded scan and only discarding the printed output afterward. A truncated scan is never written to the catalog cache. I want to be precise about the guarantee this actually gives: it bounds the overrun to roughly one file's read past the budget, not to zero - a real, meaningful bound, but not a hard real-time guarantee. `docs/SKILL-ROUTER.md` now says that plainly instead of the "never delays prompt submission" claim that was there before. Also fixed: `scripts/ci/skill-router-eval.js`'s `--min-precision`/`--min-recall` flags silently no-opped on a malformed value (`Number(undefined)` -> `NaN`, every threshold comparison false) - they now fail fast with a clear message instead.

**Still open - point 1, canonical `lean@1`/`full@1`:** I searched main, `docs/`, and every open branch/PR I can see and found nothing, same as before. Your comment above mentions "the existing M1 lean@1/full@1 train" - if that lives on a branch or repo I don't currently have read access to, could you add me as a collaborator or point me at it directly? I'd still rather do that binding myself than have it queued.

montjeffrey:
Pushed a fix for the carrier integrity findings Greptile raised on #2945 (they surface there because that branch is stacked on this one, but the code is this PR's).

**Symlinks bypassed carrier ownership validation.** `isGeneratedProfilePlugin()` trusted `computeTreeDigest()` alone to decide whether an existing carrier was an unmodified tree this generator wrote. That digest hashes what `listFilesRecursive()` enumerates, and that counts only `entry.isFile()` — so a symlink planted into a carrier *after* generation is never enumerated, never hashed, and the digest still reports "unmodified". The carrier then serves whatever the link resolves to, from outside the receipted, content-addressed tree, while ownership validation keeps accepting it and regeneration proceeds without a blocker.

Generation already refused a symlinked *source* (`previewProfilePlugin` → `findSymlinksUnder` over the plan's copy operations). This applies the same rule at the other end — an existing carrier on disk, which is the tampered-install / untrusted-source case. It reuses `findSymlinksUnder` rather than teaching the digest walker link semantics: a tree we can't fully account for isn't one we own, so it's never deleted or replaced without `--force`. Fails closed.

Verified end to end rather than just at the boolean: with a symlink planted into a freshly generated carrier, ownership returns false, regeneration refuses with `Refusing to overwrite` instead of delete-and-replacing the tree, and the file the link pointed at outside the carrier is left intact. The overwrite blocker's message named three causes and not this one, which would send someone debugging in the wrong direction — it now names the symlink case too.

Two RED→GREEN tests cover a planted file symlink and a planted directory symlink; both fail against `b7c89948` and pass here. They use a `trySymlink()` helper that skips only on genuine platform-capability errors (`EPERM`/`EACCES`/`ENOSYS`/`ENOTSUP`/`EOPNOTSUPP`) and rethrows everything else — a bare catch silently "passed" one of these tests on first write, when a missing `skills/` parent surfaced as `ENOENT` rather than a platform limit.

`node tests/run-all.js`: 4234 total, 4233 passed, 1 failed (the pre-existing `hooks/hooks.test.js` `observe.sh` ENOENT, unrelated). eslint clean; `catalog:check` and `command-registry:check` in sync.

---

**Separately, on #3037.** I have a local integration branch that binds this carrier to the context-profile contract in #3037, and I'd rather disclose it than have it appear fully formed later:

- The receipt's `contextProfile` block carries the real `registryDigest`/`profileDigest`/`compilerDigest`/`planDigest` chain instead of the current `install-profiles@unbound` / `null` placeholders (receipt schema v1 to v2, v1 still recognized for `--force`).
- The carrier's **skill** scope comes from `compileContextProfile()`'s `selectedIds`/`routedIds` rather than being derived independently here. Agents and commands stay install-plan-derived, since #3037 excludes those surfaces by design; the receipt records that split explicitly as `surfaceAuthority: {skills: "context-profile", agents: "install-plan", commands: "install-plan"}`.
- This carrier's own `--budget` gate stays the enforcement point. #3037's `estimate` is recorded alongside it as advisory, never substituted for it — verified by generating with the default 8000-token budget and watching the carrier correctly refuse at 8638 tokens, then re-running with `--allow-over-budget` and confirming 3 skills shipped / 284 routed, matching `lean@1`'s manifest exactly rather than the install profile's own skill list.

It is **not** pushed here on purpose: #3037 is still open, so pushing it would drag that PR's entire unmerged tree into this diff. It's waiting on #3037 to land, and it's a rebase away from being a real commit here. Happy to open it as a separate PR instead if that's easier to review, or to drop it entirely if the binding direction isn't what #3037 wants.

---

@affaan-m — this is the carrier-side half of the two Greptile P1s on the stack; the router-side half (the runtime read-path symlink guard, plus a routing-budget bound) landed in #2945. Both are ready for another look.

@haelyra — flagging you on the #3037 section above, since that integration builds directly on your compiler contract and records its digests in this carrier's receipt. If the binding direction is wrong for where you're taking #3037, I'd rather hear it now than after it's a PR.

chenhz01:
This thread is already thorough, so one failure path I don't see mapped: The atomic swap's restore path can itself fail and leave no target: in `swapIntoPlace`, if `renameSync(stagingRoot, pluginRoot)` throws, the restore `renameSync(previousRoot, pluginRoot)` runs bare — on Windows an open handle inside the parked tree (editor, indexer, antivirus) makes that rename throw too, the new error replaces the original, and the operator is left with neither tree in place and no pointer to the parked copy. Wrap the restore in its own try/catch, rethrow an aggregated error naming both paths plus the parked-tree location so recovery is manual-but-possible. Same guard answers the partial-state branch where `pluginRoot` exists but the staging move failed — restore is skipped and the parked tree is stranded. Happy to open a PR with working code if useful.

Issue #2790, opened 2026-08-15
install-apply replaces install-state instead of merging: a second --skills install orphans the previous profile install and uninstall leaves files behind


josuecelaya-coder:
## Summary

Each `install-apply.js` invocation **replaces** `install-state.json`'s `operations` array instead of merging with what the previous invocation recorded for the same target.

Because the Claude skill migration path treats a flat skill directory that exists on disk but is *not* recorded in install-state as user-owned, a second install with a different scope silently orphans everything the first install managed. Those files are then skipped by later installs and are never removed by `uninstall`.

This is a two-command trigger, and both commands are ones the docs recommend: `--profile <name>` and `--skills <id>`.

## Repro

Fully self-contained, using a throwaway `HOME`:

```bash
T=$(mktemp -d)
ST="$T/.claude/ecc/install-state.json"
ops() { node -e "console.log(JSON.parse(require('fs').readFileSync('$ST','utf8')).operations.length)"; }

# 1. install one module
HOME="$T" node scripts/install-apply.js --modules skill-unified-memory >/dev/null
echo "after install #1: $(ops) operations"

# 2. install one extra skill
HOME="$T" node scripts/install-apply.js --skills skill-comply >/dev/null
echo "after install #2: $(ops) operations"

# 3. re-run the first install
HOME="$T" node scripts/install-apply.js --modules skill-unified-memory 2>&1 | grep -E "Applied|Skipped|user-owned"
echo "after install #3: $(ops) operations"

# 4. compare disk against what uninstall would remove
echo "files on disk: $(find "$T/.claude/skills" -type f | wc -l)"
HOME="$T" node scripts/uninstall.js --target claude --dry-run | grep "Planned removals"
```

Observed:

```
after install #1: 8 operations
after install #2: 22 operations        <- the 8 from step 1 are gone
after install #3: ...
  Applied operations: 7
  Skipped operations: 1
  - Skipped Claude skill 'unified-memory' at <...>/.claude/skills/unified-memory:
    the flat skill directory is user-owned because it is not recorded in ECC install-state.
after install #3: 7 operations
files on disk: 23
  Planned removals: 8
```

23 managed files on disk, 8 scheduled for removal — **15 orphaned**.

## Why it compounds

The skip in step 3 is `classifySkillGroup()` in `scripts/lib/install/claude-skill-migration.js` behaving as designed:

```js
if (pathExists(flatSkillRoot) && !hasManagedFlatFile) {
[...]

Issue #2789, opened 2026-08-15
install-modules.json still misses skills/skill-comply — --profile full installs 280 of 281 skills (remainder of #2431)


josuecelaya-coder:
## Summary

`skills/skill-comply/` is not referenced by any module in `manifests/install-modules.json`, so `install-apply.js --profile full` never copies it. Every other curated skill is referenced.

This is the last remaining instance of #2431 ("install-modules.json is missing 79 curated skills"), which was closed as completed on 2026-08-04. That fix covered 78 of the 79; `skill-comply` still slips through.

Verified against the `ecc-universal@2.1.0` release tree (`VERSION` = 2.1.0).

## Repro

```bash
node -e "
const fs=require('fs');
const d=JSON.parse(fs.readFileSync('manifests/install-modules.json','utf8'));
const ref=new Set();
const walk=o=>{
  if(!o)return;
  if(typeof o==='string'){const m=o.match(/^skills\/([^\/]+)/); if(m)ref.add(m[1]); return;}
  if(Array.isArray(o))return o.forEach(walk);
  if(typeof o==='object')return Object.values(o).forEach(walk);
};
walk(d.modules||d);
const dirs=fs.readdirSync('skills',{withFileTypes:true}).filter(e=>e.isDirectory()).map(e=>e.name);
console.log('skill dirs:', dirs.length);
console.log('referenced :', ref.size);
console.log('unreferenced:', dirs.filter(x=>!ref.has(x)));
"
```

Output:

```
skill dirs: 281
referenced : 280
unreferenced: [ 'skill-comply' ]
```

## Impact

A `--profile full` install silently yields 280 of 281 skills. The gap is invisible from the install output, and `npm run catalog:check` still passes because it compares documentation counts against the repository catalog (281/94/67), not against manifest coverage:

```
Catalog counts:
- agents: 67
- commands: 94
- skills: 281

Documentation counts match the repository catalog.
```

Workaround: `node scripts/install-apply.js --skills skill-comply` installs it. Note that this currently has a side effect — see the companion issue about `install-state` being replaced rather than merged.

## Expected

Either add `skills/skill-comply/` to an appropriate module in `manifests/install-modules.json`, or, if the omission is intentional (it is a meta/measurement skill), document the exclusion so it is distinguishable from a sync miss.

## Suggested follow-up

[...]

Issue #2431, opened 2026-07-03
install-modules.json is missing 79 curated skills (including all 6 from the latest wave) — ecc-install --profile full silently skips them


Guykaganovsky1:
## Summary

79 of the 277 curated skill directories under `skills/` are not referenced by any module in `manifests/install-modules.json`, so a full-profile install never copies them. This includes all six skills added in the most recent wave: `delivery-gate`, `ecc-recipes`, `growth-log`, `loop-design-check`, `mailtrap-email-integration`, `react-native-patterns`.

This contradicts [docs/SKILL-PLACEMENT-POLICY.md](docs/SKILL-PLACEMENT-POLICY.md), which states curated skills are "Included in `manifests/install-modules.json` paths."

Verified at commit `81af4076` (which is itself a `chore(catalog): sync manifests` commit — the sync flow still leaves these 79 behind).

## Repro

```bash
# count skill dirs vs manifest references
python3 - <<'EOF'
import json, os
d = json.load(open('manifests/install-modules.json'))
referenced = {p.rstrip('/').split('/')[1]
              for m in d['modules'] for p in m.get('paths', [])
              if p.startswith('skills/')}
on_disk = {s for s in os.listdir('skills') if os.path.isdir(f'skills/{s}')}
print(len(on_disk), 'on disk |', len(referenced), 'referenced')
print(sorted(on_disk - referenced))
EOF
```

Output: `277 on disk | 198 referenced` + the 79 names below.

A full-profile install (verified 2026-07-03 into `~/.claude`) confirms it end-to-end: the dry-run plan contains no ops for e.g. `skills/delivery-gate`, and the skill is absent after install.

## The 79 unreferenced skills

<details><summary>Full list</summary>

[...]

PR #2784, opened 2026-08-13
fix(install): harden ECC installer lifecycle

haelyra:
## Emergency installer hotfix

This consolidates the recent installer failures into one reviewed fix and adds a permanent exact-package release gate.

### Fixes

- native Antigravity 2.0 installation with conservative legacy migration
- transform-aware doctor and repair convergence
- canonical JSON-to-SQLite install-health projection
- Python bytecode exclusion
- transactional, reversible legacy Codex sync across custom hooks roots, with rollback armed before the first ownership write
- portable no-follow descriptor I/O, Windows reparse-point rejection, identity verification, and inode-verified quarantine removal
- transactional repo-scan bootstrap guidance
- publish and PR gates using one exact packed artifact on Linux, macOS, and Windows
- release lifecycle runner carried as a verified artifact with no dynamic privileged checkout

### Proof completed at current head 0a7d2787

- 3,918 repository tests passed, 0 failed on native macOS
- focused installer, migration, rollback, and lifecycle suites passed on native macOS and Linux Node 22
- lint and workflow security validation passed
- runtime npm audit found 0 vulnerabilities
- current exact local tarball SHA-256 2d150d16eb05ce501b1be22e97b0ce8c6f74a1cda07977c0fc0db34d3cd9fcc7
- the current exact tarball passed the full public-bin lifecycle on native macOS: npm install, setup help, install, repeat install, clean doctor, status, induced drift, failed doctor, repair, clean doctor, repaired status, uninstall, uninstalled status, and sentinel preservation
- current-head CodeQL is clean
- current-head hosted exact-artifact lifecycles are required on Ubuntu, macOS, and Windows, together with the full Node and package-manager matrix

### Merge policy

Do not merge until CodeQL and the current-head Packed Install jobs pass on GitHub-hosted Ubuntu, macOS, and Windows, together with the existing full CI matrix.

Closes #2696
Closes #2727
Closes #2750
Closes #2774

Incorporates and preserves contributor credit from #2728, #2752, #2757, #2775, and #2780.

PR #2794, opened 2026-08-15
fix(manifests): reference skills/skill-comply so --profile full installs it

ntdatt812:
Refs #2789.

## What happens

`skills/skill-comply/` is not referenced by any module in `manifests/install-modules.json`, so `install-apply.js --profile full` never copies it. Every other curated skill is referenced.

Running the reporter's check on current `main`:

```
skill dirs: 284 | referenced: 283
unreferenced: [ 'skill-comply' ]
```

After this change:

```
skill dirs: 284 | referenced: 284
unreferenced: []
```

(The counts have moved from the 281 in the issue as skills were added; the gap is the same one.)

This is the last remaining instance of #2431 ("install-modules.json is missing 79 curated skills"), which closed as completed after covering 78 of the 79.

The gap is invisible from the install output, and `npm run catalog:check` does not catch it — it compares documentation counts against the repository catalog (68/94/284), not against manifest coverage, so it passes either way.

## Placement

`workflow-quality`, alphabetically between `production-audit` and `skill-scout`.

`skill-comply` measures whether agents actually follow the skills, rules and definitions they were given — an agent-quality concern, so it sits with `agent-introspection-debugging`, `ai-regression-testing` and `skill-stocktake` rather than in `security`, whose compliance skills (`hipaa-compliance`, `healthcare-phi-compliance`) are about regulatory domains.

## Checks

- reporter's repro script: 284 dirs / 284 referenced, empty unreferenced list
- `node scripts/ci/validate-install-manifests.js` — validated 34 install modules, 81 install components, 7 profiles
- `node scripts/ci/validate-skills.js` — validated 284 skill directories
- `npm run catalog:check` — documentation counts match the repository catalog

One line, no behaviour change beyond the skill now being installed by the profiles that carry `workflow-quality`.

Note this does not address the companion issue #2790 (`install-apply` replacing rather than merging `install-state`); it only removes the reason the `--skills skill-comply` workaround was needed.

### DIFF
--- a/manifests/install-modules.json
+++ b/manifests/install-modules.json
@@ -326,6 +326,7 @@
         "skills/plan-canvas",
         "skills/plankton-code-quality",
         "skills/production-audit",
+        "skills/skill-comply",
         "skills/skill-scout",
         "skills/skill-stocktake",
         "skills/strategic-compact",


ntdatt812:
Following up on the bot review — it found a real gap, but I think it points at a question for you rather than at a change I should make unilaterally.

The bot is right that the packaged surface does not carry this skill: `package.json` `files` enumerates 283 skills individually and `skills/skill-comply/` is not among them, so a package-based install would resolve a `workflow-quality` path that the tarball never shipped.

Where I stopped short of "just add it": `tests/scripts/npm-publish-surface.test.js` asserts the exclusion **on purpose**.

```js
for (const excludedPath of [
  "contexts/dev.md",
  "examples/CLAUDE.md",
  "plugins/README.md",
  "scripts/ci/catalog.js",
  "skills/skill-comply/SKILL.md",
]) {
```

`git log -S` puts that line in c9de8f5 (#2784, `fix/installer-hotfix-2.2`) — the most recent merge on `main`. That reads as a deliberate and very fresh decision, not an oversight, so flipping it to make my change consistent felt like the wrong call to make on your behalf.

So the two readings are:

1. **The manifest omission is the bug** (#2789's premise). Then this PR is right and `package.json` + that test expectation should follow, so npm installs ship the skill.
2. **The exclusion is intentional** — skill-comply is dev-only tooling (22 files, 111K, its own `pyproject.toml`, `fixtures/`, `tests/`) that you chose to keep out of the published package in the installer hotfix. Then #2789's premise is outdated, referencing it from `workflow-quality` is wrong for packaged installs, and this PR should be closed.

I lean towards (2) given the timing and that skill-comply carries a Python test harness rather than being a plain skill, but you own that decision.

Happy to push the `package.json` + test update if you want (1), or close this out if (2). Sorry for the extra round trip — I should have checked the publish surface before opening.

ntdatt812:
Closing this — I dug further and I think the premise of #2789 is wrong, so the change would be a regression rather than a fix.

`skills/skill-comply/SKILL.md` is listed in `tests/scripts/npm-publish-surface.test.js` alongside `contexts/dev.md`, `examples/CLAUDE.md`, `plugins/README.md` and `scripts/ci/catalog.js` — everything else in that list is repo-only tooling that deliberately does not ship in the published package. skill-comply is also the only entry under `skills/` carrying a `pyproject.toml`, with its own `fixtures/`, `prompts/`, `scripts/` and `tests/`; it reads as a Python measurement harness for the repo rather than a skill meant to land in a user's install.

Read that way, the 283-of-284 gap is intentional, not the last straggler from #2431, and adding it to `workflow-quality` would have profile installs reference a path the npm package does not carry.

Thanks to the review bot for catching the packaging side — that is what sent me back to look properly. Sorry for the noise, and thanks for the patience.

Issue #2090, opened 2026-05-29
Ran 7 stop hooks (ctrl+o to expand)     ⎿  Stop hook error: JSON validation failed


ccloud1997:
Ran 7 stop hooks (ctrl+o to expand)
    ⎿  Stop hook error: JSON validation failed

affaan-m:
Thanks for reporting this. The key detail is the expanded Stop hook output behind `ctrl+o`.

`Stop hook error: JSON validation failed` usually means one Stop hook emitted malformed JSON, partial JSON, or a normal log line on stdout where the host expected a hook response object. The current hook path should keep human-readable logs on stderr and only emit structured JSON when the hook contract requires it, so this is actionable if we can identify the hook.

Could you add:

1. the expanded list of the 7 Stop hooks;
2. your install target and version (`ecc-install --version` or package/tag used);
3. the relevant `hooks.json` Stop entries with personal paths redacted;
4. whether this came from Claude Code plugin install, manual install, Cursor/OpenCode adapter, or a copied config.

Immediate workaround: disable the specific Stop hook that emits the invalid payload, or update to the current `ecc-universal@next` rc if you are on an older hook bundle. I will keep this open until we can tie it to a specific hook script.

IZRINO:
PR #2104 includes hook JSON/stdout safety coverage and the full hook test suite is passing locally. This issue still needs the expanded Stop hook output behind ctrl+o to identify the exact malformed emitter; without that, I did not guess at a specific Stop hook. Local verification: `node tests/hooks/hooks.test.js` passed 221/0 and `npm test` passed with Windows-only shell integration skips.

PR #2796, opened 2026-08-15
fix(hooks): never echo stdin on strict-output Stop events

Noah-TaeHwan:
## Summary

On current Claude Code (observed on 2.1.231–2.1.233), **every turn stop with a disabled or no-opinion Stop hook surfaces `invalid stop hook JSON output`**. The root cause is ECC's passthrough convention (echo stdin on stdout when a hook has no opinion) colliding with the harness's per-event stdout contract: Stop stdout is parsed strictly as hook-output JSON, so an echoed Stop input payload is invalid output at **any** size. #2222/#2227 suppressed only the truncated-echo case (>1MB); sub-cap payloads still echoed on every turn stop.

This PR makes the harness-facing boundary **event-aware**: for strict-output events (`Stop`, `SubagentStop`) "no opinion" is empty stdout, never an echo. All other events keep the existing passthrough design untouched.

## Symptom

With `ECC_DISABLED_HOOKS=stop:desktop-notify` (or any Stop hook returning no opinion), Claude Code shows on every response:

```
Stop hook (failed)
  error: hook returned invalid stop hook JSON output
```

Reproduction (before this PR):

```bash
echo '{"session_id":"x","transcript_path":"/tmp/t.jsonl","hook_event_name":"Stop","stop_hook_active":false}' \
  | ECC_DISABLED_HOOKS=stop:desktop-notify CLAUDE_PLUGIN_ROOT="$PWD" \
    node scripts/hooks/run-with-flags.js stop:desktop-notify scripts/hooks/desktop-notify.js standard,strict
# → exits 0 but echoes the input payload; the harness rejects it as Stop output.
# After this PR: exits 0 with empty stdout.
```

## Root cause (three layers, all at the harness boundary)

1. **`scripts/hooks/run-with-flags.js`** — every no-opinion fallthrough (disabled hook, dry-run, missing args, missing script, `run()` error) and `resolveHookResult`'s no-opinion branch echo `raw`.
2. **`scripts/hooks/plugin-hook-bootstrap.js`** — `passthrough()` echoes `raw` when the child produced no stdout, and the missing-args / unknown-mode / spawn-error fallbacks echo `raw`.
3. **`hooks/hooks.json`** — the Stop lifecycle wrappers' unresolvable-root fallback calls `finish(raw, …)`; the SessionEnd fallback does `process.stdout.write(raw)`.

Additionally, several Stop hook scripts follow the legacy echo-stdin convention when invoked directly; the runner previously forwarded that echo to the harness.

## What changed

- **`run-with-flags.js`**: introduces `STRICT_OUTPUT_EVENTS = {Stop, SubagentStop}` (detected from `hook_event_name` in the stdin payload). For those events, `sanitizeEcho` suppresses any stdout that is byte-identical to the stdin payload — covering both the runner's own fallthroughs and hook scripts that echo by convention. Real hook output (which always differs from the input payload) is forwarded untouched. The oversized-payload (#2222) behavior is preserved verbatim.
- **`plugin-hook-bootstrap.js`**: same contract for the bootstrap's child-stdout forwarding and its three fallback paths.
- **`hooks/hooks.json`**: Stop wrapper fallbacks `finish('')` instead of `finish(raw, …)` (stderr warnings kept); the SessionEnd fallback no longer echoes stdin (harness ignores SessionEnd stdout).
- **Tests**:
  - `tests/hooks/stop-hooks-stdout.test.js` (#2090): the "registered wrapper" and dry-run expectations now assert **empty stdout** at the boundary. The previous expectation (echo the payload uncut) is precisely the output current Claude Code rejects. Direct-invocation expectations (script-level echo conventions) are unchanged.
  - `tests/hooks/run-with-flags-truncation.test.js`: adds a disabled-Stop regression test (the exact live failure) and an opinionated-output preservation test (`additionalContext` survives); all pre-existing passthrough expectations for non-strict events are untouched.

## What deliberately did NOT change

- The passthrough design for PreToolUse-family events (e.g. GateGuard's allow-echo, config-protection passthrough, dry-run echo) — those harness events tolerate it and a large body of hooks/tests depends on it.
- Hook scripts' internal echo conventions on direct invocation — the boundary suppresses them for strict events, so scripts need no migration.

## Verification

- `tests/hooks/run-with-flags-truncation.test.js`: 9/9
- `tests/hooks/stop-hooks-stdout.test.js`: 40/40
- `tests/hooks/plugin-hook-bootstrap.test.js`: 14/14, `hook-flags`: 63/63, `gateguard-fact-force`: 144/144, `config-protection`: 9/9, `lib/dry-run`: 12/12
- Full `node tests/run-all.js`: 3914/3918 (the 4 failures are pre-existing `lib/state-store` environment failures on my machine, identical before this change)
- Live check on an installed plugin (Claude Code 2.1.233): disabled Stop hook, enabled no-opinion Stop hook, and the full registered wrapper command all exit 0 with empty stdout; PreToolUse passthrough and GateGuard behavior unchanged.

## Related

- #2090 (Stop stdout flush contract — expectations updated here), #2222 / #2227 (truncated-echo suppression — preserved), #2493 (Stop output through lifecycle wrappers — real output still passes through).

Made with [Orca](https://github.com/stablyai/orca) 🐋


Noah-TaeHwan:
Closing as conflicting. This was skipped in #2870 for a rebase, and I am not keeping the branch current.

PR #17, opened 2026-01-22
Set up plugin marketplace structure

fusionfx1:
Sets up repository structure for plugin marketplace. Moves all plugin content to plugins/everything-claude-code/ and adds marketplace.json with proper schema.

## Summary by CodeRabbit

* **New Features**
  * Registered the "everything-claude-code" plugin for enhanced agents and skills.

* **Chores**
  * Updated plugin metadata and added a top-level schema and description for clearer registry info.
  * Switched plugin source to reference its GitHub repository.
  * Added a plugin configuration entry including author information and basic metadata; removed legacy homepage/repository/license/tags fields.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2799, opened 2026-08-15
docs(readme): add star-history badges and restore star history chart

haelyra:
Preview PR. README-only plus two generated chart assets.

## 1. Badges moved to the very top

The star-history badges now sit **above the language selector**, directly under the hero image:

- **GitHub Trending — #1 Repository of the Day**
- **star-history.com — Global Rank #17**

Both verified live (HTTP 200, real values, no redirect). Theme-aware via `<picture>`.

## 2. The live star history chart is dead, and not just for us

The reason the chart never appeared: **`api.star-history.com/svg` no longer returns a chart for anyone.** It returns a placeholder SVG that reads "GitHub restricted access to star data".

GitHub restricted the stargazers API on **June 30, 2026** to a repo's own admins and collaborators. Consequences:

- The public embed is broken for every repo, ours included.
- star-history.com's own chart page is now gated behind "paste a GitHub access token".
- Even with a token, star-history could not chart ECC fully: GitHub caps stargazer pagination at **40,000 records**, and ECC has **240,265 stars**. Pages past 400 return `"pagination is limited for this resource"`.

So there is no existing star history image to find, on star-history.com or anywhere else. It cannot currently be generated by that service.

## 3. What is in the PR instead

A chart generated from data we *can* read (collaborator access to the stargazers API), committed as static SVG so it never breaks:

- `assets/star-history-light.svg` / `assets/star-history-dark.svg` — theme-aware, placed above `## Install ECC`
- `assets/star-history-data.tsv` — the 35 sampled data points, so the chart is reproducible

It covers the **first 40,000 stars, Jan 18 to Feb 7, 2026** and is labeled as such on the chart itself. That is a real, accurate curve of the launch ramp, but it is **17% of the current total** and stops six months ago.

## Decision needed

The 40k chart is honest but understates the repo. Options:

1. **Ship the 40k chart as-is** (what this PR does), clearly labeled.
2. **Drop the chart**, keep just the two badges. They work, they are current, and "#17 globally / #1 of the day" is arguably a stronger signal than a truncated curve.
3. **Reconstruct the full curve** from another source. Checked and rejected so far: ClickHouse's public `github_events` (incomplete — reports 40k stars by May, GitHub says 40k by February), Wayback CDX (503 during this work, worth retrying), README git history (only rounded "50K+" marketing numbers that sat stale for months). A GH Archive / BigQuery pull is the remaining candidate.

## Separately: two things look broken on main

- `https://api.ecc.tools/badge/stars` returns **503** `{"message":"unavailable","isError":true}`, so the Stars badge already on the README is currently rendering as a grey error badge. The forks and installs badges use the same host and are worth checking.
- `README.zh-CN.md:841` still embeds a star history chart pointing at the **old repo name** `affaan-m/everything-claude-code`. Doubly dead: wrong repo, and the API is down.

Neither is touched by this PR.


haelyra:
Superseded. The badges and the generated star history chart landed directly on `main` in dcbf95bf, together with the self-hosted docs promotion. Closing this preview PR.

Note for anyone reading later: the chart is a locally generated SVG, not the `api.star-history.com` embed. That endpoint returns a "GitHub restricted access to star data" placeholder for every repo since GitHub limited the stargazers API on 2026-06-30. Details in the PR description above.

PR #2812, opened 2026-08-18
fix: accept standard tools/list request metadata

### LINKED ISSUE #2810: ecc-memory-mcp rejects Codex `_meta` parameters in tools/list and tools/call
## Summary

`ecc-memory-mcp` from `ecc-universal@2.1.0` cannot expose or execute memory tools in recent Codex clients because its JSON-RPC parameter validation rejects the MCP `_meta` field added by Codex.

The MCP configuration is recognized as enabled and `initialize` succeeds, but tool discovery fails with:

```text
MCP startup failed: Mcp error: -32602: tools/list does not accept parameters.
```

After allowing `_meta` for `tools/list`, Codex discovers `memory_search`, but invocation then fails with:

```text
Mcp error: -32602: Unknown or missing memory tool.
```

Allowing `_meta` for `tools/call` fixes the invocation as well.

## Environment

- ecc-universal: `2.1.0` (current npm latest at time of testing)
- Codex CLI bundled with ChatGPT desktop: `0.148.0-alpha.9`
- macOS
- stdio MCP transport
- MCP protocol negotiated successfully before `tools/list`

## Reproduction

Configure:

```toml
[mcp_servers.ecc-memory-vault]
command = "bash"
args = ["/path/to/ecc-memory-mcp.sh"]
startup_timeout_sec = 30
```

Wrapper:

```bash
#!/usr/bin/env bash
set -euo pipefail
export ECC_MEMORY_HARNESS="codex"
export ECC_MEMORY_ALLOW_USER_SCOPE="1"
exec /usr/local/bin/ecc-memory-mcp
```

Then start a Codex session and ask it to call `memory_search`.

Observed behavior:

1. Without a patch, `tools/list` fails because Codex includes non-empty request metadata.
2. After patching only `tools/list`, the tool becomes visible, but `tools/call` fails for the same reason.
3. After allowing `_meta` in both methods, a real Codex session successfully calls `memory_search` and receives a valid result.

## Cause

Current validation rejects every non-whitelisted top-level field:

```js
// tools/list
if (message.params && Object.keys(message.params).length > 0) {
  return jsonRpcError(message.id, -32602, 'tools/list does not accept parameters.');
}

// tools/call
Object.keys(params).some(key => !['name', 'arguments'].includes(key))
```

Codex adds the MCP-standard reserved `_meta` field.

## Suggested fix

Permit `cursor` and `_meta` for `tools/list`:

```js
const params = isRecord(message.params) ? message.params : {};
if (Object.keys(params).some(key => !['cursor', '_meta'].includes(key))) {
  return jsonRpcError(message.id, -32602, 'Invalid tools/list parameters.');
}
```

Permit `_meta` for `tools/call`:

```js
Object.keys(params).some(
  key => !['name', 'arguments', '_meta'].includes(key)
)
```

It would also be useful to add regression tests that send `_meta` with both requests.

## Verification

With those two local changes, Codex emitted an `mcp_tool_call` for `ecc-memory-vault/memory_search` and completed successfully with the expected MCP content response.

keepAwayFrom404:
## Summary

- accept MCP-reserved `_meta` on `tools/list`
- accept and validate the standard optional `cursor` field
- keep rejecting malformed metadata, malformed cursors, and unknown top-level parameters
- add regression coverage for Codex-compatible list requests

## Root cause

Recent Codex clients attach an MCP `_meta` object to `tools/list`. The memory MCP server rejected every non-empty parameter object, so initialization stopped with `-32602` before `memory_search`, `memory_save`, and the other tools could be exposed.

## Validation

```text
NODE_PATH=/usr/local/lib/node_modules/ecc-universal/node_modules node tests/scripts/memory-mcp.test.js
Results: Passed: 13, Failed: 0
```

Also verified:

- `node --check scripts/memory-mcp.mjs`
- `node --check tests/scripts/memory-mcp.test.js`
- `git diff --check`

Fixes #2810

### DIFF
--- a/scripts/memory-mcp.mjs
+++ b/scripts/memory-mcp.mjs
@@ -423,8 +423,16 @@ function createMemoryMcpService(options = {}) {
         return jsonRpcResult(message.id, {});
       }
       if (message.method === 'tools/list') {
-        if (message.params && Object.keys(message.params).length > 0) {
-          return jsonRpcError(message.id, -32602, 'tools/list does not accept parameters.');
+        const params = message.params || {};
+        if (
+          (Object.prototype.hasOwnProperty.call(params, '_meta') && !isRecord(params._meta))
+          || (
+            Object.prototype.hasOwnProperty.call(params, 'cursor')
+            && typeof params.cursor !== 'string'
+          )
+          || Object.keys(params).some(key => !['cursor', '_meta'].includes(key))
+        ) {
+          return jsonRpcError(message.id, -32602, 'Invalid tools/list parameters.');
         }
         return jsonRpcResult(message.id, {
           tools: TOOL_DEFINITIONS.map(tool => ({ ...tool })),
--- a/tests/scripts/memory-mcp.test.js
+++ b/tests/scripts/memory-mcp.test.js
@@ -132,6 +132,7 @@ async function withClient(fn, options = {}) {
 
   const client = {
     listTools: () => request('tools/list'),
+    listToolsRaw: params => request('tools/list', params),
     callTool: ({ name, arguments: toolArguments }) => request(
       'tools/call',
       { name, arguments: toolArguments }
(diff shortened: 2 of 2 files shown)


haelyra:
Thank you for catching this compatibility gap, and sorry for the delay. Support for standard tools/list request metadata, including _meta and cursor fields, has now landed through the consolidated MCP fix in #2869. The final exact-main CI matrix and CodeQL are green.

GitHub closed this PR through the consolidation. Your focused reproduction helped make the shared request path more standards-compatible, and we appreciate the contribution.

PR #2819, opened 2026-08-19
feat(skills): context-budget — account for reconnect scaffolding

ryukdev:
## Summary

`context-budget` inventories the components that load **once at startup** — agents, skills, rules, MCP schemas, CLAUDE.md. On a **reconnect**, the app re-injects much of that same inventory into the session transcript as `attachment` lines, and it lands in the same context window. The skill currently has no accounting for it.

## Measurement

On a session that had reconnected once:

| | bytes |
|---|---|
| `skill_listing` | 14,625 |
| `mcp_instructions_delta` | 8,117 |
| `deferred_tools_delta` | 7,138 |
| `agent_listing_delta` | 2,558 |
| `auto_mode` | 105 |
| **scaffolding total** | **33,884 of 49,830 — 68%** |
| actual conversation | 15,946 |

## What this PR adds

- A **measure-it-yourself** snippet (no estimation — it reads the real transcript)
- A `>40% scaffolding` flag
- The consequence for the rest of the skill: trimming a skill or an MCP server saves tokens **once per reconnect**, not once per session — which strengthens every recommendation already in there
- A caveat worth stating: transcript file size is a poor proxy for context fullness unless `attachment` lines are excluded. I hit this in my own tooling and reported a session as ~12k when the conversation in it was 0.8k

## Notes

No new skill — this extends the existing one, since `context-budget` is already the right home for it. Happy to adjust the placement or trim it down if you'd rather it sat in a different phase.

### DIFF
--- a/skills/context-budget/SKILL.md
+++ b/skills/context-budget/SKILL.md
@@ -43,6 +43,38 @@ Scan all component directories and estimate token consumption:
 - Estimate schema overhead at ~500 tokens per tool
 - Flag: servers with >20 tools, servers that wrap simple CLI commands (`gh`, `git`, `npm`, `supabase`, `vercel`)
 
+**Session scaffolding** (the transcript itself)
+
+Everything above is loaded once at startup. A *reconnect* re-injects a second copy of much of it into
+the session transcript as `attachment` lines, and that cost is charged to the same context window.
+Measure it directly rather than estimating:
+
+```sh
+# on the newest transcript for this project
+f=$(ls -t ~/.claude/projects/"${PWD//\//-}"/*.jsonl | head -1)
+python3 - "$f" <<'EOF'
+import json,sys
+tot=att=0
+for line in open(sys.argv[1]):
+    tot+=len(line)
+    try:
+        if json.loads(line).get('type')=='attachment': att+=len(line)
+    except: pass
+print(f"transcript {tot}B | scaffolding {att}B ({att*100//max(tot,1)}%) | conversation {tot-att}B")
+EOF
+```
+
+Measured on a session that had reconnected once: **33,884 of 49,830 bytes — 68% scaffolding**, broken
+down as `skill_listing` 14.6KB, `mcp_instructions_delta` 8.1KB, `deferred_tools_delta` 7.1KB,
+`agent_listing_delta` 2.6KB. The conversation itself was 15.9KB.
+
+- Flag: scaffolding >40% of the transcript — the session is paying more for its own inventory than for
+  the work
+- This is the strongest argument for the reductions elsewhere in this skill: trimming a skill or an
+  MCP server saves tokens *once per reconnect*, not once per session
+- It also means transcript file size is a poor proxy for how full a context window is. Count only
+  non-`attachment` lines when reporting "how much room is left"
+
 **CLAUDE.md** (project + user-level)
 - Count tokens per file in the CLAUDE.md chain
 - Flag: combined total >300 lines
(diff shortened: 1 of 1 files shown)


ryukdev:
Thanks both — the correctness flags were right, and I've pushed a fix.

The original text was self-contradictory: it said attachment lines are charged to the context window, then told the report to exclude them when stating remaining room. @greptile-apps' P1 is the sharper half of it — after a compaction, pre-compaction turns stay in the JSONL while the active context holds only the summary, so counting persisted bytes misreports headroom in the *other* direction as well.

Rather than pick one of the two wrong answers, the section now says a transcript is a durable log, not a view of the context window, and that remaining room must not be reported from its size at all. The measurement is reframed as a **cost signal**: scaffolding is rebuilt on every reconnect, so trimming a skill or MCP server saves tokens once per reconnect rather than once per session — which is the useful thing it tells you.

Snippet fixes from the review: the >40% threshold is now emitted by the script with one decimal instead of sitting in prose, a missing transcript is guarded, an empty one exits cleanly, bytes are counted rather than characters, and a malformed line is treated as conversation instead of crashing. Verified against all four cases.

On test coverage — happy to extract the measurement into a testable utility if you'd prefer that over an inline snippet; I left it inline to match the surrounding sections, but say the word.

ryukdev:
Good catch — that was a splice artifact from my own edit (the replacement text ended with the same anchor string it was spliced against, so the heading got emitted twice). Fixed and pushed.

Agreed on leaving the measurement inline; it isn't reused anywhere else, and matching the surrounding sections seems more valuable than extracting it for its own sake. Happy to revisit if that changes.

haelyra:
Thank you for the careful correction that distinguishes reconnect scaffolding cost from active-context headroom. That distinction is valuable for the post-2.2 `lean@1` / `full@1` profile work because it helps us measure repeated provider overhead without pretending a transcript is the context window.

We would love one more focused pass before this lands:

- **Make transcript inspection explicit and privacy-bounded.** The current default workflow opens the latest provider JSONL under `~/.claude/projects`, which may contain credential-bearing or sensitive history. Make this an optional diagnostic that requires an explicit path or confirmation, emits aggregates only, and never treats transcript access as required profile behavior. This preserves the master plan boundary that context-budget logic must not depend on provider history by default.
- **Use a tested cross-platform entry point.** The POSIX assignment, Bash path substitution, and heredoc do not run in PowerShell, and Git Bash derives the wrong Claude project slug from `/c/...`. A small checked-in script or equivalent paired commands should resolve the real Windows and POSIX locations without guessing.
- **Label the measurement as provider-specific and experimental.** `attachment` records are Claude transcript internals, not a stable cross-harness counter. Report source, provider, and degraded/unsupported states, and keep this signal separate from canonical profile inventory or token-budget authority.
- **Add fixtures for the documented edge cases.** Success is deterministic coverage for Windows and POSIX paths, missing and empty transcripts, malformed lines, reconnect attachments, and post-compaction history, with no transcript content written to output.

If you do not have bandwidth for that portability and privacy pass, that is completely okay. We will carry the bounded diagnostic into the M1 profile task queue with attribution to this contribution. We appreciate the strong correction to the original headroom claim.

chenhz01:
The reconnect-reinjection measurement is a real gap in the original skill, and the >40% flag gives it a decision rule. One correctness edge in the snippet: `ls -t ~/.claude/projects/"${PWD//\//-}"/*.jsonl | head -1` assumes the project-dir name is exactly PWD with `/` swapped for `-`. On Windows/Git Bash, `PWD` is `/c/...` while Claude Code's project directory also maps the drive colon to `-`, so the glob misses and the skill silently reports "no transcript" there. Normalizing the same way the harness does — every non-alphanumeric to `-` — fixes it:

```sh
key=$(printf '%s' "$PWD" | tr -c '[:alnum:]\n' '-')
```

Happy to open a PR with the corrected snippet if useful.

PR #2825, opened 2026-08-19
Count code lines, not total lines: 400 owes a reason, and the write hook warns instead of blocking

yeliu802:
## The problem

`rules/common/coding-style.md` says "200-400 lines typical, 800 max", and the
PreToolUse hook in `rules/web/hooks.md` enforces it with
`c.split('\n').length > 800 → process.exit(2)`. Both count **total** lines.

Two consequences, both measured on a real Swift repo rather than argued:

**1. The gate never fired on code.** The single file the 800 limit would have
blocked held **287 code lines and 485 lines of comments**. It was blocked for
being well-documented. Meanwhile three genuinely oversized test files —
638 / 538 / 418 code lines — passed, because none crossed 800 *total*.

| | old (total > 800) | new (code > 400) |
|---|---|---|
| flagged | 1 file (287 code lines) | 3 files (638 / 538 / 418 code lines) |
| missed | 3 oversized test files | — |
| wrongly flagged | the well-commented file | — |

The old metric had a 100% false-positive rate and a 100% false-negative rate
on this repo. Not a tuning problem — it measured the wrong quantity.

**2. Blocking picks the wrong fix.** The cheapest way past
`process.exit(2)` is to move code into a sibling file or delete comments.
Both make the codebase worse while the number goes green. A hard gate
selects for whichever remedy is cheapest, and that is never "split by
responsibility". Changed to `exit 0` with a stderr notice — a notice makes
someone look, a gate makes someone comply.

## What changed

- **Count code lines**: exclude comment-only and blank lines before comparing
  against any threshold. Rationale stated inline: taxing documentation at the
  same rate as code makes people delete the explanations first.
- **400 is where you owe a written reason, not where you stop.**
- **New signal that fires earlier and more accurately than any line count**:
  a type should not hold more than two independent lifecycles (tasks,
  subscriptions, timers, observers). Bugs cluster in the interleaving of
  independent lifecycles, not at line 401. In the work that prompted this,
  the only Critical-severity defect found was exactly that shape — in a type
  that no line-count rule flagged.
- **Scan the whole repo mechanically.** Added a copy-pasteable `find | grep -vcE`
  one-liner. A hand-listed set of "suspects" is how a breached limit gets
  reported as satisfied by every reviewer who checks it — which is what
  happened before this was caught.
- **`agents/code-reviewer.md`**: its HIGH-severity checklist flagged at
  >800 total lines while the rules said 400. The reviewer agent is the worst
  place to leave a stale metric, since it is the thing that does the checking.

Applied to `rules/common`, `rules/arkts`, `rules/react-native`, `rules/web`,
`agents/`, and the maintained zh-CN tree under `docs/zh-CN/rules/common`.
After this, no file under `rules/` or `docs/zh-CN/rules/` states an 800-line
limit.

## Verification

The rewritten hook was extracted from the markdown and executed, not eyeballed.
Input: 500 comment lines + 450 code lines + 30 blank lines.

```
exit code: 0
stderr: [Hook] NOTICE: 450 code lines (comments excluded). Over 400 wants a
        written reason, or a split by responsibility — not a split to buy
        line budget.
stdout: 11490 bytes (non-empty → passthrough, write proceeds)
```

## Not addressed here

`CONTRIBUTING.md` ("Under 500 lines (800 max)"), `AGENTS.md`, `.cursor/rules/`,
`.opencode/`, and `docs/ja-JP/` still carry the total-line metric. Happy to
follow up in a second PR if you want them aligned — left out to keep this
diff reviewable and because the ja-JP tree needs a translator's eye, not mine.


yeliu802:
Closing this — Greptile's findings are correct and the fix is wider than this diff.

The real defect: I changed `docs/zh-CN/rules/common/*` to the code-line metric but left `docs/zh-CN/agents/code-reviewer.md`, `docs/zh-CN/commands/code-review.md`, and `docs/zh-CN/hooks/README.md` on 800 total lines with the blocking `process.exit(2)` hook. That contradiction did not exist before this PR — I introduced it by converting half a tree. Chasing it also turned up `commands/code-review.md:50` in the English tree, which I missed entirely.

Reopening once every tree that states the rule states the same rule. Sorry for the noise.

PR #2827, opened 2026-08-20
fix(install): surface actionable errors when runtime deps are missing

### LINKED ISSUE #2822: Plugin install ships without node_modules: install-plan.js and control-pane.js crash with Cannot find module 'ajv' / 'sql.js'
## Summary

When ECC is installed through the Claude Code plugin marketplace, the marketplace directory is a **git clone** of this repository. `npm install` never runs, so no `node_modules` exists. Four runtime dependencies declared in `package.json` are therefore missing, and every script that needs them fails.

This breaks the documented install-planning path. `/project-init` and any workflow that calls `scripts/install-plan.js` cannot run at all.

## Environment

| | |
|---|---|
| ECC | `ecc-universal` 2.2.0 |
| Commit | `06c5e11` |
| Install path | Claude Code plugin marketplace (`~/.claude/plugins/marketplaces/ecc`) |
| Node | v24.16.0 |
| OS | Windows 11, Git Bash |

## Reproduction

```bash
cd ~/.claude/plugins/marketplaces/ecc
ls node_modules          # does not exist
node scripts/install-plan.js --list-profiles
```

## Actual

```
Error: Cannot find module 'ajv'
Require stack:
- .../scripts/lib/install/config.js
- .../scripts/install-plan.js
```

`scripts/install-apply.js` fails the same way, but the crash is then followed by the usage banner:

```bash
node scripts/install-apply.js --target claude-project --dry-run --skills kotlin-patterns
```

```
Error: Cannot find module 'ajv'
Require stack:
- .../scripts/lib/install/config.js
- .../scripts/install-apply.js
Usage: install.sh [--target <claude|claude-project|cursor|antigravity>] ...
```

That output reads like an argument error, which sends people looking in the wrong place. `--help` alone exits before the require and appears to work, which reinforces the wrong diagnosis.

`scripts/control-pane.js` fails equivalently on `sql.js`.

## Expected

Either the dependencies are present in a plugin install, or the failure names the actual problem and the fix.

## Scope

`package.json` declares four runtime dependencies and has **no `postinstall` script**:

```json
"dependencies": {
  "@iarna/toml": "2.2.5",
  "ajv": "8.20.0",
  "js-yaml": "4.3.1",
  "sql.js": "1.14.1"
}
```

11 files require them:

```
scripts/ci/validate-hooks.js
scripts/ci/validate-install-manifests.js
scripts/codex/merge-codex-config.js
scripts/codex/merge-mcp-config.js
scripts/lib/control-pane/state.js
scripts/lib/install/antigravity-agent.js
scripts/lib/install/config.js
scripts/lib/mcp-inventory/readers/codex.js
scripts/lib/state-store/index.js
scripts/lib/state-store/schema.js
scripts/memory-mcp.mjs
```

User-facing entry points affected: `install-plan.js`, `install-apply.js` (with real arguments), `control-pane.js`, `memory-mcp.mjs`.

The `npm i -g ecc-universal` path is unaffected — npm installs the dependencies there. Only the plugin/marketplace path is broken, and for Claude Code that is the primary documented install.

`scripts/doctor.js` runs fine and does **not** report the missing dependencies, so the built-in diagnostic does not surface this.

## Workaround

Installing the four packages elsewhere and pointing `NODE_PATH` at them works, and avoids writing into the self-updating marketplace clone:

```bash
mkdir -p /tmp/ecc-deps && cd /tmp/ecc-deps
npm install --no-save @iarna/toml@2.2.5 ajv@8.20.0 js-yaml@4.3.1 sql.js@1.14.1
NODE_PATH=/tmp/ecc-deps/node_modules node ~/.claude/plugins/marketplaces/ecc/scripts/install-plan.js --list-profiles
```

## Suggested fixes, roughly in order of value

1. **Drop the runtime dependency on `ajv`.** It is used to validate ECC's own manifests, which ship with the repo and are already validated in CI (`scripts/ci/validate-install-manifests.js`). Re-validating known-good bundled files on every user invocation buys little and costs the entire install path when the dependency is absent. A small hand-rolled shape check, or trusting CI, would remove the most damaging instance.
2. **Make the failure self-explanatory.** Catch `MODULE_NOT_FOUND` at the require boundary and print the missing package plus the install command, instead of a stack trace followed by a usage banner that implies bad arguments.
3. **Teach `doctor.js` about it** and let `repair.js` fix it, so the existing diagnostics actually catch the condition.
4. **Bootstrap on first run** — have the plugin entry point install the production dependencies once into the marketplace directory if they are absent. Weakest option: it writes into a directory that auto-updates, and on Windows that clone has had write-protected git pack files.

Happy to open a PR for (1) and (2) if that direction is welcome.

santhiprakash:
## What Changed

- Added `scripts/lib/require-runtime.js` to turn `MODULE_NOT_FOUND` into an actionable message that tells plugin/marketplace users to run `npm install` from the ECC root.
- Lazy-loaded `ajv`/`sql.js`/`@iarna/toml` in install and control-pane code paths so listing commands do not require `node_modules`.
- Deferred `install-plan.js` config loading until a config-backed code path runs.
- Stopped `install-apply.js` from appending the usage banner when the failure is a missing runtime dependency.

## Why This Change

Fixes #2822. Claude Code plugin/marketplace installs are git clones without `node_modules`, so `install-plan.js`, `install-apply.js`, and `control-pane.js` crashed with raw `Cannot find module 'ajv'/'sql.js'` stack traces. `install-apply.js` also printed its usage banner after the stack trace, which looked like an argument error.

## Testing Done
- [x] Automated tests pass locally (`node tests/lib/require-runtime.test.js`, `node tests/scripts/install-plan.test.js`, `node tests/lib/install-config.test.js`)

## Type of Change
- [x] `fix:` Bug fix

- [x] Shell scripts pass shellcheck (if applicable)

- [ ] Ran `yarn install --mode=update-lockfile` and committed the `yarn.lock` change. CI runs Yarn in hardened mode on public PRs and fails if the lockfile would be modified, so an out of date `yarn.lock` breaks the build even when nothing else is wrong.

- [ ] Not applicable

## Documentation
- [ ] Updated relevant documentation
- [ ] README updated (if needed)

### DIFF
--- a/scripts/install-apply.js
+++ b/scripts/install-apply.js
@@ -195,7 +195,11 @@ async function main() {
       printHumanPlan(result, false);
     }
   } catch (error) {
-    process.stderr.write(`Error: ${error.message}${getHelpText()}`);
+    const { isMissing } = require('./lib/require-runtime');
+    const suffix = error.code === 'ECC_RUNTIME_DEPENDENCY_MISSING' || isMissing(error)
+      ? '\n'
+      : getHelpText();
+    process.stderr.write(`Error: ${error.message}${suffix}`);
     process.exit(1);
   }
 }
--- a/scripts/install-apply.js
+++ b/scripts/install-apply.js
@@ -195,8 +195,7 @@ async function main() {
       printHumanPlan(result, false);
     }
   } catch (error) {
-    const { isMissing } = require('./lib/require-runtime');
-    const suffix = error.code === 'ECC_RUNTIME_DEPENDENCY_MISSING' || isMissing(error)
+    const suffix = error.code === 'ECC_RUNTIME_DEPENDENCY_MISSING'
       ? '\n'
       : getHelpText();
     process.stderr.write(`Error: ${error.message}${suffix}`);
--- a/scripts/install-plan.js
+++ b/scripts/install-plan.js
@@ -9,10 +9,6 @@ const {
   listInstallProfiles,
   resolveInstallPlan,
 } = require('./lib/install-manifests');
-const {
-  findDefaultInstallConfigPath,
-  loadInstallConfig,
-} = require('./lib/install/config');
 const { normalizeInstallRequest } = require('./lib/install/request');
 
 function showHelp() {
@@ -237,6 +233,10 @@ function main() {
       return;
     }
 
+    const {
+      findDefaultInstallConfigPath,
+      loadInstallConfig,
+    } = require('./lib/install/config');
     const defaultConfigPath = options.configPath
       ? null
       : findDefaultInstallConfigPath({ cwd: process.cwd() });
--- a/scripts/lib/control-pane/state.js
+++ b/scripts/lib/control-pane/state.js
@@ -4,8 +4,7 @@ const fs = require('fs');
 const os = require('os');
 const path = require('path');
 
-const initSqlJs = require('sql.js');
-const toml = require('@iarna/toml');
+const { requireRuntime } = require('../require-runtime');
 
 const { buildControlPaneActions } = require('./actions');
 
@@ -86,6 +85,7 @@ function normalizeConfig(rawConfig = {}, options = {}) {
 
 function readTomlConfig(configPath) {
   const raw = fs.readFileSync(configPath, 'utf8');
+  const toml = requireRuntime('@iarna/toml');
   return toml.parse(raw);
 }
 
@@ -113,6 +113,7 @@ function resolveControlPaneConfig(options = {}) {
 
 async function openSqlDatabase(dbPath) {
   if (!dbPath || !fs.existsSync(dbPath)) return null;
+  const initSqlJs = requireRuntime('sql.js');
   const SQL = await initSqlJs();
   const buffer = fs.readFileSync(dbPath);
   return new SQL.Database(buffer);
--- a/scripts/lib/install/antigravity-agent.js
+++ b/scripts/lib/install/antigravity-agent.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { requireRuntime } = require('../require-runtime');
+
 const TOOL_NAMES = Object.freeze({
   Read: 'view_file',
   Write: 'write_to_file',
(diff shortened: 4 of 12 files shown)


haelyra:
Thank you for turning raw module-resolution crashes into clear diagnostics and for isolating the eager imports. The bug report is valid, and I am sorry this sat without a maintainer response.

This needs one distribution-contract revision before merge. A marketplace or plugin install that exposes an ECC command but then tells the user to run `npm install` inside a provider-managed cache is still a broken install: it mutates a cache outside the receipt-backed installer, can resolve dependencies different from the shipped lockfile, and fails offline. Lazy loading usefully keeps unrelated commands alive, but it cannot make manual dependency repair the supported runtime path.

A mergeable version should:

- make commands whose runtime dependencies are absent report that the capability was not installed, with the exact supported ECC install or repair command, not a raw `npm install`;
- ensure the packed or installer-managed surface carries the pinned runtime closure for every advertised capability, with an install receipt and uninstall/repair coverage;
- distinguish source checkout, npm package, and provider marketplace layouts without asking users to mutate marketplace caches;
- add packed-artifact and plugin-cache fixtures proving list/help remain available, unavailable capabilities fail closed, and the supported repair restores them offline from the qualified artifact.

This belongs in the ECC 2 distribution foundation and advances the truthful capability contract, but the installer must remain the authority over dependency state. If you do not have bandwidth for that packaging pass, please tell us and we will carry the actionable-error work into the maintainer queue with attribution. We appreciate the focused diagnosis and tests.

santhiprakash:
Thanks for the thorough review and the clear distribution-contract framing.

We verified your premise against the current installer surface before answering: `install.sh` / `install.ps1` run `npm install` in `$SCRIPT_DIR` only when `node_modules` is absent — a step that never runs for plugin/marketplace loads — and `scripts/repair.js` rebuilds ECC-managed files from install-state but records no runtime-dependency closure. So there is today no supported command that can restore missing runtime deps inside a provider-managed cache: the PR's `npm install` pointer was indeed the wrong answer, and the "supported install/repair command" in your first bullet only becomes real once the packed closure from your second bullet exists.

Given that dependency between the bullets, we'll take your offered path: please carry the packaging pass (pinned runtime closure + install receipt + uninstall/repair coverage + packed-artifact/plugin-cache fixtures) into the ECC 2 distribution foundation queue, with attribution. The actionable-error diagnosis in this PR — lazy-load isolation, the `ECC_RUNTIME_DEPENDENCY_MISSING` classification, and the affected call sites — is ours to contribute and yours to fold in wherever it fits. We'd rather not freelance a receipt/uninstall contract change on the installer core from the outside.

Leave this PR open for your consolidation — fold, supersede, or close it as the foundation work dictates. Thanks again for the precise review.

santhiprakash:
Closing to keep the review queue clean — this has been open a while without traction. Happy to reopen or resubmit a fresh version if it would still be useful.

PR #2828, opened 2026-08-21
fix(hooks): deny dd if= arguments whose value starts with a non-word char

### LINKED ISSUE #2642: H-03: destructive commands ending in / . or a quote bypass DESTRUCTIVE_BASH (installed release; fixed on main)
# H-03: destructive commands ending in `/`, `.`, or a quote are allowed

**Summary:** In the installed ECC release, the trailing `\b` on two `DESTRUCTIVE_BASH` alternation arms prevents matching commands whose next character is `/`, `.`, or a quote.

**Affected version:** `2.0.0-rc.1`, commit `841beea45cb2`.

## Environment

The observation is against the **installed copy** of `gateguard-fact-force.js`, which is 487 lines. The plugin version is `2.0.0-rc.1` at commit `841beea45cb2`.

## Reproduction

Save as `repro.js` and run with `node repro.js`. Replace `<plugin>` with the installed
plugin directory.

```js
const os = require('os'), fs = require('fs'), path = require('path');

// Isolate hook state so the repro cannot touch a real ~/.gateguard directory.
process.env.HOME = fs.mkdtempSync(path.join(os.tmpdir(), 'ggrepro-'));

// The module exports exactly one function: run(rawInput).
const { run } = require('<plugin>/scripts/hooks/gateguard-fact-force.js');

// A deny is signalled by run() returning something OTHER than the object passed in.
function decide(command) {
  const input = { tool_name: 'Bash', tool_input: { command }, session_id: 'repro' };
  return run(input) === input ? 'ALLOWED' : 'DENIED';
}

decide('printf ready');   // prime — see note below

for (const command of [
  'dd if=/dev/zero of=/dev/sda',
  'dd if=x',
  'git checkout -- .',
]) console.log(decide(command).padEnd(8), command);
```

Two notes that otherwise make this confusing to reproduce:

- **`tool_name` is required.** Without it, `run()` falls through every branch and returns
  the input unchanged, so every command reads as ALLOWED and the defect looks absent.
- **Prime the session.** The first Bash command of a session trips a separate routine
  gate, which masks the destructive arm on whichever command happens to be first.

## Observed results

| Command | Installed decision |
| --- | --- |
| `dd if=/dev/zero of=/dev/sda` | ALLOWED |
| `dd if=x` | DENIED |
| `git checkout -- .` | ALLOWED |

## Impact

The classic disk-wipe spelling can pass the gate, allowing a command to overwrite a disk. `git checkout -- .` can discard all uncommitted work.

## Suggested fix

Drop the trailing `\b` from the affected alternation arms. This strictly widens coverage and cannot narrow it. Upstream `main` already contains a structural `findGitSubcommand()` parser that addresses this failure mode. Please cut a release containing that parser and add regression tests for these cases.

santhiprakash:
## Summary

Closes #2642. The destructive-command gate allowed every realistic disk-wipe spelling of `dd if=` because the regex used a trailing `\b` anchor on the whole alternation, and `\b` cannot match after a non-word character (the `=` in `dd if=`). The companion `git checkout -- .` arm mentioned in the same report is already covered by the structural `findGitSubcommand()` parser on `main`.

## Problem

```js
const DESTRUCTIVE_SQL_DD = /\b(drop\s+table|delete\s+from|truncate|dd\s+if=)\b/i;
```

For `dd if=/dev/zero of=/dev/sda`, the engine matches `dd if=` but then needs `\b` at the position where `/` is the next character. Since `=` and `/` are both non-word characters, there is no word boundary there and the match fails. The hook then falls through to the other gates (rm, git, find -exec, ...), all of which miss the bare `dd` invocation.

## Root cause

The trailing `\b` is needed for the SQL arms to reject benign substrings (`truncatefile`, `delete fromtable`). For the `dd if=` arm it is both wrong (the `=` is non-word) and unnecessary (the arm already requires a literal `if=` token, which has no shorter superstring to confuse with).

## Fix

Split the alternation so the SQL arms keep their trailing `\b` while the `dd if=` arm drops it:

```js
const DESTRUCTIVE_SQL_DD = /\b(?:drop\s+table|delete\s+from|truncate)\b|\bdd\s+if=/i;
```

The regex still requires a literal whitespace token `if=` after `dd`, so false positives (`ddiff`, `dd ifbackup`) remain impossible.

## Verification

Reproduction against the published v2.1.0 hook vs. the fix:

| Command | Pre-fix | Post-fix |
| --- | --- | --- |
| `dd if=/dev/zero of=/dev/sda` | ALLOWED | DENIED |
| `dd if=/dev/urandom of=/dev/sdb` | ALLOWED | DENIED |
| `dd if=/dev/zero` | ALLOWED | DENIED |
| `dd if=backup.img of=/dev/sdc` | ALLOWED | DENIED |
| `dd if=x` (control: was already denied) | DENIED | DENIED |
| `drop table users;` | DENIED | DENIED |
| `delete from t` | DENIED | DENIED |
| `truncate -s 0 f` | DENIED | DENIED |
| `delete fromtable` (anti-regression) | ALLOWED | ALLOWED |
| `truncatefile` (anti-regression) | ALLOWED | ALLOWED |

`node tests/hooks/gateguard-fact-force.test.js` now runs 158 tests (up from 144), all green, with 14 new regression cases covering every command spelling called out in the issue, the command-substitution and compound-command variants, and the two anti-regression checks that the SQL arms still keep their `\b`.

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -48,7 +48,16 @@ const ECC_ENABLE_VALUES = new Set(['1', 'true', 'on', 'enabled', 'enable', 'yes'
 // phrases without shell-flag ordering concerns. Quoted strings are
 // stripped before this regex runs so a commit message mentioning
 // "drop table" no longer triggers a false positive.
-const DESTRUCTIVE_SQL_DD = /\b(drop\s+table|delete\s+from|truncate|dd\s+if=)\b/i;
+//
+// The `dd\s+if=` arm deliberately omits a trailing word-boundary anchor:
+// `dd if=` ends with `=`, a non-word character, so `\b` would refuse to
+// match before the next non-word character (most commonly `/` for
+// `/dev/*` targets, but also `.`, quotes, `;`, `|`, `&`, end-of-string,
+// etc.). Anchoring that arm with `\b` was the cause of issue #2642:
+// `dd if=/dev/zero of=/dev/sda` slipped past the gate. The SQL arms keep
+// their trailing `\b` so words like `delete fromtable` or `truncatefile`
+// do not produce false positives.
+const DESTRUCTIVE_SQL_DD = /\b(?:drop\s+table|delete\s+from|truncate)\b|\bdd\s+if=/i;
 
 // Operator-supplied additional destructive patterns. Lazily compiled from
 // `GATEGUARD_BASH_EXTRA_DESTRUCTIVE` (regex source) on first use, then
--- a/tests/hooks/gateguard-fact-force.test.js
+++ b/tests/hooks/gateguard-fact-force.test.js
@@ -2376,6 +2376,165 @@ function runTests() {
     passed++;
   else failed++;
 
+  // --- Issue #2642: trailing \b on the `dd\s+if=` arm fails open when the
+  // argument begins with a non-word character (`/`, `.`, quote, ...), which
+  // covers every realistic disk-wipe spelling. The fix splits the alternation
+  // so the SQL arms keep their trailing word-boundary anchor (no
+  // `delete fromtable` / `truncatefile` false positives) while the `dd if=`
+  // arm matches whatever follows the `=`.
+
+  clearState();
+  if (
+    test('denies `dd if=/dev/zero of=/dev/sda` (issue #2642 disk-wipe spelling)', () => {
+      expectDestructiveDeny('dd if=/dev/zero of=/dev/sda', 'dd if=/dev/zero of=/dev/sda');
+    })
+  )
+    passed++;
+  else failed++;
+
+  clearState();
+  if (
+    test('denies `dd if=/dev/urandom of=/dev/sdb`', () => {
+      expectDestructiveDeny('dd if=/dev/urandom of=/dev/sdb', 'dd if=/dev/urandom of=/dev/sdb');
+    })
+  )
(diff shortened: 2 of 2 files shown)


santhiprakash:
Closing in favor of #2829, which provides a more thorough structural fix to the same issue.

#2829 supersedes this PR by replacing the regex-based `DESTRUCTIVE_SQL_DD` arm with a token-based `isDestructiveDd(tokens)` check that:

1. Fixes the same `dd if=/dev/zero` slash-prefixed bypass this PR was written for (issue #2642).
2. Extends coverage to additional spellings the regex approach cannot catch:
   - reversed operands: `dd of=/dev/sda if=/dev/zero`
   - intervening options: `dd bs=1M if=/dev/zero of=/dev/sda`
   - wrapper commands: `sudo dd if=/dev/zero of=/dev/sda`
3. Avoids the `echo dd if=x` and `grep dd if=...` text-match false positives my regex change introduced.
4. Carries an updated regression suite that covers every form above plus the existing SQL deny cases and the `delete fromtable` / `truncatefile` anti-regression cases.

Both CodeRabbit and Greptile approved #2829 after the rework (last reviewed commit `4ef558f`). Keeping both PRs open would create a duplicate edit on `scripts/hooks/gateguard-fact-force.js` line 51 and force a maintainer to choose between the two approaches.

Thanks to @ntdat812 for picking up the structural rework.

PR #2831, opened 2026-08-21
fix(rules): stop prescribing JS casing for every language in common/

### LINKED ISSUE #2830: common/coding-style.md prescribes camelCase for all languages, contradicting python/ and rust/
## Problem

`rules/common/coding-style.md` is the language-agnostic style file and has no `paths:` frontmatter, so it loads for every source file regardless of language. Its Naming Conventions section, however, prescribes JavaScript/TypeScript casing:

```markdown
## Naming Conventions

- Variables and functions: `camelCase` with descriptive names
- Booleans: prefer `is`, `has`, `should`, or `can` prefixes
- Interfaces, types, and components: `PascalCase`
- Constants: `UPPER_SNAKE_CASE`
- Custom hooks: `camelCase` with a `use` prefix
```

That contradicts two language files in the same package:

- `rules/python/coding-style.md:12` — "Follow **PEP 8** conventions" (i.e. `snake_case`)
- `rules/rust/coding-style.md:45` — "`snake_case` for functions, methods, variables, modules, crates"

Both language files carry `paths:` frontmatter, so for a `.py` or `.rs` file the agent gets `common/coding-style.md` **and** the language file in the same context — two opposite naming rules.

**Correction to my original wording:** `rules/README.md` does state that language-specific rules take precedence, so the precedence is defined — I was wrong to write that nothing says which wins. The narrower point stands: that statement lives in the README, not in the rule files delivered to the agent at inference time, so the two rules arrive side by side with nothing marking one as the override. `rules/README.md` also documents a `> **Language note**:` marker for exactly this case, which the Naming Conventions section does not use.

The last line is also misplaced regardless of casing: "Custom hooks" is a React concept in a file that is supposed to be language-agnostic, and it is already covered in `rules/react/coding-style.md` (line 24 for the `useCamelCase` symbol rule, line 84 for the `eslint-plugin-react-hooks` enforcement).

## Impact

Anyone using this rule package on a Python or Rust project gets a hard contradiction injected on every file touch. In practice the model picks one, and which one it picks is not predictable from the rules.

## Suggested fix

Keep in `common/` only what is genuinely language-independent (descriptive names, boolean prefixes, constants and types being visually distinct from values), and defer casing to the language-specific file using the `> **Language note**:` marker the README already prescribes.

I have this running locally as a patch and am happy to open a PR if the direction looks right.

clydeRbuegel:
Closes #2830.

## Problem

`rules/common/coding-style.md` has no `paths:` frontmatter, so it is loaded for every source file regardless of language. Its Naming Conventions section nevertheless prescribed JavaScript/TypeScript casing:

```markdown
- Variables and functions: `camelCase` with descriptive names
...
- Custom hooks: `camelCase` with a `use` prefix
```

That is not idiomatic for several languages this package supports, and the language files say the opposite:

- `rules/python/coding-style.md` — "Follow **PEP 8** conventions" (`snake_case`)
- `rules/rust/coding-style.md` — "`snake_case` for functions, methods, variables, modules, crates"

Both carry `paths:` frontmatter, so for a `.py` or `.rs` file the agent is handed both files and two opposite naming rules in the same context.

`rules/README.md` does state that language-specific rules take precedence, so the precedence *is* defined — but it is defined in the README, not in the rule files the agent actually receives at inference time. From the model's point of view the two rules arrive side by side with nothing marking one as the override.

The "Custom hooks" line is also misplaced regardless of casing: it is a React concept in the language-agnostic file.

## Change

- The casing list is replaced by the canonical `> **Language note**:` marker documented in `rules/README.md`, which is the mechanism this package already prescribes for exactly this situation.
- `common/` keeps only what is genuinely language-independent: descriptive names, boolean prefixes, and constants and types being visually distinct from ordinary values — the last one qualified, since not every language draws that distinction.
- The per-language examples name only languages whose own `coding-style.md` actually states a casing standard: PEP 8 for Python, `snake_case` for Rust, `camelCase` for Java and Kotlin. (Python states the standard rather than the casing itself, hence the wording.)
- The "Custom hooks" line is dropped and linked to `react/coding-style.md`, which documents it as the `useCamelCase` symbol rule and again as the `eslint-plugin-react-hooks` enforcement note.

One file changed, +13 / −5.

## Two things worth your judgement

**1. React hook rule loses some path coverage.** `rules/react/coding-style.md` is scoped to `**/*.tsx`, `**/*.jsx`, `**/components/**/*.{ts,js}` and `**/hooks/**/*.{ts,js}`. A hook colocated outside those paths — say `src/features/foo/useFoo.ts` — previously picked up the rule from the always-loaded `common/` file and now matches neither React path. Removing a React rule from the language-agnostic file seems right to me either way, but the clean fix may be to widen the React `paths:` rather than to keep the line in `common/`. Happy to follow whichever you prefer.

**2. JavaScript/TypeScript has no naming section at all.** `rules/typescript/coding-style.md` covers types, immutability, error handling, input validation and console.log, but never states a casing rule — so after this change, generic JS/TS variable and function casing is not written down anywhere in the package. This PR deliberately does not add one, since inventing a rule for a language the maintainers have not documented felt out of scope. Say the word and I will add a `## Naming` section there in this same PR.

### DIFF
--- a/rules/common/coding-style.md
+++ b/rules/common/coding-style.md
@@ -59,18 +59,15 @@ ALWAYS validate at system boundaries:
 ## Naming Conventions
 
 > **Language note**: This rule may be overridden by language-specific rules for
-> languages where this pattern is not idiomatic. Casing in particular belongs to
-> the language file — e.g. PEP 8 for Python, `snake_case` for Rust, `camelCase`
-> for Java and Kotlin. React hook naming lives in
-> [react/coding-style.md](../react/coding-style.md).
+> languages where a pattern is not idiomatic. Casing and framework-specific
+> prefixes belong to the applicable language or package rule.
 
 Language-independent:
 
 - Descriptive names: the name says what the thing holds or does, without a comment.
 - Booleans read as a claim: prefix with `is`, `has`, `should` or `can`.
 - Where the language draws the distinction, constants and types are visually
-  distinct from ordinary values (`UPPER_SNAKE_CASE` and `PascalCase` in many
-  languages) — whether it draws it at all is for the language file to say.
+  distinct from ordinary values in the form its language or package rule defines.
 
 ## Code Smells to Avoid
 
--- a/rules/common/coding-style.md
+++ b/rules/common/coding-style.md
@@ -65,7 +65,8 @@ ALWAYS validate at system boundaries:
 Language-independent:
 
 - Descriptive names: the name says what the thing holds or does, without a comment.
-- Booleans read as a claim: prefix with `is`, `has`, `should` or `can`.
+- Boolean names read clearly as claims under the applicable language or package
+  convention.
 - Where the language draws the distinction, constants and types are visually
   distinct from ordinary values in the form its language or package rule defines.
 
--- a/rules/common/coding-style.md
+++ b/rules/common/coding-style.md
@@ -58,11 +58,19 @@ ALWAYS validate at system boundaries:
 
 ## Naming Conventions
 
-- Variables and functions: `camelCase` with descriptive names
-- Booleans: prefer `is`, `has`, `should`, or `can` prefixes
-- Interfaces, types, and components: `PascalCase`
-- Constants: `UPPER_SNAKE_CASE`
-- Custom hooks: `camelCase` with a `use` prefix
+> **Language note**: This rule may be overridden by language-specific rules for
+> languages where this pattern is not idiomatic. Casing in particular belongs to
+> the language file — e.g. PEP 8 for Python, `snake_case` for Rust, `camelCase`
+> for Java and Kotlin. React hook naming lives in
+> [react/coding-style.md](../react/coding-style.md).
+
+Language-independent:
+
+- Descriptive names: the name says what the thing holds or does, without a comment.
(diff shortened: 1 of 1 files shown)


haelyra:
Thank you for identifying the cross-language conflict and for keeping the original patch tightly scoped. The core change is right. I applied the remaining review adjustment directly to this branch at `f2a7aeb3`: the common rule now contains no language-specific casing examples or framework prefix rule.

No further work is needed from you. I preserved your original commit and authorship in same-repository integration PR #2937 so the full protected CI matrix can run against exact current `main`. Once that gate is green, I will merge the integration and close this PR as landed through the maintained patch. Thank you again for the precise report and for calling out the two scope questions explicitly.

haelyra:
Thank you for the original shared-linter cleanup and the patience while maintainers consolidated it. Your original commit and authorship are now on verified green main through #3071, with a bounded follow-up for language-neutral terminology and regression coverage. Main commit c9148d0bb239ed01a95724a5928b98cdf9c30658 passes the full 42-job CI matrix, CodeQL, and Supply-Chain Watch. I am closing this source PR as integrated.

PR #2837, opened 2026-08-22
fix(block-no-verify): match abbreviated --no-verify long flags

ntdat812:
## What Changed

`hasNoVerifyFlag` compared the flag with `===` against the full spelling `--no-verify`. Git resolves any **unambiguous prefix** of a long option, so the abbreviated forms skip the hooks just the same and walked past the gate.

## Why This Change

Measured against real git, in a throwaway repo with a `pre-commit` hook that prints `HOOK RAN` and exits 1:

```
git commit                -> HOOK RAN, commit refused
git commit --no-verify    -> committed, hook skipped
git commit --no-verif     -> committed, hook skipped
git commit --no-veri      -> committed, hook skipped
```

And through the published hook runner (`run-with-flags.js`, `pre:bash:block-no-verify`, profile `standard`):

| command | before | after |
| --- | --- | --- |
| `git commit -m "x"` | allow | allow |
| `git commit --no-verify -m "x"` | BLOCK | BLOCK |
| `git commit --no-verif -m "x"` | **allow** | BLOCK |
| `git commit --no-veri -m "x"` | **allow** | BLOCK |
| `git push --no-verif` | **allow** | BLOCK |

Two characters off the end was the whole bypass. The rest of this hook is careful — it already handles the `--` terminator, options that consume a value, inline `=value` forms, short-option clusters like `-an`, and case-variant `core.HOOKSPATH` — so this reads as a gap rather than an oversight in approach.

## How it is bounded

The check now matches any prefix of `--no-verify` longer than `--no-`.

- **`--no-verbose` is not affected.** It is not a prefix of `--no-verify`, so `'--no-verify'.startsWith('--no-verbose')` is false. Covered by a test.
- **The shortest forms (`--no-v`, `--no-ve`) are deliberately blocked.** They are ambiguous — git rejects them itself with a usage error — so no working command uses them, and treating them as a bypass attempt costs nothing. I would rather over-match a form git already refuses than leave a gap.
- `core.hooksPath` is untouched: git config keys are not abbreviated, so the same class of bypass does not apply there.

Verified still allowed: `git commit --no-verbose`, `git commit --no-edit --amend`, `git commit -m "no-verif was the bug"`.

## Testing Done

9 new cases in `tests/hooks/block-no-verify.test.js`, driven end-to-end through `run-with-flags.js`.

Reverting only `scripts/hooks/block-no-verify.js`:

```
✗ blocks abbreviated --no-verif on git commit
✗ blocks abbreviated --no-veri on git commit
✗ blocks abbreviated --no-ver on git commit
✗ blocks abbreviated --no-ve on git commit
✗ blocks abbreviated --no-v on git commit
✗ blocks abbreviated --no-verif on git push
Passed: 32  Failed: 6
```

With the change: **38 passed, 0 failed**. The 3 must-allow cases pass either way by design — they are the over-blocking guard, not evidence of the bug.

Sibling suite `tests/hooks/cursor-block-no-verify.test.js`: **14 passed, 0 failed**.

**Why `run-all.js` is unchecked:** it stops in my environment at `tests/lib/claude-plugin-setup.test.js`, which performs real marketplace `git clone`s (`fatal: fetch-pack: invalid index-pack output`). That is environmental and unrelated. The two files above are the only ones that reference this hook (`grep -rl block-no-verify tests/`), so the suites I did run cover this change's blast radius.

## Type of Change

- [x] `fix:` Bug fix

- [x] JSON files validate cleanly (none touched)
- [x] Shell scripts pass shellcheck (none touched)

## Relationship to #2829 / #2832

Independent — different hook, different function, branches from `main`. All three can merge in any order.

### DIFF
--- a/scripts/hooks/block-no-verify.js
+++ b/scripts/hooks/block-no-verify.js
@@ -206,12 +206,52 @@ function findCommandSegmentEnd(input, start) {
   return input.length;
 }
 
+/**
+ * True when `value` is the option itself or an abbreviation git resolves to it.
+ *
+ * Git accepts any unambiguous prefix of a long option, so `--push-opti` is
+ * `--push-option` and consumes the following token exactly like the full
+ * spelling. An exact-name lookup missed that and refused
+ * `git push --push-opti --no-verify`, a command git runs with its hooks intact
+ * (the flag becomes the option's value).
+ *
+ * Only value-taking options are prefix-matched, and only when the prefix is
+ * unique among them. A prefix git itself rejects as ambiguous may still resolve
+ * here (`--rec` is also `--recurse-submodules`), but git refuses to run such a
+ * command at all, so it cannot skip a hook either way. The inline `=value` form
+ * is one token and consumes nothing, so it is never a match.
+ */
+function optionConsumesNextValue(value, optionsWithValue) {
+  if (optionsWithValue.has(value)) {
+    return true;
+  }
+
+  if (!value.startsWith('--') || value.length <= 2 || value.includes('=')) {
+    return false;
+  }
+
+  let resolved = null;
+  for (const option of optionsWithValue) {
+    if (!option.startsWith('--') || !option.startsWith(value)) {
+      continue;
+    }
+
+    if (resolved !== null) {
+      return false;
+    }
+
+    resolved = option;
+  }
+
+  return resolved !== null;
+}
+
 function commitOptionConsumesNextValue(value) {
   if (isCommitNoVerifyShortFlag(value)) {
     return false;
   }
 
-  if (COMMIT_OPTIONS_WITH_VALUE.has(value)) {
+  if (optionConsumesNextValue(value, COMMIT_OPTIONS_WITH_VALUE)) {
     return true;
   }
 
@@ -469,7 +509,7 @@ function hasNoVerifyFlag(input, command, offset) {
       }
     }
 
-    if (command === 'push' && PUSH_OPTIONS_WITH_VALUE.has(value)) {
+    if (command === 'push' && optionConsumesNextValue(value, PUSH_OPTIONS_WITH_VALUE)) {
       skipNext = true;
       continue;
     }
--- a/scripts/hooks/block-no-verify.js
+++ b/scripts/hooks/block-no-verify.js
@@ -262,8 +262,13 @@ const PUSH_OPTIONS_WITH_VALUE = new Set([
   '--receive-pack',
   '--exec',
   '--repo',
-  '--force-with-lease',
 ]);
+// NOT in the set: `--force-with-lease`. Its value is OPTIONAL and inline-only,
+// so the bare form consumes nothing and the next token is a real flag. Listing
+// it would make `git push --force-with-lease --no-verify` skip the very flag
+// this hook exists to catch. Confirmed against git: every option above answers
+// "requires a value" when given none, while a bare `git push
+// --force-with-lease` parses and fails later on the push destination.
 
 /**
  * Git resolves any UNAMBIGUOUS prefix of a long option, so `--no-verif` and
--- a/scripts/hooks/block-no-verify.js
+++ b/scripts/hooks/block-no-verify.js
@@ -247,6 +247,30 @@ function getCommitShortValueOption(value) {
   return null;
 }
 
+/**
+ * Git resolves any UNAMBIGUOUS prefix of a long option, so `--no-verif` and
+ * `--no-veri` bypass the hooks exactly as `--no-verify` does. Matching the full
+ * spelling alone let two characters off the end walk straight past this gate.
+ *
+ * Verified against real git with a failing pre-commit hook:
+ *   git commit                -> hook ran, commit refused
+ *   git commit --no-verif     -> committed, hook skipped
+ *   git commit --no-veri      -> committed, hook skipped
+ *
+ * Matching every prefix cannot catch `--no-verbose`, because that is not a
+ * prefix of `--no-verify`. The shortest forms (`--no-v`, `--no-ve`) are
+ * ambiguous and git rejects them itself, so treating them as a bypass attempt
+ * costs nothing — no working command uses them.
+ *
+ * @param {string} value
+ * @returns {boolean}
+ */
+function isNoVerifyLongFlag(value) {
+  // `--no-` alone is not an attempt at anything; require at least one more char.
+  if (value.length <= '--no-'.length) return false;
+  return '--no-verify'.startsWith(value);
+}
+
 function isCommitNoVerifyShortFlag(value) {
   if (!value.startsWith('-') || value.startsWith('--') || value === '-') {
     return false;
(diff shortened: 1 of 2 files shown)


ntdat812:
Correct, and worth being precise about who broke what: **three of these were already refused on `main`.** Matching abbreviations widened an existing false positive rather than creating one.

Measured through `run-with-flags.js`, `main` vs this branch before the fix:

| command | `main` | branch (before) |
| --- | --- | --- |
| `git push --push-option --no-verify` | **BLOCK** | BLOCK |
| `git push -o --no-verify` | **BLOCK** | BLOCK |
| `git commit --trailer --no-verify -m x` | **BLOCK** | BLOCK |
| `git push --push-option --no-verif` | allow | **BLOCK** ← widened by my change |
| `git commit --trailer --no-verif -m x` | allow | **BLOCK** ← widened by my change |

Git sends those values verbatim and runs its hooks, so all five are legitimate. Fixed at the root in `1c3d6a87` — handling the arity closes the pre-existing three as well as the two I widened.

**commit:** `--trailer` added to `COMMIT_OPTIONS_WITH_VALUE`.

**push:** new `PUSH_OPTIONS_WITH_VALUE` — `-o`, `--push-option`, `--receive-pack`, `--exec`, `--repo`, `--force-with-lease`. `push` had no arity handling at all; the `skipNext` branch was gated on `command === 'commit'`.

One thing I made sure not to break while fixing it: the inline `=value` form consumes nothing, so a real flag after it must still be caught, or the arity handling becomes its own bypass.

```
git push --push-option=ci.skip --no-verify        -> BLOCK
git commit --trailer=Key:val --no-verif -m "msg"  -> BLOCK
```

Both have tests.

## Verification

Against `origin/main` with only the hook reverted: **35 passed, 11 failed** — the 6 abbreviation cases plus the 5 false-positive cases. With the change: **46 passed, 0 failed**. Sibling suite `cursor-block-no-verify`: **14 passed, 0 failed**. `eslint` clean.

ntdat812:
Both of you are right, and this one is entirely mine — I introduced a bypass while fixing a false positive, which is exactly the failure mode I flagged one comment earlier and then walked into.

```
git push --force-with-lease --no-verify   -> ALLOWED   (bypass I created)
git push --force-with-lease --no-verif    -> ALLOWED   (bypass I created)
```

Fixed in `404d6cec` — `--force-with-lease` removed from `PUSH_OPTIONS_WITH_VALUE`. Both are blocked again.

## How I settled the arity this time

By asking git instead of reading docs. Every option **left** in the set answers "requires a value" when given none; the bare lease option parses and fails later on the destination:

```
git push --push-option      -> error: option `push-option' requires a value
git push -o                 -> error: switch `o' requires a value
git push --receive-pack     -> error: option `receive-pack' requires a value
git push --exec             -> error: option `exec' requires a value
git push --repo             -> error: option `repo' requires a value
git push --force-with-lease -> fatal: No configured push destination.
```

That last line is the whole difference: `--force-with-lease` has an **optional, inline-only** value, so `--force-with-lease=main:abc123` is one token and the bare form consumes nothing.

I should have run that check before adding the entry rather than after being told. The five remaining entries are now measured, not assumed.

## Coverage

Four new tests, and I verified they are load-bearing rather than decorative — re-adding `--force-with-lease` to the set fails two of them:

```
✗ still blocks no-verify after --force-with-lease: git push --force-with-lease --no-verify
✗ still blocks no-verify after --force-with-lease: git push --force-with-lease --no-verif
Passed: 48  Failed: 2
```

They cover both spellings after the bare option, the inline `=main:abc123` form followed by `--no-verify`, and a bare `git push --force-with-lease` staying allowed.

**Suite: 50 passed, 0 failed.** Sibling `cursor-block-no-verify`: 14 passed. `eslint` clean.

ntdat812:
Confirmed, and this one is a genuine gap rather than something the earlier commits introduced: I taught the hook that `--no-verif` is `--no-verify`, but left the value-taking option lookup an exact-name match. Git abbreviates both sides.

```
BLOCK  git push --push-opti --no-verify
BLOCK  git push --push-o --no-verif
BLOCK  git commit --trail --no-verify -m "msg"
BLOCK  git commit --mes --no-verify
```

All four are legitimate — the flag becomes the option's *value*. Proven end to end against git 2.51 in a scratch repo with a `pre-commit` hook:

```
$ git commit --mes --no-verify
PRE_COMMIT_RAN
[main 579b3ff] --no-verify

$ git log -1 --pretty=%s
--no-verify
```

The hook ran and `--no-verify` is the commit message.

## Fix

One resolver, `optionConsumesNextValue(value, optionsWithValue)`, shared by the commit and push lookups. It prefix-matches **only** the value-taking options, and **only** when the prefix is unique among them.

Two properties keep that from becoming a bypass:

1. **Ambiguous prefixes resolve to nothing.** `--re` matches both `--receive-pack` and `--repo`, so nothing is skipped and a following `--no-verify` is still caught. Git rejects such a command outright anyway (`error: ambiguous option: re (could be --recurse-submodules or --receive-pack)`), so a prefix that is unique in this set but ambiguous to git can only ever allow a command git refuses to run.
2. **The inline `=value` form is one token** and consumes nothing, so it can never match and the flag after it is still a flag.

I also checked the shadowing case that would make prefix matching unsafe — a *boolean* option whose full name is a strict prefix of a value-taking one, where git would resolve the token differently than this set does. Enumerating `git commit -h` and `git push -h` against both sets: none exists.

## Coverage

Eight tests, and I mutation-tested each group rather than assuming.

Reverting the resolver to the exact lookup fails exactly the four false-positive tests:

```
✗ does not block an abbreviated value-taking option: git push --push-opti --no-verify
✗ does not block an abbreviated value-taking option: git push --push-o --no-verif
✗ does not block an abbreviated value-taking option: git commit --trail --no-verify -m "msg"
✗ does not block an abbreviated value-taking option: git commit --mes --no-verify
Passed: 54  Failed: 4
```

Dropping the uniqueness guard (first match wins) fails exactly the two ambiguity tests:

```
✗ abbreviation handling still blocks: git push --re --no-verify
✗ abbreviation handling still blocks: git commit --f --no-verify -m "msg"
Passed: 56  Failed: 2
```

**Suite: 58 passed, 0 failed.** Sibling `cursor-block-no-verify`: 14 passed. `eslint` clean.

chenhz01:
Prefix resolution is the right fix, and the failure direction is safe: an incomplete value-option list only false-blocks (`git push --push-opti --no-verify` before this PR), it never lets a real bypass through. Two maintenance notes: (1) the sets are hand-maintained per command — main's commit set already drifted once (no `--trailer`), and git keeps adding value options; a CI check that parses `git commit --help` / `git push --help` long options and diffs them against these sets catches the next drift mechanically instead of by bug report. (2) The uniqueness early-`return false` is only sound because git itself rejects ambiguous prefixes across the full option set, not just value-taking ones — worth a comment cross-referencing that guarantee so a future refactor doesn't drop the assumption. Available as a paid engagement too, if a delivered PR beats a review comment for you.

Issue #2834, opened 2026-08-21
Hook overhead ~1.6s per Bash call (npx, missing `if` filters), and `skill-stocktake` frontmatter is missing `name:`


alissonaguiar:
Three findings from profiling ECC v1.9.0 on macOS 15.5 / Claude Code 2.1.238, Node via nvm. Happy to send a PR for any of them.

## 1. `block-no-verify` is invoked through `npx` — 787ms of avoidable overhead per Bash call

`hooks/hooks.json` PreToolUse/Bash runs `npx block-no-verify@1.1.2`. Measured, average of 3 runs with a synthetic PreToolUse payload:

| invocation | time |
|---|---|
| `npx block-no-verify@1.1.2` | **856 ms** |
| the same cached binary directly (`~/.npm/_npx/<hash>/node_modules/.bin/block-no-verify`) | **69 ms** |
| `node -e ""` (spawn floor) | 50 ms |

The package is already in the npx cache, so this is pure npm CLI bootstrap on every single Bash tool call. Because the version is pinned with `@1.1.2`, npx cannot take the PATH shortcut either.

This is also the single hook that users cannot turn off: it is the only Bash hook not wrapped in `run-with-flags.js`, so neither `ECC_HOOK_PROFILE` nor `ECC_DISABLED_HOOKS` reaches it. And Claude Code's settings schema has no way to disable an individual plugin-provided hook — only `disableAllHooks`. So the options today are "keep paying 856ms" or "disable every hook in the plugin".

Suggested fix: resolve the binary directly (or vendor the check — it is a substring match on `--no-verify`), and route it through `run-with-flags.js` like the rest.

## 2. `hooks.json` never uses the `if` field, so hooks spawn Node only to no-op

Claude Code's hook schema supports `if` with permission-rule syntax, documented as "Only runs if the tool call matches the pattern. Avoids spawning hooks for non-matching commands." ECC does not use it anywhere.

Measured cost per Bash call, synchronous hooks only (these block the tool call):

| hook | ms |
|---|---|
| `npx block-no-verify` | 856 |
| `pre:governance-capture` | 194 |
| `pre:mcp-health-check` | 150 |
| `pre:insaits-security` | 133 |
| `post:bash:pr-created` | 133 |
| `auto-tmux-dev` | 72 |
| `post:governance-capture` | 72 |
| **total** | **~1610** |

Most of that is spent discovering there is nothing to do:

[...]

PR #2838, opened 2026-08-22
fix(hooks): tighten mcp-health-check matchers to mcp__.* to eliminate no-op spawns

santhiprakash:
## What Changed
- Changed `pre:mcp-health-check` matcher from `*` to `mcp__.*` so Claude Code skips the hook entirely on non-MCP tool calls
- Changed `post:mcp-health-check` matcher from `*` to `mcp__.*` for the same reason
- Added a regression test verifying both matchers are `mcp__.*`

## Why This Change
Per #2834, `pre:mcp-health-check` and `post:mcp-health-check` both used `matcher: "*"` (catch-all), causing Node to spawn on every tool call only to no-op when the tool name does not start with `mcp__`. The reporter measured ~150ms and ~194ms overhead per Bash call respectively.

Tightening the matcher to `mcp__.*` at the hooks.json level means Claude Code does not even evaluate these hook groups for non-MCP tools, eliminating the process spawn entirely. Behavior is identical for matching MCP tool calls.

## Scope
- Findings 1 and 3 from #2834 were verified already fixed on main and are not addressed here.
- `pre:governance-capture` is env-gated (`ECC_GOVERNANCE_CAPTURE=1`) at the script level; the Claude Code `if` field uses permission-rule syntax and cannot check env vars, so no config-level guard is possible. The matcher is already `Bash|Write|Edit|MultiEdit`.
- `auto-tmux-dev` and the insaits wrapper are not standalone hooks.json entries (they are sub-hooks within the bash dispatcher or unregistered scripts respectively).

## Testing Done
- `node tests/hooks/hooks.test.js` — 250/250 passed
- New test: `MCP health-check hooks use tight mcp__.* matcher instead of wildcard`

## Type of Change
- [x] `fix:` Bug fix
- [x] `test:` Tests

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -73,7 +73,7 @@
         "id": "pre:config-protection"
       },
       {
-        "matcher": "*",
+        "matcher": "mcp__.*",
         "hooks": [
           {
             "type": "command",
@@ -162,7 +162,7 @@
     ],
     "PostToolUseFailure": [
       {
-        "matcher": "*",
+        "matcher": "mcp__.*",
         "hooks": [
           {
             "type": "command",
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -73,7 +73,7 @@
         "id": "pre:config-protection"
       },
       {
-        "matcher": "mcp__.*",
+        "matcher": "^mcp__",
         "hooks": [
           {
             "type": "command",
@@ -162,7 +162,7 @@
     ],
     "PostToolUseFailure": [
       {
-        "matcher": "mcp__.*",
+        "matcher": "^mcp__",
         "hooks": [
           {
             "type": "command",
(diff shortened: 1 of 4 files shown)


santhiprakash:
Closing to keep the review queue clean — this has been open a while without traction. Happy to reopen or resubmit a fresh version if it would still be useful.

chenhz01:
The thread already settled the matcher question; one scope concern I don't see raised: Matcher tightening is the correct layer to fix #2834 — no spawn beats a fast no-op. One scope concern: the same PR also rewrites `DESTRUCTIVE_SQL_DD` in `gateguard-fact-force.js` (the `dd if=` word-boundary escape referencing #2642, restructuring the alternation into `\b(?:drop\s+table|delete\s+from|truncate)\b|\bdd\s+if=`). That is a separate behavioral fix with its own issue and deserves its own PR — both for review clarity and because #2755 and #3052 are concurrently editing the same file. As bundled, a future bisect on a gate regression cannot cleanly attribute matcher versus regex changes. The regression test asserting both matchers is a good addition; keeping it with the matcher-only diff would make it the whole story.

PR #2844, opened 2026-08-24
feat(skills): add skill profile selection to stay in listing budget

### LINKED ISSUE #2694: Skills have no selection mechanism (hooks do) — 773 skills overflow the skill-listing budget
## Summary

ECC ships 773 skills and publishes all of them unconditionally. On Claude Code that overflows the skill-listing budget, which truncates the listing — so skills become invisible to routing, including skills from *other* plugins and the user's own project commands.

Hooks already have exactly the mechanism that would fix this. Skills don't.

## Measurements

Taken against `ecc@2.0.0-rc.1` on Claude Code 2.1.223:

- `find skills -name SKILL.md` → **773 files**
- Sum of frontmatter `name` + `description` across those 773 → **~170,000 characters, ~42,500 est. tokens** (chars/4)
- Claude Code budgets the skill listing at roughly 1% of the context window and truncates past it. On a 1M-token window that is ~10,000 tokens, so ECC alone is roughly **4x the whole budget** — before any other plugin, project command, or user skill is counted.

For comparison, everything else in my install came to ~1,600 est. tokens combined (superpowers 14 skills, 12 project commands, 4 user skills).

## Why this matters

The listing is what the model routes on. When it is truncated:

- Skills that are cut simply cannot be selected — they are not "low priority", they are absent.
- Which ones get cut is not something a user can see, predict, or control.
- It is not just ECC's own skills at risk. The budget is shared, so a single plugin publishing 773 entries can crowd out a user's own project commands.

I found this while auditing my own setup: I have 12 project commands that are the core of a personal workflow, and I have no way to know whether they are still visible to routing.

## The asymmetry with hooks

ECC solves this properly for hooks. `scripts/lib/hook-flags.js` provides:

- `ECC_HOOK_PROFILE=minimal|standard|strict` (default `standard`)
- `ECC_DISABLED_HOOKS=comma,separated,hook,ids`

and each hook in `hooks/hooks.json` carries a profile CSV, so `minimal` drops 17 of the 18 flag-gated hooks. That is a good design and it works.

Skills have no equivalent. `.claude-plugin/plugin.json` declares:

```json
"skills": ["./skills/"]
```

One directory, all 773, no groups, no profiles, no opt-out.

## What users are left with

None of these are viable:

1. **Claude Code's `skillOverrides`** — takes exact skill names, so this means 773 entries in `settings.json`.
2. **Deleting directories from the plugin cache** — overwritten on the next plugin update.
3. **Disabling ECC entirely** — throws away a plugin I use constantly (6,000+ dispatches in my install).

So today the only real option is "accept the overflow".

## Suggested fix

Mirror the hook design:

- `ECC_SKILL_PROFILE=minimal|standard|full`, with skills tagged by profile — a `minimal` set covering the commonly-used core.
- and/or `ECC_ENABLED_SKILL_GROUPS` / `ECC_DISABLED_SKILL_GROUPS` keyed on the existing `skills/` subdirectory structure, which already groups things naturally.

Either would let users bring ECC's listing footprint down to something that fits the budget alongside their other tooling, without giving up the plugin.

`docs/capability-surface-selection.md` already argues for "the narrowest surface that ... keeps token cost under control" and "prefer the lower token overhead". This is that principle applied to ECC's own skill surface.

## Environment

- ECC 2.0.0-rc.1
- Claude Code 2.1.223, native install, Windows 11

LovePlayCode:
## Summary
- Fixes #2694: Claude Code was listing every curated skill (`./skills/`), which overflows the shared skill-listing budget and hides ECC skills plus the user's own project commands.
- Adds hook-style controls: `ECC_SKILL_PROFILE=minimal|standard|full` (default `standard`), plus `ECC_ENABLED_SKILL_GROUPS`, `ECC_DISABLED_SKILL_GROUPS`, and `ECC_DISABLED_SKILLS`. Groups are the existing install-module IDs.
- Default plugin listing is now the standard subset (80 skills, under the ~10k-token budget). `ecc skills-profile --profile full` restores the previous wholesale `./skills/` listing. `install-apply` is unchanged unless `--skill-profile` is passed explicitly.

## What Changed
- New `scripts/lib/skill-flags.js` and `scripts/lib/skill-catalog.js` plus `manifests/skill-profiles.json`.
- `scripts/apply-skill-profile.js` / `ecc skills-profile` writes the selected paths into `.claude-plugin/plugin.json`.
- Optional `--skill-profile` filter on `install-apply`.
- Plugin `userConfig.skill_profile` and README / capability-surface docs.

## Why This Change
Hooks already have `ECC_HOOK_PROFILE`. Skills had no equivalent, so the plugin published the entire catalog unconditionally. That is the defect reported in #2694.

## Testing Done
- [x] Automated tests pass locally for the new suites and the touched install/plugin surfaces (`skill-flags`, `skill-catalog`, `apply-skill-profile`, `skill-profiles`, `install-manifests`, `install-request`, `plugin-manifest`, `ecc`, `npm-publish-surface`, `catalog`)
- [ ] Manual testing completed in a live Claude Code session
- [x] Edge cases considered and tested (invalid profile fallback, group opt-in/out, extra `--skills` force-include, `full` keeps `./skills/`)

## Type of Change
- [ ] `fix:` Bug fix
- [x] `feat:` New feature
- [ ] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `ci:` CI/CD changes

- [x] Added `scripts/apply-skill-profile.js` to `package.json` `files` and the publish-surface allowlist. No dependency or lockfile change is required for that path-only `files` update.

- [x] Wired `ecc skills-profile` through `scripts/ecc.js` and the publish allowlist
- [ ] Not a new skill/agent/hook, so catalog/command-registry/docs tables were not regenerated

## Documentation
- [x] README hook/env section and `docs/capability-surface-selection.md` updated

## Test plan
- [ ] Install the plugin from this branch and confirm Claude Code lists the standard subset, not all 286 skills
- [ ] `ecc skills-profile --profile minimal` then confirm the listing shrinks to the daily-core set
- [ ] `ECC_ENABLED_SKILL_GROUPS=framework-language ecc skills-profile` adds that group without switching to `full`
- [ ] `ecc skills-profile --profile full` restores `"skills": ["./skills/"]`
- [ ] `node scripts/install-apply.js --profile full --dry-run` still plans every skill; adding `--skill-profile minimal` drops full-only skills such as `tasteforge-video`

Made with [Cursor](https://cursor.com)


haelyra:
Thank you for tackling the skill-listing budget directly. The measured need is real: publishing the full catalog into every harness overwhelms discovery, and users need a small default with an explicit path to the complete surface. Together with #2788, this is strong community evidence for budgeted context profiles.

This implementation should not land as a second profile authority beside the accepted ECC 2-to-3 direction. We would love a focused port with these changes:

- **Project onto the canonical `lean@1` and `full@1` registry.** This PR creates `minimal`, `standard`, and `full` plus a separate manifest and group-ranking model, while the master plan reserves the initial base-profile authority to versioned `lean@1` and `full@1`. Success is one registry and vocabulary; module groups and adapter-specific subsets are compiler inputs or projections, not additional base profiles.
- **Keep context selection separate from installation.** `--skill-profile` currently filters installer copy operations, so choosing a smaller listing also removes content from disk and changes repair/uninstall state. A context profile answers what is loaded; install modules answer what is present. Keep the canonical skill store complete, then materialize a harness carrier that exposes only the selected profile.
- **Replace manifest mutation with receipt-bound carrier activation.** `apply-skill-profile.js` rewrites the installed `.claude-plugin/plugin.json` in place, with no before-image, destination binding, profile digest, rollback, crash recovery, or protection from auto-update replacement. Success is an atomic carrier switch with an immutable receipt that names the destination, registry version, exact selected skill digests, prior state, and deterministic rollback/recovery behavior.
- **Enforce a truthful token ledger, not a skill count estimate.** The PR describes 80 skills as under an approximate 10K listing budget but does not measure the actual harness payload or fail when descriptions change. The canonical Lean gate is a hard 8K budget. Generate and test the provider-specific listing payload, record the token method and version, and fail closed when the selected projection exceeds its declared budget.
- **Keep environment and persisted state precedence explicit.** `ECC_SKILL_PROFILE` only influences the later rewrite command; it cannot dynamically change Claude discovery, while managed config and plugin options can disagree with the already-mutated manifest. Preview must show requested versus effective profile, source precedence, destination, and digest before activation, and doctor must report drift honestly.

The cleanest contribution would port the catalog grouping, deterministic selection tests, and Claude listing projection into the existing M1 profile train, preserving your authorship, rather than maintaining this parallel state model. If you do not have bandwidth for that port, we will add those pieces to the post-release profile task queue and carry them forward with attribution.

LovePlayCode:
Thank you for the review — this is the right call.

I agree this should not land as a second profile authority beside the ECC 2-to-3 `lean@1` / `full@1` direction. After syncing latest `main` and the current remote branches, that registry, carrier, and receipt model are not in the tree yet, so I will not invent a parallel `lean@1` / `full@1` surface in this PR.

Please add the salvageable pieces to the post-release profile task queue and carry them forward with attribution:

- catalog grouping
- deterministic skill-selection tests
- Claude listing projection (keep the canonical store complete; expose a harness listing subset)

Happy to close or convert this PR if that makes the queue cleaner. I will not push further commits on the `minimal` / `standard` / `full` path.

chenhz01:
Converging with the lean@1/full@1 registry direction makes sense. One transitional gap while both exist: `plugin.json` now hardcodes the ~80-entry standard list against a skills tree that gains entries most weeks — a merged skill PR that doesn't re-run `apply-skill-profile.js` silently stays unlisted, and a renamed or deleted skill leaves a dangling path in the manifest. Until the registry owns this, a small CI guard keeps the static list honest:

```js
assert.deepStrictEqual(
  JSON.parse(read('.claude-plugin/plugin.json')).skills,
  computeStandardProfile(require('../manifests/skill-profiles.json'), readdir('skills')));
```

That also gives the future registry migration a red/green baseline to migrate from. Happy to open a PR with the guard if useful.

PR #2845, opened 2026-08-24
fix(config-protection): protect ignore files and current config spellings

ntdat812:
## The gap

`config-protection` exists so an agent fixes the code instead of weakening the checks. It blocks the config files — but not the cheapest way to make a linter pass, which is adding one path to an ignore file. Measured against the hook with the file already on disk:

```
.eslintrc.json       -> BLOCK
.eslintignore        -> allow      <-- silences any failing file
.prettierignore      -> allow
.stylelintignore     -> allow
.markdownlintignore  -> allow
```

`echo "src/failing-file.ts" >> .eslintignore` achieves exactly what editing `rules` would, and the guard never sees it.

The second gap is naming. Only the legacy spellings were listed, so a project on the current ones was unprotected end to end:

```
stylelint.config.js       -> allow      (the spelling stylelint documents)
stylelint.config.mjs      -> allow
.stylelintrc.js           -> allow
.markdownlint.jsonc       -> allow
.markdownlint-cli2.jsonc  -> allow      (markdownlint-cli2's own config name)
```

## The change

Adds the four ignore files, the `stylelint.config.{js,cjs,mjs}` and `.stylelintrc.{js,cjs,mjs,yaml}` spellings, `.markdownlint.{jsonc,yml}`, and the four `.markdownlint-cli2.*` names.

No logic changes: everything goes through the same `PROTECTED_FILES` set, so the existing behaviour carries over unchanged — case-insensitive matching, the truncated-payload refusal, and (the one that matters here) **first-time creation is still allowed**. Scaffolding a brand-new `.prettierignore` in a project that has none is unaffected; only editing one that already exists is blocked, which is the same rule the configs already follow.

I did not add `pyproject.toml`, for the reason the file already documents: it carries project metadata alongside the ruff config.

## Tests

Three tests in `tests/hooks/config-protection.test.js`, all through `run-with-flags.js` with real files on disk:

- the four ignore files, each asserted on exit code **and** on the `BLOCKED: Modifying <name>` message
- the five current config spellings
- the two cases that must stay allowed: a first-time `.prettierignore` write, and `.eslintignore.bak`, which must not be swept up by a looser match

Against `main` with only `scripts/hooks/config-protection.js` reverted:

```
✗ blocks edits to an existing linter ignore file
✗ blocks the current stylelint and markdownlint config spellings
Results: Passed: 10, Failed: 2
```

The third test passes on both, which is what it is for — it pins the allow-path so a later widening cannot quietly start blocking bootstrap writes.

With the change: **12 passed, 0 failed.** `eslint` clean on both files.


ntdat812:
Both right, and I checked each name against the tool's own resolution order rather than taking the list as given:

- **stylelint** resolves through cosmiconfig, which searches `stylelint.config.{js,cjs,mjs,ts,mts,cts}` — I had only the first three.
- **markdownlint** reads `.markdownlint.{json,jsonc,yaml,yml,cjs,mjs}` — I had everything except the two script forms.

Added all five in `9c8f5b0`, each covered by the existing-file runner test: `stylelint.config.ts`, `stylelint.config.mts`, `stylelint.config.cts`, `.markdownlint.cjs`, `.markdownlint.mjs`.

On the markdownlint-cli2 point: correct that this repo pins `markdownlint-cli` `0.48.0`, so `.markdownlint-cli2.*` is dead weight *here*. I kept those four names because this hook ships as a plugin and runs against consumer projects, where cli2 is common — and an unused entry in a denylist costs nothing, while a missing one costs the whole guard. Happy to drop them if you would rather the list describe only this repo.

**12 passed, 0 failed.** `eslint` clean.

chenhz01:
Closing the ignore-file gap is the important half — but as the test comment itself notes, first-time creation stays allowed. That leaves the exact bypass one `Write` away: an agent blocked from editing `.eslintrc.json` can create a fresh `.eslintignore` containing `src/failing-file.ts` in a single step, since the existence check only guards files already on disk. Ignore files are the one category where creation is itself the attack, so they need creation gating too:

```js
const isCreate = !fs.existsSync(target);
if (IGNORE_FILES.has(basename(target)) && isCreate && content.trim()) return block();
```

If this becomes a priority, I contract for exactly this kind of fix — fixed price, working code.

PR #2878, opened 2026-08-25
fix(continuous-learning-v2): count every instinct extension in observer status (#2859)

### LINKED ISSUE #2859: fix(continuous-learning-v2): start-observer status counts only *.yaml, reports Instincts: 0 for .md instincts
## Summary

`start-observer.sh status` reports `Instincts: 0` while instincts exist and are being loaded. The status counter globs `*.yaml`, but the observer prompt explicitly instructs the analyzer to write `<id>.md`, and `instinct-cli.py` accepts `.yaml`, `.yml`, and `.md`. So the one command an operator runs to confirm that learning works under-reports to zero.

This matters mainly because it removes the signal you would use to detect a silent observer failure such as #2673: an operator who checks `status`, sees `Instincts: 0`, and concludes nothing has been learned cannot tell that apart from instincts existing but not being counted.

## Reproduction

ECC 2.1.0 (`ff15079`), macOS (Darwin 24.6.0), zsh, observer model `haiku`.

With 11 `.md` instincts on disk, both commands run against the same project directory at the same moment:

```
$ find "$INSTINCTS_DIR" -name '*.md' | wc -l
11
$ find "$INSTINCTS_DIR" -name '*.yaml' | wc -l
0

$ bash skills/continuous-learning-v2/agents/start-observer.sh status
Project: <project> (<hash>)
Observations:       21 lines
Instincts:        0          <-- wrong

$ python3 skills/continuous-learning-v2/scripts/instinct-cli.py status
  INSTINCT STATUS - 11 total
  Project instincts: 11      <-- correct
```

## Root cause

The producer writes `.md` by explicit instruction — `agents/observer-loop.sh:166`:

```
If you find 3+ occurrences of the same pattern, you MUST write an instinct file directly to ${INSTINCTS_DIR}/<id>.md using the Write tool.
```

The consumer accepts three extensions — `scripts/instinct-cli.py:130`:

```python
ALLOWED_INSTINCT_EXTENSIONS = (".yaml", ".yml", ".md")
```

The status counter accepts one — `agents/start-observer.sh:160`:

```sh
instinct_count=$(find "$INSTINCTS_DIR" -name "*.yaml" 2>/dev/null | wc -l)
```

A scan of `skills/continuous-learning-v2/` shows this is the only remaining yaml-only assumption in production code; the other two hits are in `scripts/test_parse_instinct.py`, which creates its own `.yaml` fixtures and is unaffected.

## Suggested fix

Align the counter with `ALLOWED_INSTINCT_EXTENSIONS`:

```sh
instinct_count=$(find "$INSTINCTS_DIR" \( -name "*.yaml" -o -name "*.yml" -o -name "*.md" \) 2>/dev/null | wc -l)
```

Applied locally and `status` then reports `Instincts: 11`, matching `instinct-cli.py`.

If the intent is instead for the analyzer to emit YAML, the fix belongs at `observer-loop.sh:166` — but every instinct written by the current prompt is Markdown with YAML frontmatter, so changing the counter is the non-breaking direction.

## Notes

Found while investigating why a freshly enabled observer had produced no instincts. The underlying cause there was #2673; this counter behaviour is a separate defect that made that diagnosis harder, and it stays wrong even after #2753 merges, since that PR touches only `observer-loop.sh` and its test.

ntdat812:
Fixes #2859.

## What Changed

`start-observer.sh status` counted instinct files with `find "$INSTINCTS_DIR" -name "*.yaml"`. It now counts every extension the loader accepts, the same way the loader enumerates them:

```sh
instinct_find_expr=( \( -iname "*.yaml" -o -iname "*.yml" -o -iname "*.md" \) )
instinct_count=$(find "$INSTINCTS_DIR" -maxdepth 1 -type f "${instinct_find_expr[@]}" 2>/dev/null | wc -l | tr -d "[:space:]")
```

Three things had to match `_load_instincts_from_dir` in `scripts/instinct-cli.py`, not one:

| loader | counter |
| :-- | :-- |
| `suffix.lower() in ALLOWED_INSTINCT_EXTENSIONS` → `.yaml`, `.yml`, `.md` | all three, and `-iname` for the case-insensitive compare |
| `Path.iterdir()` → top level only | `-maxdepth 1` (plain `find` recurses) |
| `file.is_file()` → directories skipped | `-type f` |

`tr` drops the column padding BSD `wc` emits, which is why the reported output read `Instincts:        0` with that gap.

## Why This Change

The producer writes `.md` **by explicit instruction** — `agents/observer-loop.sh:166` tells the analyzer to write `${INSTINCTS_DIR}/<id>.md` — and the consumer accepts three extensions. Only the status counter accepted one, so the single command an operator runs to confirm that learning is working reported `Instincts: 0` on a perfectly healthy install.

That matters beyond the wrong number: `Instincts: 0` is exactly what a silently dead observer looks like (#2673). An operator who checks `status` cannot tell "nothing has been learned" from "instincts exist and are not being counted", which removes the signal the status check exists to provide.

## Testing Done

**End to end against the shipped script.** A temp homunculus dir (`CLV2_HOMUNCULUS_DIR` + `CLAUDE_PROJECT_DIR`), a live PID in `.observer.pid`, and an instincts dir holding 3 × `.md`, `a.yaml`, `b.yml`, `c.YAML`, a `notes.txt`, and `nested/deep.md`:

```
before:  Instincts: 1     # only a.yaml
after:   Instincts: 6     # the same six files instinct-cli.py loads
```

`notes.txt` and `nested/deep.md` are correctly excluded — the loader skips both.

**New test: `tests/skills/observer-status-instinct-count.test.js`, 11 assertions.** It parses `ALLOWED_INSTINCT_EXTENSIONS` out of `instinct-cli.py` and requires the counter to match *whatever it says*, so adding a fourth extension to the loader fails this test instead of silently under-reporting again. Then it runs the real `start-observer.sh status` for the three scenarios above. Integration coverage skips cleanly on Windows without `ECC_TEST_BASH`, following `tests/skills/repo-scan-install.test.js`.

Mutation-tested:

- restore `-name "*.yaml"` → **fails** (`status count must match .yml`)
- drop `-maxdepth 1` → **fails** (`status count must not recurse — the loader does not`)
- the script exits `1` on drift and `0` when clean, so `run-all.js` picks it up correctly

**On `node tests/run-all.js`:** run twice, same command, same machine — once on a clean `origin/main` and once on this branch:

| | tests | passed | failed |
| :-- | --: | --: | --: |
| `origin/main` | 3869 | 3816 | 53 |
| this branch | 3880 | 3827 | 53 |
| delta | **+11** | **+11** | **0** |

The +11 is exactly this file. The 53 failures are byte-identical between the two runs — the same eight groups (`lib/claude-plugin-setup`, `lib/claude-scope-migration`, `lib/codex-legacy-sync`, `lib/memory-vault`, `lib/state-store`, `scripts/ecc-universal-bin`, `scripts/memory-mcp`, `scripts/setup`), none of which read `skills/continuous-learning-v2/`. They are pre-existing on this machine, so I have left the "tests pass locally" box unticked rather than tick it on a suite that is not green — but this PR adds nothing to that number. `npx eslint` is clean on the new file.

**Review follow-up (second commit).** The static assertions originally ran at the top level, so a failure exited before the `Passed:`/`Failed:` tokens `run-all.js` totals — it still went red via the exit code, but the per-case counts were lost and you only ever saw the first broken expectation. Every case now goes through the same `runTest()` wrapper the integration cases already used, and the case list is built inside a try so a renamed `instinct-cli.py` is a reported failure rather than a crash. Reverting the fix now prints `Passed: 4, Failed: 7` and exits 1, naming all seven.

## Type of Change
- [x] `fix:` Bug fix
- [ ] `feat:` New feature
- [ ] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `ci:` CI/CD changes

- [x] JSON files validate cleanly (none changed)
- [x] Shell scripts pass shellcheck (if applicable) — `bash -n` clean; the change is one `find` expression

Not applicable — no new component. One line changed in an existing skill script, plus one test file under `tests/skills/`, which `run-all.js` discovers by glob.

## Documentation
- [x] Updated relevant documentation — the comment above the counter records *why* the three extensions and the depth/case flags have to match the loader, so the next person editing it does not narrow it back.
- [ ] README updated (if needed) — not needed

### DIFF
--- a/skills/continuous-learning-v2/agents/start-observer.sh
+++ b/skills/continuous-learning-v2/agents/start-observer.sh
@@ -156,8 +156,14 @@ case "$ACTION" in
         echo "Observer is running (PID: $pid)"
         echo "Log: $LOG_FILE"
         echo "Observations: $(wc -l < "$OBSERVATIONS_FILE" 2>/dev/null || echo 0) lines"
-        # Also show instinct count
-        instinct_count=$(find "$INSTINCTS_DIR" -name "*.yaml" 2>/dev/null | wc -l)
+        # Also show instinct count. Count every extension the loader accepts
+        # (ALLOWED_INSTINCT_EXTENSIONS in scripts/instinct-cli.py) - the
+        # observer prompt tells the analyzer to write "<id>.md", so a
+        # *.yaml-only count reports 0 on a working install. Depth and case
+        # match the loader's iterdir() + suffix.lower(): top level only,
+        # case-insensitive. tr strips the padding BSD wc emits.
+        instinct_find_expr=( \( -iname "*.yaml" -o -iname "*.yml" -o -iname "*.md" \) )
+        instinct_count=$(find "$INSTINCTS_DIR" -maxdepth 1 -type f "${instinct_find_expr[@]}" 2>/dev/null | wc -l | tr -d "[:space:]")
         echo "Instincts: $instinct_count"
         exit 0
       else
--- a/tests/skills/observer-status-instinct-count.test.js
+++ b/tests/skills/observer-status-instinct-count.test.js
@@ -1,0 +1,145 @@
+/**
+ * Regression tests for #2859: `start-observer.sh status` counted only *.yaml.
+ *
+ * The producer writes `<id>.md` (agents/observer-loop.sh instructs the analyzer
+ * to) and the loader accepts .yaml/.yml/.md (ALLOWED_INSTINCT_EXTENSIONS in
+ * scripts/instinct-cli.py), so the one command an operator runs to confirm that
+ * learning works reported `Instincts: 0` on a working install — and an operator
+ * cannot tell that apart from a silently dead observer.
+ */
+
+'use strict';
+
+const assert = require('assert');
+const { spawnSync } = require('child_process');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+const repoRoot = path.resolve(__dirname, '..', '..');
+const skillRoot = path.join(repoRoot, 'skills', 'continuous-learning-v2');
+const observerScript = path.join(skillRoot, 'agents', 'start-observer.sh');
+const detectProject = path.join(skillRoot, 'scripts', 'detect-project.sh');
+const instinctCli = path.join(skillRoot, 'scripts', 'instinct-cli.py');
+const bashBinary = process.env.ECC_TEST_BASH || (process.platform === 'win32' ? null : 'bash');
+
+let passed = 0;
+
+function toShellPath(filePath) {
+  const normalized = filePath.split(path.sep).join('/');
+  return normalized.replace(/^([A-Za-z]):\//, (_, drive) => `/${drive.toLowerCase()}/`);
+}
+
+// ── The counter must accept every extension the loader accepts ──
+
+const cliSource = fs.readFileSync(instinctCli, 'utf8');
+const allowedMatch = cliSource.match(/ALLOWED_INSTINCT_EXTENSIONS\s*=\s*\(([^)]*)\)/);
+assert.ok(allowedMatch, 'ALLOWED_INSTINCT_EXTENSIONS not found in instinct-cli.py');
+const allowedExtensions = allowedMatch[1]
+  .split(',')
+  .map(part => part.trim().replace(/^["']|["']$/g, ''))
+  .filter(Boolean);
+assert.ok(allowedExtensions.length >= 3, `expected several extensions, got ${allowedExtensions}`);
+passed++;
+
+const observerSource = fs.readFileSync(observerScript, 'utf8');
+const statusCount = observerSource
+  .split('\n')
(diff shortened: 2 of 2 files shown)


chenhz01:
Fix and regression coverage both look right. One brittleness in the new test: `readAllowedExtensions()` extracts `ALLOWED_INSTINCT_EXTENSIONS` by regex-matching the Python source text (`/\(([^)]*)\)/`), which breaks the moment the tuple is reformatted across lines, renamed, or switched to a different container — the test would fail for cosmetic reasons on an otherwise-correct loader. Since these tests already invoke `instinct-cli.py`, consider exposing the list (`--print-allowed-extensions`) and asserting against real output instead of source text; same idea for `readStatusCounter()`. The find-expression and maxdepth behavior itself matches the loader as described. If this becomes a priority, I contract for exactly this kind of fix — fixed price, working code.

PR #2879, opened 2026-08-25
fix(docs): stop translated agent docs from contradicting the shipped agent

### LINKED ISSUE #2861: main is red: skills/ito-inference and skills/ito-training unreferenced by any install module (fd27a0ec)
## Summary

`main` is currently red. `scripts/ci/validate-install-manifests.js` exits 1 because two curated skills are not referenced by any install module.

## Reproduction

```
$ node scripts/ci/validate-install-manifests.js
ERROR: curated skill skills/ito-inference is not referenced by any install module
ERROR: curated skill skills/ito-training is not referenced by any install module
$ echo $?
1
```

This also fails the aggregate run:

```
$ node tests/run-all.js
... every other group reports Failed: 0 ...
ERROR: curated skill skills/ito-training is not referenced by any install module
```

## Root cause

fd27a0ec (#2700) added the two skills but registered them nowhere:

```
$ git show --stat fd27a0ec
 skills/ito-inference/SKILL.md | 59 ++++++++++
 skills/ito-training/SKILL.md  | 60 ++++++++++
 2 files changed, 119 insertions(+)
```

Two files, both `SKILL.md`. Missing at minimum:

- `manifests/install-modules.json` — no module lists either path, which is what `validate-install-manifests.js:168` fails on
- `package.json` `files` — both absent, so neither skill would publish to npm even once the validator passes

This is the checklist at `.github/PULL_REQUEST_TEMPLATE.md` lines 33-39 ("Registered in `package.json` (`bin` and `files`), `manifests/install-components.json`, `manifests/install-modules.json`...") not being run, rather than a deliberate exclusion.

## Why this is not a one-line fix

The natural home is the `ito-compute` module — both skills state they chain off it:

> ito-inference: "Use after ito-compute has booked GPU nodes and the user wants an OpenAI-compatible endpoint on that metal."
> ito-training: "Use after ito-compute has booked GPU nodes and the user wants pre-training, fine-tuning, or RL on that metal."

But `tests/ci/ito-compute-skill.test.js` pins that module exactly:

```js
:93   assert.deepStrictEqual(module.paths, ["skills/ito-compute"]);
:107  description: "Authenticated Itô GPU inventory, RFQ, status, and explicitly gated
                     node-qualification workflows through the separately installed canonical CLI.",
```

I confirmed locally that adding the two paths to `ito-compute` turns the validator green and immediately fails that assertion. Given how tightly this module is pinned (paths, dependencies, defaultInstall, stability, component description, profile membership), the pin looks intentional, so extending it is a decision for a maintainer rather than a drive-by fix.

## Options

**A. Extend `ito-compute`.** Add both paths, update `:93` and the component description at `:107` and in `manifests/install-components.json`, add both to `package.json` `files`. Smallest conceptual change and matches how the skills describe themselves — but relaxes a deliberately pinned contract, and the `capability:ito-compute` description would need to widen from "inventory, RFQ, status, node-qualification" to include running workloads.

**B. New module (e.g. `ito-workloads`).** Own module + `capability:` component + profile membership, depending on `ito-compute`. Keeps the `ito-compute` contract untouched and lets the two workload skills be installed independently. More wiring, and adds a module for two skills.

**C. Add to `INTENTIONALLY_UNSHIPPED_SKILL_IDS`** (`scripts/ci/validate-install-manifests.js:21`). One line, turns trunk green now. Only correct if these are not yet meant to ship; the commit that added them suggests otherwise, so this reads as a deferral rather than a fix.

## Notes

I have not opened a PR because A, B, and C encode different intentions about whether and how these skills ship, and I did not want to guess at the reason `ito-compute` is pinned. Happy to send a PR for whichever you prefer — A and C are both small.

Unrelated but noticed while verifying: `docs/zh-TW/agents/doc-updater.md` and the other locale copies carry `model: opus` where `agents/doc-updater.md` declares `model: haiku`, so the translated agent docs have drifted from canonical.

ntdat812:
Noticed while verifying #2861, whose reporter flagged one instance of this in passing (`docs/zh-TW/agents/doc-updater.md` declaring `model: opus` where the canonical agent declares `haiku`). It is not one file — it is 54.

## What Changed

Sync `model` and the `tools` **set** in `docs/<locale>/agents/*.md` to the canonical `agents/*.md`, and add a CI test so it cannot drift back.

`scripts/ci/validate-agents.js` reads only `AGENTS_DIR = agents/`. The seven translated copies were never validated against the agent they describe, and drifted in two machine-checkable ways:

**Model tier — 39 files.** Every single one names a *costlier* tier than ships, which is the signature of translations made from an older revision and never re-synced:

| agent | canonical | locale docs |
| :-- | :-- | :-- |
| `security-reviewer` | `sonnet` | `opus` (ja-JP, zh-TW) |
| `code-reviewer` | `sonnet` | `opus` (ja-JP, zh-TW) |
| `tdd-guide` | `sonnet` | `opus` (ja-JP, zh-TW) |
| `doc-updater` | `haiku` | `opus` (ja-JP, zh-TW) |
| `docs-lookup` | `haiku` | `sonnet` (es, ja-JP, tr, zh-CN) |
| … 14 more | | |

**Tool set — 15 files.** These are the ones that actually mislead:

- `security-reviewer` and `database-reviewer` list `Write` and `Edit` in **all seven locales**. Both agents ship read-only (`Read, Grep, Glob, Bash`).
- `seo-specialist` in zh-CN lists a `Bash` the canonical agent does not have.

Documenting a *reviewer* as able to write is the kind of statement someone auditing what these agents can touch would act on, and it is wrong in every translation of the file.

## Why This Change

`agents/*.md` is what ships and what the validator checks; the locale copies are translations of it. Where the frontmatter disagrees, the translation is simply false — a reader of `docs/ja-JP/agents/security-reviewer.md` believes the security reviewer runs on opus and can edit files. Neither is true.

Scope is deliberately narrow: **only the fields that carry meaning**. The `["a", "b"]` vs `a, b` list-style difference is consistent per locale and semantically identical, so it is left alone — folding it in would have buried 58 real corrections under ~200 lines of formatting churn. The diff is 58 changed lines across 54 files, and every one of them is a value that was wrong.

`.kiro/agents/` is excluded on purpose: it is a different schema (`allowedTools: [read, shell]`, no `model`), not a translation of this one.

## Testing Done

**New test: `tests/ci/locale-agent-frontmatter.test.js`.** Checks 199 locale docs against 68 agents:

```
  ✓ there are locale agent docs to check
  ✓ every locale agent doc has parseable frontmatter
  ✓ locale agent docs name the same agent as canonical
  ✓ locale agent docs declare the canonical model tier
  ✓ locale agent docs declare the canonical tool set
  ✓ a read-only reviewer is never documented with Write or Edit
```

Design notes:

- `tools` is compared as a **set**, so the per-locale list style stays legal and only a real difference in membership fails.
- Prose is never compared — these are translations.
- A field absent on one side is skipped rather than flagged, so a locale copy that simply omits `model` is not forced to add one.
- The last case pins the specific failure this was written for, independent of the general rule.
- Failure output names every offending file and the exact mismatch, so fixing it does not require re-deriving the diff.

Mutation-tested: reverting a single file (`docs/ja-JP/agents/security-reviewer.md`) fails **3 of the 6 cases** and the script exits `1`; with the tree clean it exits `0`.

Also run: `node scripts/ci/validate-agents.js` → `Validated 68 agent files`, exit 0. `node scripts/ci/validate-skills.js` → exit 0. `eslint .` clean (it is the first half of `npm run lint`, which reached `markdownlint`).

**On `node tests/run-all.js`:** run twice, same command, same machine — once on a clean `origin/main` and once on this branch:

| | tests | passed | failed |
| :-- | --: | --: | --: |
| `origin/main` | 3869 | 3816 | 53 |
| this branch | 3876 | 3823 | 53 |
| delta | **+7** | **+7** | **0** |

The +7 is exactly this file. The 53 failures are byte-identical between the two runs — the same eight groups (`lib/claude-plugin-setup`, `lib/claude-scope-migration`, `lib/codex-legacy-sync`, `lib/memory-vault`, `lib/state-store`, `scripts/ecc-universal-bin`, `scripts/memory-mcp`, `scripts/setup`), none of which read `docs/` or `agents/`. They are pre-existing on this machine, so the "tests pass locally" box stays unticked rather than tick it on a suite that is not green — but this PR adds nothing to that number.

**Review follow-up (second commit).** A locale file with no counterpart in `agents/` was skipped, so every case here silently passed over it — retiring an agent would leave its seven translations behind with nothing to compare against and nothing to report it. Those are now collected and asserted empty. Checked before adding it: zero orphans today across 199 locale docs, so this changes no current result; planting one takes the file to `Passed: 6, Failed: 1` and exit 1.

## Type of Change
- [x] `fix:` Bug fix
- [ ] `feat:` New feature
- [x] `docs:` Documentation
- [ ] `test:` Tests
- [ ] `ci:` CI/CD changes

- [x] JSON files validate cleanly (none changed)
- [x] Shell scripts pass shellcheck (if applicable) — none changed

Not applicable — no new component, and no canonical agent is added or changed. `agents/*.md` is the source of truth here and is untouched; only its translations move toward it, plus one test file under `tests/ci/`.

## Documentation
- [x] Updated relevant documentation — that *is* the change.
- [x] Added comments for complex logic — the test's header records what drifted and why prose and list style are deliberately not compared.
- [ ] README updated (if needed) — not needed

## Note for whoever picks up #2861

This does not fix #2861 and does not touch `manifests/` — that one still needs the maintainer's call between options A, B, and C. The two overlap only in that I found this while checking the reporter's closing observation.


chenhz01:
This fixes the right class of bug — 39 model overstatements all pointing at stale-from-birth translations is a convincing pattern. One hardening question for the new validator: if it pins only `model` and `tools`, other locale-invariant frontmatter keys can drift the same way and stay invisible. Worth pinning the full invariant set in one loop:

```js
const INVARIANT_KEYS = ['name', 'model', 'tools'];
assert.deepStrictEqual(localeFm[key], canonicalFm[key], `${localeFile}: ${key}`);
```

so the next key that drifts fails CI instead of waiting for a third report like #2861. Also consider failing on locale files whose `name` has no canonical counterpart (deleted or renamed agents leave orphan docs). Happy to open a PR with the extended validator if useful.

PR #2897, opened 2026-08-28
fix(hooks): block-no-verify no longer false-positives on quoted text

Davemaina1:
## Summary

`scripts/hooks/block-no-verify.js` scans the raw command string for a git
invocation, but doesn't track shell quoting from the start of the command —
so it can't tell a real `git commit --no-verify` apart from a command that
merely *mentions* the phrase as text inside a different command's argument
(an `echo` message, a heredoc, a logged/test fixture).

Reproduced live against a real PreToolUse hook invocation:

```
$ echo "note: never run git commit --no-verify in this repo"
BLOCKED: --no-verify flag is not allowed with git commit. Git hooks must not be bypassed.
```

That's a harmless echo of documentation text, not a git invocation.

## Fix

Tracks top-level quote spans globally (`computeQuoteSpans`) and only treats
a `git ...` match found inside a quote as a real invocation when that quote
is the argument to something that actually re-executes strings — `bash -c`,
`sh -lc`, `eval`, etc. (`isExecutedSpan`/`computeIgnoredSpans`). So a real
bypass hidden behind those still blocks, while quoted mentions elsewhere
(echo, commit messages, JSON fixtures) don't.

## Type
- [x] Hook

## Testing

- Existing 29-case suite in `tests/hooks/block-no-verify.test.js`: all pass, no regressions (covers the chained-command, quoted-message, and case-insensitivity edge cases this fix had to stay compatible with).
- Added 7 new cases: the false-positive itself (plain + single-quoted echo, an actual echo of JSON-looking text), plus confirming real bypasses hidden behind `bash -c` / an absolute-path `bash -c` / `sh -lc` / `eval` still block.
- Re-verified RED (reconstructed pre-fix code, 2 of the new cases fail) -> GREEN (fixed code, all 36 pass) rather than only adding tests alongside the fix.
- `node scripts/ci/validate-no-personal-paths.js` and `npx eslint` on both changed files: clean.
- Full `npm test`: 3998/4013 passing — same 15 pre-existing failures reproduce identically with this fix stashed out (verified before submitting), so none are caused by this change.

## Checklist
- [x] Tested with Claude Code
- [x] Clear descriptions

### DIFF
--- a/scripts/hooks/block-no-verify.js
+++ b/scripts/hooks/block-no-verify.js
@@ -285,13 +285,117 @@ function isInComment(input, idx) {
 }
 
 /**
- * Find the next 'git' token in the input starting from a position.
+ * Shell built-ins/interpreters that re-execute a quoted string as code
+ * (`bash -c '...'`, `eval "..."`, etc.). A `git ...` phrase found inside a
+ * quoted span whose preceding word is NOT one of these is just string data —
+ * e.g. an `echo` message, a heredoc, a JSON test fixture — not a real
+ * invocation, and must not be treated as one.
  */
-function findGit(input, start) {
+const QUOTE_EXEC_COMMANDS = new Set(['eval', 'bash', 'sh', 'zsh', 'ksh', 'dash', 'source', '.', 'exec']);
+
+/**
+ * Find every top-level (non-nested — shell quotes don't nest) quoted span in
+ * the input, tracking quote state from index 0 so a quote opened at the very
+ * start of the command (e.g. `echo '...'`) is honored for matches found deep
+ * inside it.
+ */
+function computeQuoteSpans(input) {
+  const spans = [];
+  let quote = null;
+  let spanStart = null;
+  let escaped = false;
+
+  for (let i = 0; i < input.length; i++) {
+    const char = input.charAt(i);
+
+    if (escaped) {
+      escaped = false;
+      continue;
+    }
+
+    if (quote) {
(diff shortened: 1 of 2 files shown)


chenhz01:
One correctness issue in the executor set: QUOTE_EXEC_COMMANDS includes env, exec, source, and ., but none of them re-parse a quoted string as shell code. env 'git commit --no-verify' execs a program literally named by the whole string; source 'git commit --no-verify' treats it as a filename; exec passes its args straight to execvp. The shell never runs git in these forms, yet the new logic blocks them — reintroducing the false-positive class this PR set out to remove. The set that genuinely re-executes quoted strings is eval plus bash/sh/zsh/ksh/dash -c. Suggest narrowing accordingly and adding env/source allow-cases to the suite. Available as a paid engagement too, if a delivered PR beats a review comment for you.

Davemaina1:
Thanks for the sharp catch — verified independently and you're right on all counts: `env`/`exec` pass the string straight to execvp as a literal program name, and `source`/`.` treat it as a filename; none of them re-parse it as shell code. Only `env -S`/`--split-string` actually does.

Fixed in 0e62a7a1: dropped `exec`/`source`/`.` from `QUOTE_EXEC_COMMANDS` entirely, and gated `env` on `-S`/`--split-string` specifically so the real bypass stays caught. Added regression tests for all four forms plus the `--split-string` variant. Appreciate the review.

PR #2914, opened 2026-08-31
fix(skills): reduce plan-orchestrate to a registry-validated command router

### LINKED ISSUE #2862: plan-orchestrate ships but emits /orchestrate invocations that no longer resolve
## Summary

`skills/plan-orchestrate` ships in the default install, and its entire output contract is emitting `/ecc:orchestrate` (or `/orchestrate`) invocations for the user to paste. That command is no longer in the default command surface, so the skill's output does not resolve.

## Evidence

The skill ships:

```
package.json:404                 "skills/plan-orchestrate/",
manifests/install-modules.json:793   "skills/plan-orchestrate"
```

Its purpose is generating those invocations. `skills/plan-orchestrate/SKILL.md:10`:

> Bridge a plan document to `/orchestrate custom` by emitting one ready-to-paste invocation per step. The skill is generative only — it never executes `/orchestrate`. The user pastes each line when ready.

`:33-36` defines the output shape as authoritative:

> ## Authoritative `/orchestrate` shape (do not deviate)
> ```
> {ORCH_CMD} custom "<agent1>,<agent2>,...,<agentN>" "<task description>"
> ```

`:52-55` resolves `{ORCH_CMD}` to `/ecc:orchestrate` under a plugin install and `/orchestrate` under a legacy bare install. The worked example at `:244` emits:

```bash
/ecc:orchestrate custom "ecc:tdd-guide,ecc:database-reviewer,ecc:python-reviewer,ecc:security-reviewer" "[Plan: ...] Implement EncryptedString ..."
```

But the command is not in the shipped surface:

- There is no `commands/orchestrate.md`.
- The only remaining definition is `legacy-command-shims/commands/orchestrate.md`, and `legacy-command-shims/README.md` states: *"These slash-entry shims are no longer loaded by the default plugin command surface."*
- Grepping `.claude-plugin/`, `manifests/`, and `package.json` for a shipped `orchestrate` command entry returns nothing (excluding `plan-orchestrate` itself).

Retired in 06f9eca8 (`fix: retire legacy command shims from default surface`). That retirement looks intentional; updating this consumer appears to have been missed.

## Impact

A user installs ECC, runs `plan-orchestrate` on a plan document as designed, receives a set of per-step bash blocks, pastes the first one, and it does not resolve. The failure surfaces at the end of the workflow, after the skill has done its decomposition and chain-composition work, so the wasted effort is maximal.

The stale translated copies still document the old interface (`docs/es|ja-JP|zh-CN|zh-TW|ko-KR|tr|pt-BR/commands/orchestrate.md` describe `/orchestrate [workflow-type] [task]`), which may make this look supported to anyone reading localized docs. There is no English `commands/orchestrate.md`.

## Options

**A. Retarget the emitted form.** Keep the skill's decomposition, tagging, and chain-composition logic (which is genuinely useful and has no equivalent elsewhere) and change what it emits — for example a `Task`-tool dispatch per step, or an `orch-*` skill invocation. Requires rewriting the "Authoritative shape" section and both worked examples.

**B. Restore `/orchestrate` to the default surface.** Reverses part of 06f9eca8; presumably unwanted, since the retirement was deliberate.

**C. Unship the skill.** Remove from `package.json` and `manifests/install-modules.json`, or move it alongside the shims. Loses the decomposition and chain-composition tables, which are the most reusable "plan → per-step agent chain" asset in the repo.

A looks best from the outside, but the right call depends on where orchestration is heading — `workflows/README.md` suggests the fan-out-heavy segments are moving to native Workflow scripts, which may be the intended emit target.

## Notes

Not opening a PR since A, B, and C are materially different product decisions and A's correct emit target depends on the roadmap. Happy to implement whichever you prefer.

Found while auditing the orchestration surface; related to #2861 (unrelated root cause).

santhiprakash:
Fixes #2862

**Revision note (round 2, for reviewers who read the earlier head):** per @haelyra's review, this PR now takes the first, smaller option — a truthful command router. The agent-chain machinery is removed entirely, not just reworded: the skill no longer composes or advertises an "Agent chain", because the emitted command does not consume one. Steps whose command cannot carry plan scope now fail closed instead of being emitted as bare commands.

## What Changed

`skills/plan-orchestrate/SKILL.md` no longer emits the retired `/orchestrate` / `/ecc:orchestrate` commands (retired in `06f9eca8`) and is reduced to a single truthful role: a **command router**.

- **Scope-carrying only.** The skill emits a command only when that command's own file documents a free-form task-description argument that can carry the step marker `[Plan: <path>#step-<id>]`: `/orch-add-feature`, `/orch-fix-defect`, `/orch-change-feature`, `/orch-refine-code`, `/plan`.
- **Fail closed otherwise.** Steps mapping to commands that cannot carry plan scope — `/build-fix` and `/test-coverage` (no argument form), `/update-docs` (no argument form), `/code-review` (`[pr-number | pr-url | blank]`), `/security-scan` (`[path]` + flags), `/loop-start` (`[pattern]` + `--mode`) — are reported `BLOCKED — /<command> cannot carry plan scope; run manually if wanted: <exact documented form>`, listed outside the Batch block, never emitted as unscoped batch entries. A multi-step plan can no longer produce repeated bare `/update-docs` commands with no link to the intended step.
- **Registry-derived argument forms.** The quoted argument forms are a validated snapshot of the command files (frontmatter `argument-hint` / `## Usage` sections; `docs/COMMAND-REGISTRY.json` is the generated index), not an independent source of truth — if a command file and the skill disagree, the command file wins.
- **Chain claims removed.** The agent-chain catalogue, composition rules, `Agent chain` output field, and `--lang` chain-resolution flag are gone; the intro states the skill does not influence what a command runs internally.
- Tag precedence, the feature-verb override, decomposition, and task-description compression are preserved (minus chain references).

## Why This Change

Earlier revisions of this PR kept an advisory `Agent chain` field, but `/plan`, `/update-docs`, `/code-review`, and the `/orch-*` commands each own their own execution behavior — the emitted command does not consume a declared chain, so the claim was false. Separately, steps mapped to argument-less commands lost their step payload entirely (repeated bare `/test-coverage` / `/update-docs` in a batch, with no machine-readable link to the intended step). The router now emits only what is truthful and fails closed when a command cannot carry the step scope.

## Testing Done

- [x] New `tests/skills/plan-orchestrate-registry.test.js` validates every command the skill names against `commands/<name>.md` and `docs/COMMAND-REGISTRY.json`: existence, no retired/legacy-shim references outside the self-check prohibition, scope-carrying claims match each command file's documented argument form, fail-closed commands still document no free-form argument, fail-closed rule stated, chain machinery absent, emitted examples carry the `[Plan:` marker. Sabotage-checked: reintroducing a chain table, claiming scope for `/test-coverage` or `/checkpoint`, and referencing `/orchestrate` each fail the suite.
- [x] `node tests/skills/plan-orchestrate-registry.test.js` — 7 fixture groups pass.
- [x] `npm test` — 4152/4152 tests passed (includes `validate-skills`, `validate-commands`, `catalog:check`, `command-registry:check`).
- [x] `npm run lint` (eslint + markdownlint) — clean.

## Type of Change
- [x] `fix:` Bug fix

### DIFF
--- a/skills/plan-orchestrate/SKILL.md
+++ b/skills/plan-orchestrate/SKILL.md
@@ -33,7 +33,7 @@ Skip when:
 
 ## Authoritative output shape (do not deviate)
 
-The output is one ready-to-paste command per step. The exact form depends on the command, because not every ECC command accepts a quoted task description.
+The output is one ready-to-paste command per step. The exact form is command-specific, because not every ECC command accepts a quoted task description. The `/<command> "<task description>"` placeholder is only valid for the commands listed in this section; all other commands use the bare or documented flag form from the 'Commands that run self-contained or take specific flags' table.
 
 ### Commands that accept a quoted task description
 
@@ -72,7 +72,7 @@ Classify each step by its primary tag and emit the matching command. The command
 
 | Tag | Trigger words | Command | Output form | Why |
 |---|---|---|---|---|
-| `build` | build, compile, lint failure, CI | `/build-fix` | `/build-fix` | Fix build/type/lint errors. |
+| `build` | build, compile, lint, CI, or build-failure context | `/build-fix` | `/build-fix` | Fix build/type/lint/CI errors. Use the `build used as a feature verb` special-case override for feature-creation phrasing such as 'Build a new authentication API' or 'Build the user profile page'. |
 | `fix` | fix, bug, broken, defect, repair, regression | `/orch-fix-defect` | `/<command> "<task description>"` | Existing behavior is wrong. |
 | `test` | test, coverage, e2e, integration | `/test-coverage` | `/test-coverage` | Add or analyze tests. |
 | `db` | schema, migration, index, SQL, Postgres, alembic, sqlmodel | `/orch-add-feature` | `/<command> "<task description>"` | New schema/migration is a net-new capability in the current codebase. |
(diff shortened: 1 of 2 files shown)


chenhz01:
Round 2 looks like the right scope, and the registry test is the strongest part of this PR. One gap in that test: it validates `SKILL.md` against `docs/COMMAND-REGISTRY.json`, but nothing asserts the registry itself is fresh. That JSON is a generated artifact — the realistic failure mode six months out is a contributor adding a command, forgetting to regenerate, and the router test then passing against a stale surface (fail-closed sets computed from outdated argument-hints). Cheap fix inside the same test file: build the expected registry in memory from `commands/` frontmatter and assert it equals the committed JSON, so the failure reads "regenerate COMMAND-REGISTRY" instead of a confusing router mismatch. Also worth one sentence in SKILL.md on what happens when a target command is retired again (the `06f9eca8` scenario): fail-closed with the BLOCKED marker, same as no-argument commands. Happy to add the freshness assertion.

PR #2951, opened 2026-09-04
fix(hooks): support Cursor ApplyPatch payloads

robertdoherty:
## Summary

- recognize Cursor's single-file `ApplyPatch` envelope at the hook wrapper boundary
- translate it to the existing Write/Edit JSON shape so warning and blocking policies receive the target path and patch content
- suppress only the synthetic no-op echo that Cursor otherwise rejects as invalid hook JSON
- preserve Claude Code JSON pass-through, malformed-input compatibility, hook denials, and additional context

## Type

- [x] Hook

## Testing

- `node tests/hooks/run-with-flags.test.js` — 9 passed
- `node tests/hooks/config-protection.test.js` — 9 passed
- `node tests/hooks/gateguard-fact-force.test.js` — 174 passed
- `npm test` — 4,168 passed, 0 failed
- manually reproduced the original two-hook Cursor failure against the active 1.8.0 cache, then verified empty no-op stdout and target-path visibility after the fix

## Checklist

- [x] Follows existing hook wrapper patterns
- [x] Tested with Claude Code-compatible JSON and Cursor ApplyPatch input
- [x] Blocking policies remain effective for protected targets
- [x] No sensitive information included

### DIFF
--- a/scripts/hooks/run-with-flags.js
+++ b/scripts/hooks/run-with-flags.js
@@ -150,9 +150,42 @@ function buildDryRunPreview(hookId, relScriptPath, profilesCsv, raw) {
   return parts.join(' ') + '\n';
 }
 
+function parseApplyPatchPayload(value) {
+  if (typeof value !== 'string') return null;
+
+  const lines = value.replace(/\r\n/g, '\n').split('\n');
+  if (lines.at(-1) === '') lines.pop();
+  if (lines[0] !== '*** Begin Patch' || lines.at(-1) !== '*** End Patch') {
+    return null;
+  }
+
+  const operation = lines[1]?.match(/^\*\*\* (Add|Update) File: (.+)$/);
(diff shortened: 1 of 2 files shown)


chenhz01:
The single-file guard has a gap for mixed envelopes: the middle-line rejection regex only catches `Begin Patch`, `End Patch`, and `Add|Update File:`. A `*** Delete File:` (or `*** Move to:`) line inside an otherwise-valid Update envelope passes the guard, so hooks are shown only the first file's path while the patch deletes or renames a second file — a policy hook scoping writes to an allowlist never sees the affected path. Fail closed like the multi-file case:

```js
/^\*\*\* (?:Begin Patch|End Patch|(?:Add|Update|Delete) File: |Move to: )/
```

(or map Delete/Move to structured ops so blocking hooks can judge them). The rest of the translation — empty no-op suppression while keeping denials audible — reads clean. Happy to open a PR with the tightened guard and a mixed-envelope test if useful.

PR #2954, opened 2026-09-04
fix(gan): require Playwright capability in harness

### LINKED ISSUE #2674: gan-harness.sh: extract_score always returns 0.0 (variable-length PCRE lookbehind, all platforms) + bash 4.3 array syntax on macOS + evaluator allowedTools omits Playwright MCP
## Summary

`scripts/gan-harness.sh` has three defects. The first makes the harness **unable to ever pass**, on any platform — it is not a macOS-only problem as it first appears. The result is that every run burns the full `GAN_MAX_ITERATIONS` (default 15) of Generator/Evaluator agents at the configured model (default `opus`) and then reports a final score of `0.0`.

Note `/gan-build` and `/gan-design` are unaffected — `commands/gan-build.md` orchestrates the same three agents via the Task tool and lets the model read the score, so it never touches this script. Only the standalone shell path is broken.

## Environment

| | |
|---|---|
| ECC | 2.1.0 (`ff15079`) |
| OS | macOS 26.5 (Darwin 25.5.0, arm64) |
| `/bin/bash` | GNU bash 3.2.57 (the macOS system bash; script shebang is `#!/bin/bash`) |
| `grep` | BSD grep 2.6.0-FreeBSD (GNU compatible); `-P` delegates to ugrep |

---

## Defect 1 — `extract_score()` always returns `0.0` (invalid PCRE, all platforms)

`scripts/gan-harness.sh:61-70`

```sh
extract_score() {
  local file="$1"
  grep -oP '(?<=\*\*TOTAL\*\*.*\*\*)[0-9]+\.[0-9]+' "$file" 2>/dev/null \
    || grep -oP '(?<=TOTAL.*\|.*\| \*\*)[0-9]+\.[0-9]+' "$file" 2>/dev/null \
    || grep -oP 'Verdict:.*([0-9]+\.[0-9]+)' "$file" 2>/dev/null | grep -oP '[0-9]+\.[0-9]+' \
    || echo "0.0"
}
```

The first two patterns use **variable-length lookbehind** — `(?<= ... .* ... )`. PCRE does not support this; a lookbehind must be fixed-length. So these fail regardless of which PCRE-backed grep is used, including GNU `grep -P` on Linux.

Reproduction on macOS:

```
$ echo "**TOTAL** | **7.5**" | grep -oP '(?<=\*\*TOTAL\*\*.*\*\*)[0-9]+\.[0-9]+'
ugrep: error: error at position 4
(?m)(?<=\*\*TOTAL\*\*.*\*\*)[0-9]+\.[0-9]+
    \___length of lookbehind assertion is not limited
$ echo $?
2
```

`2>/dev/null` hides the error, the `||` chain falls through, and the function returns `"0.0"`.

Consequence in the loop:

```sh
score_passes() { awk -v s="$score" -v t="$threshold" 'BEGIN { exit !(s >= t) }'; }
```

`0.0 >= 7.0` is never true, so the loop never breaks early. **Every run executes all `GAN_MAX_ITERATIONS` iterations** (default 15) of generator + evaluator at `GAN_*_MODEL` (default `opus`) and then reports failure, no matter how good the output actually was. This is a silent, expensive failure: no error is printed, only a wrong number.

On macOS specifically there is a second, independent problem with the same lines: BSD grep has no `-P` at all unless a ugrep/pcregrep shim is installed.

**Suggested fix** — avoid PCRE entirely; `awk`/`sed` need no lookbehind and are portable:

```sh
extract_score() {
  local file="$1" s
  s=$(awk '
    /\*\*TOTAL\*\*/ || /Verdict:/ {
      if (match($0, /[0-9]+\.[0-9]+/)) { print substr($0, RSTART, RLENGTH); exit }
    }' "$file" 2>/dev/null)
  if [ -n "$s" ]; then printf '%s\n' "$s"; else return 1; fi
}
```

Sketch only — the substantive points are (a) no variable-length lookbehind, and (b) it would be worth **failing loudly** when the score cannot be parsed rather than silently substituting `0.0`, since the current default is indistinguishable from a genuinely terrible score.

---

## Defect 2 — `${SCORES[-1]}` requires bash 4.3+; macOS ships bash 3.2

`scripts/gan-harness.sh:244`

```sh
FINAL_SCORE="${SCORES[-1]:-0.0}"
```

Negative array subscripts were added in bash 4.3. The script's shebang is `#!/bin/bash`, which on macOS is **bash 3.2.57** (Apple has not shipped bash 4+ since the GPLv3 licence change).

Reproduction:

```
$ /bin/bash -c 'SCORES=(6.5 7.2); echo "${SCORES[-1]:-0.0}"'
/bin/bash: SCORES: bad array subscript
0.0
```

The `:-0.0` default masks the error, so the Build Report prints `0.0` as the final score even when the array holds real values. Combined with Defect 1 this is doubly wrong, but it is independent — it would still corrupt the report after Defect 1 is fixed.

**Suggested fix** — bash 3.2-compatible:

```sh
FINAL_SCORE="${SCORES[$(( ${#SCORES[@]} - 1 ))]:-0.0}"
```

Or switch to `#!/usr/bin/env bash` plus an explicit version guard, so the failure is loud rather than silent.

---

## Defect 3 — evaluator `--allowedTools` omits MCP, so `playwright` eval mode degrades silently

`scripts/gan-harness.sh:186-187`

```sh
claude -p --model "$EVALUATOR_MODEL" \
  --allowedTools "Read,Write,Bash,Grep,Glob" \
  "You are the Evaluator in a GAN-style harness. ...
Eval mode: $EVAL_MODE
Dev server: http://localhost:$DEV_PORT
```

`GAN_EVAL_MODE` defaults to `playwright`, and SKILL.md describes the evaluator as using "**Playwright MCP** to interact with the live application" — that live-app testing is the harness's core value proposition. But no `mcp__playwright__*` tool is allowed, so the evaluator cannot drive the browser. The prompt still states `Eval mode: playwright`, so the agent believes it is in that mode while silently falling back to reading code.

This yields scores that look like live-app evaluations but are not, which specifically undermines the Craft and Functionality criteria.

**Suggested fix** — include the MCP tools when the mode requires them:

```sh
EVAL_TOOLS="Read,Write,Bash,Grep,Glob"
[ "$EVAL_MODE" = "playwright" ] && EVAL_TOOLS="$EVAL_TOOLS,mcp__playwright__*"
```

...and ideally warn or refuse when `playwright` mode is requested but no Playwright MCP server is configured, instead of degrading quietly.

Related: the shipped agent definitions `agents/gan-{planner,generator,evaluator}.md` also declare `tools: Read, Write, Bash, Grep, Glob` with no MCP entry, so the Task-tool path in `commands/gan-build.md` has the same gap. Possibly worth a separate issue.

---

## Impact summary

| Defect | Platform | Symptom | Cost |
|---|---|---|---|
| 1 | **all** (PCRE constraint) | score always `0.0`, threshold never met | full `MAX_ITERATIONS` at opus, every run |
| 2 | macOS (bash 3.2) | final report shows `0.0` | wrong report |
| 3 | all | `playwright` mode silently becomes code-only | invalid scores |

Defects 1 and 2 both fail **silently** by substituting `0.0`, which is why a run can complete looking merely disappointing rather than broken.

## Reproduction

1. macOS, `/bin/bash` 3.2, ECC 2.1.0.
2. `./scripts/gan-harness.sh "Build a small dashboard"` with any brief.
3. Observe: every iteration runs, no early break, Build Report shows `0.0` regardless of the evaluator's actual written score in `gan-harness/feedback/feedback-*.md`.

Defects 1 and 2 also reproduce standalone with the two one-liners quoted above.

wellkilo:
## Summary

- fail closed before setup when `GAN_EVAL_MODE=playwright` does not have a connected MCP server named `playwright`
- recheck that capability immediately before each evaluator launch
- grant the standalone evaluator the same six Playwright MCP tools already approved for `gan-evaluator`, while keeping screenshot and code-only modes on the base tool set
- reject unknown evaluation modes before filesystem or Git side effects

This addresses the remaining standalone Playwright capability gap in #2674. The portable score parser and Bash 3.2 final-score fix are already on `main`.

## Testing

- `node tests/gan-harness.test.js` (10 passed, 0 failed)
- `node tests/ci/gan-evaluator-tools.test.js`
- `node scripts/ci/validate-agents.js`
- `node scripts/ci/check-unicode-safety.js`
- `npx --no-install eslint tests/gan-harness.test.js`
- `bash -n scripts/gan-harness.sh`
- `git diff --check`

The repository-wide suite reached 4,155 passing checks. Its 17 remaining failures were unrelated and reproduced from `origin/main`: 15 validator fixture collisions during the parallel/full run, one sandbox-denied write to `~/.claude/package-manager.json`, and one flaky Plan Canvas SSE timeout. The validator group passes 176/176 when rerun in isolation with dependencies available.

Closes #2674

### DIFF
--- a/scripts/gan-harness.sh
+++ b/scripts/gan-harness.sh
@@ -19,6 +19,7 @@
 #   GAN_PROJECT_DIR     — Working directory (default: current dir)
 #   GAN_SKIP_PLANNER    — Set to "true" to skip planner phase
 #   GAN_EVAL_MODE       — playwright, screenshot, or code-only (default: playwright)
+#                        playwright requires a connected MCP server named "playwright"
 
 set -euo pipefail
 
@@ -96,6 +97,38 @@ score_passes() {
   awk -v s="$score" -v t="$threshold" 'BEGIN { exit !(s >= t) }'
 }
 
+playwright_mcp_is_connected() {
+  local status
+  local status_value
+  local check_mark
+  local heavy_check_mark
+  status=$(NO_COLOR=1 claude mcp get playwright 2>/dev/null) || return 1
+  status_value=$(printf '%s\n' "$status" | awk '
+    /^[[:space:]]*Status:[[:space:]]*/ {
+      sub(/^[[:space:]]*Status:[[:space:]]*/, "")
+      sub(/[[:space:]]*$/, "")
+      print
+      exit
+    }
+  ')
+  check_mark=$(printf '\342\234\223')
+  heavy_check_mark=$(printf '\342\234\224')
+
+  [ "$status_value" = "$check_mark Connected" ] || \
+    [ "$status_value" = "$heavy_check_mark Connected" ]
+}
+
+evaluator_tools_for_mode() {
+  local base_tools="Read,Write,Bash,Grep,Glob"
+  local playwright_tools="mcp__playwright__browser_navigate,mcp__playwright__browser_click,mcp__playwright__browser_take_screenshot,mcp__playwright__browser_snapshot,mcp__playwright__browser_type,mcp__playwright__browser_fill_form"
+
+  if [ "$1" = "playwright" ]; then
+    printf '%s,%s\n' "$base_tools" "$playwright_tools"
+  else
+    printf '%s\n' "$base_tools"
+  fi
+}
+
 elapsed() {
   local now=$(date +%s)
   local diff=$((now - START_TIME))
@@ -103,6 +136,24 @@ elapsed() {
 }
 
 # ─── Setup ───────────────────────────────────────────────────────────────────
+
+case "$EVAL_MODE" in
+  playwright)
+    if ! playwright_mcp_is_connected; then
+      fail "GAN_EVAL_MODE=playwright requires a connected MCP server named 'playwright'."
+      fail "Run 'claude mcp get playwright' to inspect its status, or choose GAN_EVAL_MODE=screenshot or code-only."
+      exit 1
+    fi
+    ;;
+  screenshot|code-only)
+    ;;
+  *)
+    fail "Unsupported GAN_EVAL_MODE. Expected playwright, screenshot, or code-only."
+    exit 1
+    ;;
+esac
+
+EVALUATOR_TOOLS=$(evaluator_tools_for_mode "$EVAL_MODE")
 
 phase "GAN-STYLE HARNESS — Setup"
 
@@ -205,8 +256,14 @@ Update gan-harness/generator-state.md." \
   # ── EVALUATE ──
   echo -e "${RED}>> EVALUATOR (iteration $i)${NC}"
 
+  if [ "$EVAL_MODE" = "playwright" ] && ! playwright_mcp_is_connected; then
+    fail "The Playwright MCP server disconnected before evaluator iteration $i."
+    fail "Run 'claude mcp get playwright' to inspect its status, then retry the harness."
+    exit 1
+  fi
+
   claude -p --model "$EVALUATOR_MODEL" \
-    --allowedTools "Read,Write,Bash,Grep,Glob" \
+    --allowedTools "$EVALUATOR_TOOLS" \
     "You are the Evaluator in a GAN-style harness. Read agents/gan-evaluator.md for full instructions.
 
 Iteration: $i
--- a/tests/gan-harness.test.js
+++ b/tests/gan-harness.test.js
@@ -12,7 +12,9 @@ const { spawnSync } = require('child_process');
 
 const repoRoot = path.resolve(__dirname, '..');
 const harnessPath = path.join(repoRoot, 'scripts', 'gan-harness.sh');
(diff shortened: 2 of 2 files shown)


chenhz01:
Fail-closed gating is the right call, and rechecking before each evaluator launch covers the disconnect-mid-run case. One fragility in `playwright_mcp_is_connected`: it parses `claude mcp get playwright`'s human-readable output and matches the check glyph byte-for-byte (`\342\234\223 Connected` / `\342\234\224 Connected`). That couples the gate to CLI presentation — a version that reformats status, handles `NO_COLOR` differently, or localizes output produces false negatives (usable setups rejected) or, worse, misses a disconnect. Prefer structured output when the CLI version supports it, keeping text parsing as fallback:

```sh
claude mcp get playwright --json 2>/dev/null | jq -e '.status == "connected"' || fall_back_to_text_parse
```

Happy to open a PR with that if useful.

PR #2960, opened 2026-09-04
docs: correct GitHub Copilot support status for Copilot CLI

swigerb:
## What Changed

Corrects the GitHub Copilot entries in `README.md`. Documentation only.

Three claims are no longer accurate for GitHub Copilot CLI:

| Current README | Actual behavior (Copilot CLI 1.0.83) |
|---|---|
| "No ECC hooks, runtime agents, delegation, or **native skill discovery**" | Copilot CLI discovers `SKILL.md` from `.github/skills/`, `.agents/skills/`, `.claude/skills/`, `~/.copilot/skills/`, `~/.agents/skills/` |
| Skills: "Prompt/instruction references only" | ECC's 286 skills load unmodified — frontmatter is already compatible |
| Agents: "Not supported (Copilot has no subagent API)" | `.github/agents/*.md` + `--agent <name>` runs ECC's agent files as-is |

Adds a short "Copilot CLI: skills and agents" subsection with reproducible commands, and rewrites Limitations to state what genuinely does not work.

## Why This Change

ECC already ships a skill subset in `.agents/skills/` for Codex. Copilot CLI independently treats `.agents/skills/` as a discovery path — so **a plain clone of this repo already loads 42 ECC skills into Copilot CLI with zero configuration.** That works today and is undocumented. Readers following the current matrix will conclude it's impossible and won't try.

This is deliberately documentation-only. #2524 was closed for adding an unsafe execution surface; this PR adds none. It corrects statements of fact and documents behavior that already exists.

## Testing Done

Verified against GitHub Copilot CLI 1.0.83 on a clean clone of `main`.

**1 — Plain clone, no configuration:**

```console
$ git clone https://github.com/affaan-m/ECC.git && cd ECC
$ copilot skill list --json | jq '[.[]|select(.source=="project")]|length'
42        # 39 from .agents/skills/, 3 from .claude/commands/
```

**2 — Full catalog:**

```console
$ cp -r skills .github/skills
$ copilot skill list --json | jq '[.[]|select(.source=="project")]|length'
286
$ copilot skill list --json | jq '[.[]|select(.enabled==false)]|length'
0
```

**3 — Agents:**

```console
$ cp -r agents .github/agents
$ copilot --agent architect -p "Reply with exactly: LOADED"
Warning: Custom agent "architect" specifies model "opus" which is not available; using "gpt-5.6-sol" instead
LOADED
```

That warning is the only diagnostic emitted, and the new section documents it.

[...]

PR #2963, opened 2026-09-04
feat(install): add a copilot project target for skills and agents

swigerb:
## Summary

Adds a `copilot` install target that places ECC's skill and agent catalog where GitHub Copilot CLI discovers it: skills to `.github/skills/`, agents to `.github/agents/`.

Stacked on #2960 — that PR corrects the README's Copilot support claims, this one makes the placement automatic. The docs commits appear here until #2960 merges; the reviewable change is the last commit.

## Scope

Deliberately limited to agents and skills:

- **Hooks** are written against Claude Code's event model and are not ported.
- **Commands** rely on slash-command argument substitution Copilot CLI does not provide.
- **Rules** have no Copilot CLI surface.

Nothing is written outside `.github/`.

## Agent frontmatter transform

Agent files go through an allowlist (`name`, `description`) rather than a passthrough, because the remaining Claude keys do not transfer:

| Key | Why it is dropped |
|---|---|
| `model` | Copilot resolves the session model from user configuration and plan entitlement. A Claude model id makes it warn and silently fall back on **every** invocation. |
| `tools` | Claude tool names are not Copilot tool names. Copilot governs tool access per session via `--allow-tool`/`--deny-tool`, so carrying the list asserts a restriction Copilot does not apply. |
| `color` | Claude Code presentation only. |

An allowlist rather than a denylist means keys added to ECC agents later cannot silently leak into the Copilot copy.

The warning this eliminates, observed directly:

```
# agent copied verbatim
Warning: Custom agent "architect" specifies model "opus" which is not available; using "gpt-5.6-sol" instead

# same agent installed through the target
(no output)
```

## Implementation note

`supportsModule` stays permissive and `planOperations` does the narrowing. Filtering in `supportsModule` instead looks tidier but breaks dependency resolution: modules that act purely as dependency anchors (`rules-core`, `commands-core`, `platform-configs`) get skipped, and every module depending on them goes with it. An unsupported path now contributes zero operations rather than removing the module. This matches the convention the other adapters use.

## Verification

Installed into a clean fixture with `--profile full --target copilot`, then exercised against **GitHub Copilot CLI 1.0.83**:

- 285 skills and 68 agents placed; footprint confined to `.github/` (verified by directory listing).
- `copilot skill list --json` → 285 project skills, **285 enabled, 0 disabled, empty stderr**.
- `copilot --agent architect -p ...` → runs, replies, **no warnings**.
- Counterfactual above confirms the transform is doing real work.

The 285 is one fewer than a manual `cp -r skills .github/skills`: `dmux-workflows` belongs to the `orchestration` module, which ships tmux and shell worker scripts and does not target Copilot. Documented in the README.

Test results:

| Check | Result |
|---|---|
| `tests/lib/copilot-install-target.test.js` | 13/13 (new) |
| `tests/lib/harness-capabilities.test.js` | 10/10 |
| `validate-install-manifests` | 36 modules, 83 components, 7 profiles |
| `npm run harness:adapters` | PASS, 12 adapters |
| `npm run catalog:check` | counts match |
| `npm test` (full) | 3985 tests, 3954 passed, 31 failed |

The 31 failures are **pre-existing**. Baseline `npm test` on `main` at `e04ea0b`: 3972 tests, 3941 passed, **31 failed**. This branch adds 13 tests and 13 passes, with no change in the failure count. Two files differed between the two full runs (`lib/install-state-projection`, `integration/plan-canvas-e2e`); running each in isolation on both refs gives identical results (8/0 and 8/1 respectively), so both are pre-existing or flaky under full-run contention rather than caused by this change.

`npm run lint` could not be executed locally — the corporate npm proxy 404s on `ignore@7.0.8` and the public registry is TLS-intercepted. The new files follow the existing adapter style closely (`copilot-agent.js` mirrors `antigravity-agent.js`; the `module` parameter shadowing matches every other adapter), so CI lint should be the authority here.


chenhz01:
The allowlist transform is the right call for Copilot's frontmatter contract, and the planner/ownership fixes in the thread cover the earlier findings. What the design leaves open: installed agents become a projection, not a copy. When the allowlist itself changes (a new Claude key gets dropped, or `description` handling shifts), previously installed `.github/agents/` files stay at the old transform until someone reinstalls. The ownership manifest knows the files; it doesn't know which transform produced them. Suggest stamping it:
```json
{"module":"copilot-agents","transform":"v1","keys":["name","description"]}
```
in the install state, so a future transform change can detect stale projections and prompt a re-sync instead of silently serving outdated agent definitions. Happy to open a PR with working code if useful.

PR #2965, opened 2026-09-05
fix(block-no-verify): skip "git" that sits in data context (false positive on literals + later -n)

groy75:
## Problem

`block-no-verify.js` scans the raw command string for `git <subcmd>` with no notion of whether that text is a command or an argument. A `git commit` literal inside a quoted string of some **other** command — `echo "…"`, a `VAR="…"` assignment, a printf JSON payload, a python heredoc — is picked up as a real invocation. `hasNoVerifyFlag` then re-tokenizes from that inner offset with a fresh quote state; a stray `"` runs the "segment" to end of input, and an unrelated later `bash -n` / `sed -n` / `grep -n` is read as the short no-verify flag.

Minimal repro on `main` (exit 2, should be 0):

```sh
echo '{"tool_input":{"command":"echo \"git commit -n\""}}' | node scripts/hooks/block-no-verify.js
```

Two legitimate commands were blocked this way in one working session (a Python patch script containing the literal, followed by `bash -n`).

## Fix

`isDataContext(input, idx)` walks the input to the candidate `git` tracking quotes, escapes, `$(…)` inside double quotes, and heredoc bodies. A `git` is **data** — and skipped — when it sits inside a quoted argument of a non-shell command, or inside a heredoc whose consumer is not a shell.

Command contexts that must still be scanned are preserved and tested: `bash/sh/zsh/dash/ksh -c "…"`, `eval "…"`, `ssh host …`, `"$(…)"` / backtick substitutions, and heredocs fed to `bash`/`sh`/`ssh`/`eval`.

## Tests

8 added to `tests/hooks/block-no-verify.test.js`: 5 false-positive shapes that now pass, 3 command-context shapes that must still block. `33/33`; the Cursor variant shares the module, `14/14`.

Verified through the same `run-with-flags.js pre:bash:block-no-verify` path the dispatcher uses.

https://claude.ai/code/session_01EyuZY1LZNuShHiewExekLT

### DIFF
--- a/scripts/hooks/block-no-verify.js
+++ b/scripts/hooks/block-no-verify.js
@@ -162,11 +162,14 @@ function tokenizeShellWords(input, start = 0, end = input.length) {
   return tokens;
 }
 
-function findCommandSegmentEnd(input, start) {
+/**
+ * Find the end of a shell command segment without scanning beyond `limit`.
+ */
+function findCommandSegmentEnd(input, start, limit = input.length) {
   let quote = null;
   let escaped = false;
 
-  for (let i = start; i < input.length; i++) {
+  for (let i = start; i < limit; i++) {
     const char = input.charAt(i);
 
     if (escaped) {
@@ -200,7 +203,7 @@ function findCommandSegmentEnd(input, start) {
     }
   }
 
-  return input.length;
+  return limit;
 }
 
 function commitOptionConsumesNextValue(value) {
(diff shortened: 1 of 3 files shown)


chenhz01:
Extracting `shell-scan.js` makes it shared infrastructure — the header says "shared by hook-bypass matchers", plural — but its only tests are indirect, through `block-no-verify.test.js`. The tricky primitives deserve direct coverage before a second consumer depends on them: `tokenizeShellWords` (quotes, escapes, ANSI-C `$'...'`, backslash-newline continuations), `HEREDOC_START` against `<<-` and quoted delimiters, and `lineContinuationEnd` for CRLF. A `tests/hooks/lib/shell-scan.test.js` pinning those, plus the failure modes that motivated the extraction (stale quote state running a segment to end of input), keeps future refactors of the lib from being masked by the caller's integration suite. Cheap now while the extraction is fresh; expensive after gateguard-style consumers accrete. Want me to take a crack at it? Happy to open a PR.

PR #2983, opened 2026-09-05
fix: session-activity-tracker never records + live dashboard Activity view

c10vis-poem:
## The bug

`session-activity-tracker.js` is correctly registered in `posttooluse-dispatcher.js`'s `SYNC_HOOKS` and fires on every tool call — but `buildActivityRow()` only ever read the session id from `ECC_SESSION_ID`/`CLAUDE_SESSION_ID` environment variables. Claude Code does not set either as an actual environment variable for hook subprocesses, so every row came back `null` and nothing was ever written to `~/.claude/metrics/tool-usage.jsonl` — silently, no error anywhere.

The session id is available in the hook's own stdin JSON payload (`input.session_id`) in real invocations. This fix prefers that, keeping the env vars as a fallback for any caller that does set them.

## The feature

Adds a live **Activity** tab to the capabilities dashboard (`scripts/dashboard-web.js`), reading the same metrics file via a new `/api/activity` endpoint, polling every 3s. `Skill`/`Agent` calls are visually distinct from raw `Bash`/`Edit`/`Write`, so it's now possible to actually *see* whether a session is using the tools ECC ships, in real time, instead of manually tailing and `jq`-ing a JSONL file.

Also documents the dashboard's loopback-only/no-auth design in its own usage comment, with the standard SSH local-port-forwarding pattern for viewing it from another device you control.

## Testing

- Added a new test (`reads session_id from stdin JSON when no session env var is set`) covering the actual fix.
- All 18 tests in `tests/hooks/session-activity-tracker.test.js` pass (17 existing + 1 new).
- `node --check` clean on both modified files.
- The dashboard change was verified live against a real running session before this PR — `/api/activity` returned real entries end-to-end.

## Why this matters beyond one setup

This bug means the tracker has been silently producing empty output for every ECC install, for anyone who has it active — the exact "is my session actually using the tools it has" visibility question has never actually been answerable from this file, regardless of hook profile. The dashboard view turns the fix into something a user can actually look at, not just a JSONL file that now happens to have content.

### DIFF
--- a/scripts/dashboard-web.js
+++ b/scripts/dashboard-web.js
@@ -6,6 +6,15 @@
  * Usage: node scripts/dashboard-web.js [port]
  * Open http://localhost:3456
  *
+ * The server only binds to loopback and has no authentication, by design --
+ * it is meant to be viewed on the same machine it runs on. To check it from
+ * another device you control (a tablet, a laptop), use standard SSH local
+ * port forwarding instead of exposing the port itself:
+ *   ssh -L 3456:127.0.0.1:3456 <user>@<this-machine> -p <ssh-port>
+ * then open http://127.0.0.1:3456 on the *remote* device. This requires
+ * only SSH access you already control on your own machine -- nothing here
+ * opens the dashboard to the network.
+ *
  * Contribution: https://github.com/affaan-m/ECC
  */
 
--- a/scripts/dashboard-web.js
+++ b/scripts/dashboard-web.js
@@ -12,6 +12,7 @@
 const fs = require('fs');
 const path = require('path');
 const http = require('http');
+const os = require('os');
 const {
   LOOPBACK_HOSTNAMES,
   buildAllowedHostnames,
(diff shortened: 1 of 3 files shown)


chenhz01:
The root-cause fix is right: session ids only arrive on hook stdin, so preferring `input.session_id` over env vars is what actually makes the tracker record, and applying the same reasoning to `hook_event_name` closes the sibling bug. One scaling note on the dashboard side: if `/api/activity` re-reads and re-parses the whole `tool-usage.jsonl` on every 3-second poll, response cost grows linearly with file size — a long-lived install with months of rows gets sluggish exactly for the users most likely to be watching it. An incremental tail keeps poll cost constant: persist a byte offset per poll, re-parse fully only when size drops below the offset (rotation/truncation signal). The `sanitizeSessionId` reuse from `session-bridge` matches the sibling hooks — good consistency. Happy to open a PR with the offset-tail endpoint change.

PR #2988, opened 2026-09-06
feat: map FastAPI projects to fastapi-patterns and api-design

jlwainwright:
## What Changed

Adds a `fastapi` stack entry to `config/project-stack-mappings.json`, placed beside the existing `django` entry.

It resolves to skills that already ship in this repo:

- `fastapi-patterns`
- `api-design`
- `python-patterns`, `python-testing`, `tdd-workflow`, `verification-loop`

Indicators follow the `django` pattern — `fastapi` appearing in `requirements.txt` or `pyproject.toml`. Commands cover `uvicorn main:app --reload` and `fastapi dev`, with `uvicorn *` added to the allow list.

## Why This Change

`skills/fastapi-patterns/` and the `fastapi-reviewer` agent both ship today, but nothing routed to them. With no `fastapi` entry, `/project-init` matched FastAPI projects on the generic `python` indicators (`pyproject.toml`, `requirements.txt`) and stopped there — so a FastAPI project got `python-patterns` and never the FastAPI skill written for it.

There is a second effect on install size. Because `coding-standards`/`frontend-patterns`/`backend-patterns` resolve through the coarse `framework-language` module, a generic-Python plan pulls that whole bundle (android, angular, cpp, csharp, dart, dotnet, laravel, perl, quarkus, springboot…). Requesting `fastapi-patterns` and `api-design` instead resolves to their own narrow modules — on the repo I tested this against, 13 modules dropped to 7.

## Testing Done

- [x] Automated tests pass locally (`node tests/run-all.js`) — **4159 passed, 0 failed**

Verified on a real FastAPI + Next.js codebase:

- JSON parses; `stacks` goes 20 → 21
- All 6 referenced skill directories and both rule directories resolve on disk
- Both indicators match (`fastapi` present in `requirements.txt` and `pyproject.toml`)
- `install-plan.js --skills fastapi-patterns,api-design,…` resolves cleanly

## Type of Change

- [x] `feat:` New feature

- [ ] Shell scripts pass shellcheck (n/a — no shell changed)

## Notes for reviewers

Data-only. No code reads `project-stack-mappings.json` — `grep -rln project-stack-mappings` returns only `commands/project-init.md` and its ja-JP translation, so this changes what `/project-init` detects, not installer behaviour. No new skill, command, agent or hook is added, so the registration/catalog checklist does not apply.

Placement next to `django` is by convention only; entries are not evaluated in order.

### DIFF
--- a/config/project-stack-mappings.json
+++ b/config/project-stack-mappings.json
@@ -486,6 +486,34 @@
       }
     },
     {
+      "id": "fastapi",
+      "name": "FastAPI (Python)",
+      "indicators": [
+        { "file": "requirements.txt", "contains": "fastapi" },
+        { "file": "pyproject.toml", "contains": "fastapi" }
+      ],
+      "rules": ["common", "python"],
+      "skills": [
+        "fastapi-patterns",
+        "api-design",
+        "python-patterns",
+        "python-testing",
+        "tdd-workflow",
+        "verification-loop"
+      ],
+      "commands": {
+        "build": ["pip install -e ."],
+        "test": ["pytest", "python -m pytest"],
+        "lint": ["ruff check .", "mypy ."],
+        "format": ["ruff format .", "black ."],
+        "dev": ["uvicorn main:app --reload", "fastapi dev"]
+      },
+      "permissions": {
+        "allow": ["python *", "pip install *", "pytest *", "ruff *", "black *", "mypy *", "uvicorn *"],
+        "deny": []
+      }
+    },
+    {
       "id": "android",
       "name": "Android (Kotlin/Java)",
       "indicators": [


rajpratham1 (changes requested):
The FastAPI mapping currently introduces a missing comma, making config/project-stack-mappings.json invalid and preventing /project-init from loading the mappings. The fastapi dev command also needs its dependency/availability handling and permission allow-list corrected. Please fix these issues and re-run the relevant validation checks before merging.

rajpratham1 (changes requested):
The FastAPI mapping currently introduces invalid JSON due to the missing comma before the new object, which prevents /project-init from loading the mappings. There are also dependency and permission mismatches around uvicorn / fastapi dev. Please fix the JSON syntax, ensure these commands are only exposed when their dependencies are available, and add the required fastapi * permission (or remove the command) before merging.

chenhz01:
One dependency subtlety in the dev commands: `fastapi dev` is provided by `fastapi-cli`, which ships only with `fastapi[standard]` — but the indicators match bare `fastapi` in requirements/pyproject, which a minimal install satisfies. So the mapping can route a project to a dev command that fails with 'command not found' on its first run. Two options that keep the mapping honest: keep `uvicorn main:app --reload` as the only dev default and document `fastapi dev` as the standard-extra alternative, or add an indicator line `{ "file": "requirements.txt", "contains": "fastapi[standard]" }` to gate it. Everything else in the entry — placement beside `django`, the skill set, the `uvicorn *` allow rule — lines up with the existing schema. Happy to open a PR with working code if useful.

PR #2993, opened 2026-09-06
chore(skills): declare license and agent compatibility in skill frontmatter

cagrisolakoglu:
`skillforge validate skills/` reports two warnings on nearly every skill: **SF1009** (no license declared, 268 of 286) and **SF1010** (no agent compatibility declared, 286 of 286). This closes both for the canonical `skills/` tree.

> **Split for reviewability.** This PR was 325 files, over CodeRabbit's 300-file cap, which has no bypass. It is now **289 files** — the `skills/` tree plus three test updates. The `.agents/skills` Codex mirror moved to **#2997** (39 files). The two are independent; neither depends on the other landing first.

## Changes

**`license: MIT`** — added only where no license was declared. The 18 skills that already declare one (8 `Apache-2.0`, 10 `MIT`) are left exactly as they are; those values read as deliberate, carried over from wherever the material came from, and a blanket rewrite would misstate them.

**`compatibility: [claude-code, codex]`** — added to all 286. ECC ships the `.agents/skills` mirror for Codex, so both providers apply. Validated against each afterwards: no new findings from either.

Frontmatter only — **+593 lines, 0 deletions** in `skills/`, no body text touched.

## Three test updates

Three tests pinned specific skills' frontmatter tightly enough that any added top-level key broke them. `@greptile-apps` found the equivalent problem in the Codex mirror; these three surfaced from running the full suite.

| Test | What it asserted | Change |
|---|---|---|
| `ci/ito-training-skill.test.js` | regex requiring `---` immediately after the `metadata` block | dropped the trailing `---\n`; required keys still asserted in order from the start of the file |
| `ci/tasteforge-video-skill.test.js` | same shape | same |
| `skills/terminal-opener.test.js` | `deepStrictEqual(keys, ['name', 'description'])` | extended to include `license` and `compatibility` — that assertion guards against frontmatter bloat, so it is extended rather than loosened |

## Why compatibility is the part that matters

Without a declared `compatibility`, SkillForge never runs its provider-specific rules, so a whole class of problems stays invisible. The one worth naming is the **1024-character description limit**: past that, Claude Code truncates the description in the skill list, and what gets cut is the tail — exactly where the `Use when …` activation triggers live. The skill quietly stops being selected, with no error anywhere.

No ECC skill exceeds it today. The point is that from now on, if one does, `skillforge validate` will say so.

## Verification

```
skillforge validate skills/                     607 → 53 warnings
node scripts/ci/validate-skills.js              Validated 805 skill directories
node tests/ci/codex-skill-surface.test.js       4 passed, 0 failed
node tests/ci/ito-training-skill.test.js        7 passed, 0 failed
node tests/ci/tasteforge-video-skill.test.js   11 passed, 1 skipped
node tests/skills/terminal-opener.test.js      27 passed, 0 failed
npm run catalog:check                           counts match
```

## Deliberately out of scope

- **6 × SF0008** — `../../rules/react/…`, `../../docs/…`, `../../ECC-Tools` references. These reach outside a skill directory, which is simply how ECC is laid out; same reasoning as #2618.
- **20 × SF1011** (sibling-skill references), **16 × SF1006**, **7 × SF4001**, **3 × SF5001**, **6 × SF1007** (all `rm -rf` sites reviewed — each is a quoted temp-dir cleanup behind a guard or a `trap`).
- The localized trees under `docs/*/skills/` and the `.kiro/` and `.cursor/` mirrors — downstream copies; folding them in would triple the diff without adding a decision to review.

### DIFF
--- a/skills/accessibility/SKILL.md
+++ b/skills/accessibility/SKILL.md
@@ -4,6 +4,8 @@ description: Design, implement, and audit inclusive digital products using WCAG 
   standards. Use this skill to generate semantic ARIA for Web and accessibility traits for Web and Native platforms (iOS/Android).
 metadata:
   origin: ECC
+license: MIT
+compatibility: [claude-code, codex]
 ---
 
 # Accessibility (WCAG 2.2)
--- a/skills/agent-architecture-audit/SKILL.md
+++ b/skills/agent-architecture-audit/SKILL.md
@@ -4,6 +4,8 @@ description: Full-stack diagnostic for agent and LLM applications. Audits the 12
 metadata:
   origin: oh-my-agent-check
 tools: Read, Write, Edit, Bash, Grep, Glob
+license: MIT
+compatibility: [claude-code, codex]
 ---
 
 # Agent Architecture Audit
--- a/skills/agent-eval/SKILL.md
+++ b/skills/agent-eval/SKILL.md
@@ -5,6 +5,7 @@ license: MIT
 metadata:
   origin: ECC
 tools: Read, Write, Edit, Bash, Grep, Glob
+compatibility: [claude-code, codex]
 ---
 
 # Agent Eval Skill
--- a/skills/agent-harness-construction/SKILL.md
+++ b/skills/agent-harness-construction/SKILL.md
@@ -3,6 +3,8 @@ name: agent-harness-construction
 description: Design and optimize AI agent action spaces, tool definitions, and observation formatting for higher completion rates. Use when defining or revising an agent's tool set, action space, or observation format.
 metadata:
   origin: ECC
+license: MIT
+compatibility: [claude-code, codex]
 ---
 
 # Agent Harness Construction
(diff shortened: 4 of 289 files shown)


chenhz01:
Two flags before this lands. First, `compatibility: [claude-code, codex]` goes on all 286, but the `.agents/skills` mirror is a manually-synced subset — that is exactly what #3188's lint enforces — so a skill claiming Codex compat with no mirror copy is uninvocable in Codex, and the field asserts something the tree does not guarantee. Either gate the addition on mirror presence or let the lint fill the gap first. Second, blanket `license: MIT` over skills with third-party `metadata.origin` (e.g. `agent-architecture-audit` carried from oh-my-agent-check): the 18 pre-existing declarations were excluded precisely because they read as deliberate carry-overs — the origin-carryover set deserves the same check of the upstream repo's license before MIT is asserted on its behalf. Happy to script the origin-to-upstream-license audit if useful.

PR #2998, opened 2026-09-06
feat(ecc2): port package manager detection logic to Rust

mawquf28:
## What Changed
Ported the package manager detection and command pattern logic from `scripts/lib/package-manager.js` into a new native Rust crate inside `ecc2/package_manager`.

## Why This Change
Part of the core migration roadmap to Rust. Moving this logic to native binary increases execution speed, enhances cross-platform stability (especially for native Windows users), and completely eliminates Node.js startup overhead during package detection cycles.

## Testing Done
* [x] Manual testing completed
* [ ] Automated tests pass locally (`node tests/run-all.js`)
* [x] Edge cases considered and tested

Tested natively on Fedora Linux via `cargo run` and verified successful detection.

## Type of Change
- [x] `feat`: New feature

- [x] JSON files validate clearly
- [x] Pre-commit hooks pass locally

### DIFF
--- a/ecc2/package_manager/Cargo.toml
+++ b/ecc2/package_manager/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "ecc-package-manager"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+thiserror = "1.0"
+regex = "1.10"
+dirs = "5.0"
+chrono = { version = "0.4", features = ["serde"] }
+which = "6.0"
+once_cell = "1.19"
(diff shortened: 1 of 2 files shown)


chenhz01:
Cross-checked the port against `scripts/lib/package-manager.js` on main: same four managers, same `pnpm > bun > yarn > npm` priority, same `bun.lockb` alias and detection order — parity is faithful. Two gaps: (1) the PR ships only `Cargo.toml` + `main.rs`, so no `Cargo.lock` — the Rust build is not reproducible, which sits oddly next to the exact-pinned Node deps in package.json; commit the lockfile. (2) Testing was manual (`cargo run` on Fedora); the JS side's edge cases deserve `#[cfg(test)]` ports — `split('@')` parsing of `pnpm@9.x`, alias lock-file fallback, `CLAUDE_PACKAGE_MANAGER` precedence over project config. Also consider a lib target alongside the binary so sibling ecc2 crates can reuse detection instead of shelling out. If this becomes a priority, I contract for exactly this kind of fix — fixed price, working code.

PR #3008, opened 2026-09-07
fix(install): gate OpenCode hook activation

samartomar:
Default and `--no-hooks` OpenCode installs now leave ECC hook entrypoints inert, including files OpenCode discovers automatically in its plugins directory. Explicit `hooks-runtime` selection plus consent restores the original entrypoints.

Install, doctor, and repair check the desired transforms and managed-file ownership. They reject modified, unverified, or unexpectedly active files, recheck before writes, and retain the previous consent decision when deactivation does not complete. Legacy installs are inspected without replaying historical operations. Current-main installer protections and the original contributor commits are preserved.

Validation: 11 hook-consent, 24 installer-executor, and 73 lifecycle checks passed during the repair; the final legacy adjustment separately passed 9 existing migration checks with an inert compiler fixture. That fixture does not establish real compilation or live OpenCode acceptance. Focused lint and Unicode checks passed. Independent code and security reviews cover the final seven-file delta. Hosted CI is running for `4be4eb4cf33f4bd2be6cce3d0b1d3731fedf9d51`.

### DIFF
--- a/scripts/lib/install-lifecycle.js
+++ b/scripts/lib/install-lifecycle.js
@@ -8,7 +8,10 @@ const { loadInstallManifests } = require('./install-manifests');
 const { readInstallState, validateInstallState } = require('./install-state');
 const { assertWithinTrustedRoot } = require('./path-safety');
 const { createInstallPlanFromRequest } = require('./install/runtime');
-const { getRecordedHookConsent } = require('./install/hook-consent');
+const {
+  disableOpenCodeHookPluginRegistration,
+  getRecordedHookConsent,
+} = require('./install/hook-consent');
 const {
   prepareClaudeSkillMigration,
 } = require('./install/claude-skill-migration');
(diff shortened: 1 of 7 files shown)


chenhz01:
The consent-gated design is the right call — deactivating entrypoints instead of deleting them means doctor and repair can restore without re-running the installer. One refactor worth doing while the surface is small: the two contentTransform dispatch chains (transformCopyFileContent in install-lifecycle.js and transformInstallContent in apply.js) both grew identical opencode branches, duplicating the existing antigravity pattern. A shared dispatcher would keep the two write paths from drifting:

const { applyContentTransform } = require('./install/content-transforms');

with each call site passing only its local extras. Separately, openCodeActivationKind's regex recognizes plugins/<dir>/index.* and package.json but not nested entry files — fine if OpenCode only auto-discovers those two shapes, but worth a comment saying so, since the next contributor will assume arbitrary plugin layouts count. Happy to extract the shared dispatcher if useful.

PR #3013, opened 2026-09-07
fix(docs): correct broken asset paths in Turkish guides

cnYui:
## What Changed

Fixed broken image/asset links in the Turkish translations of the three main guides:

- `docs/tr/the-longform-guide.md` (9 links)
- `docs/tr/the-security-guide.md` (7 links)
- `docs/tr/the-shortform-guide.md` (10 links)

All references were changed from `../assets/...` to `../../assets/...` (26 links total).

## Why This Change

From `docs/tr/`, the path `../assets/` resolves to `docs/assets/`, which does not exist — so every embedded image in the Turkish guides rendered as broken.

The assets actually live at the repository root under `assets/`. The correct relative prefix from `docs/tr/` is therefore `../../assets/`. This is already what the sibling translations use:

- Canonical root guides (e.g. `the-longform-guide.md`) use `./assets/...`
- `docs/es/` and `docs/zh-CN/` guides already use `../../assets/...`

Only the Turkish guides were off by one directory level. This change brings them in line with the other locales.

## Testing Done

- [x] Verified that after the change, 25 of the 26 corrected paths resolve to existing files in the repo.
- [x] The one remaining reference (`../../assets/images/security/badrudi-exploit.mp4`) points to a video that is not committed in any locale — the canonical `the-security-guide.md` and the `docs/es/` version reference the same not-yet-committed file, so this PR intentionally leaves it consistent rather than inventing a target.
- [x] `markdownlint-cli` run on the three files with the repo's `.markdownlint.json` — passes clean.
- [x] `git diff --check` — no whitespace errors.

## Type of Change

- [ ] `fix:` Bug fix
- [x] `docs:` Documentation

- [x] No sensitive data exposed
- [x] Change is limited to relative-path text inside existing Markdown image links; no structural or content changes

## Documentation

- [x] Documentation-only change (Turkish locale image paths)


chenhz01:
The path-depth fix is right, and matching the `../../assets/` convention already used by `docs/es/` and `docs/zh-CN/` is the correct normalization. The one remaining broken reference the PR itself flags — `../../assets/images/security/badrudi-exploit.mp4` — points at a file committed on no branch, so that embed stays broken after merge. Two options: replace the embed with a one-line Turkish callout pointing at the corresponding section of the English security guide, or commit the asset. I'd also suggest a small CI guard following the existing `scripts/ci/` convention (next to `check-unicode-safety.js`): walk `docs/**/*.md`, resolve every relative image/video ref, fail on missing targets — that would have caught all 26 of these before review instead of after. Happy to send a PR with the guard.

PR #3019, opened 2026-09-08
feat(hooks): add Hermes-native config-protection and console-log guards

ravitejarenangi:
## Summary

The Hermes install target (`install.sh --target hermes`) ships ECC's rules, skills, and commands — but no hooks. ECC's `hooks/hooks.json` is Claude Code format, and Hermes instead has its own shell-hook system: a `hooks:` block in `~/.hermes/config.yaml`, Claude-Code-compatible JSON payloads on stdin, per-tool regex matchers, a first-use consent allowlist, and `hermes hooks list | test | doctor`.

This PR ports the two ECC hooks that are harness-independent policy (rather than Claude Code plumbing) as self-contained Hermes-native Python scripts — no ECC runtime, no `run-with-flags.js`, no `.pi` extension:

- **`hooks/hermes/config-protection.py`** (`pre_tool_call`) — blocks edits to existing linter/formatter configs (`.eslintrc*`, `eslint.config.*`, `.prettierrc*`, `biome.json`, `ruff.toml`, `.stylelintrc*`, markdownlint) and lint ratchet baselines. An agent under pressure weakens the check instead of fixing the code; this hook refuses that path. Creating a config from scratch and reading configs stay allowed. Emits a JSON `decision: block` (never exit-code-based), so a malformed payload fails open and cannot brick the agent.
- **`hooks/hermes/check-console-log.py`** (`post_tool_call`) — warns via stderr (never blocks; Hermes surfaces hook stderr as a warning) when an edited TS/JS file now contains `console.log()`. Tests, `.d.ts`, and scripts excluded.

`docs/HERMES-HOOKS.md` documents why only these two were ported, the config block (Hermes wants the mapping keyed by event name, not a list — worth knowing, it cost me one round-trip), the consent/allowlist step, and verification commands.

## Type
- [ ] Skill
- [ ] Agent
- [x] Hook
- [ ] Command
- [x] Docs

## Testing

- Payload-level proof of every path (block, allow, allow-new-config, warn, silent):
  - edit of an existing `eslint.config.mjs` → `{"decision":"block",...}`
  - edit of a lint ratchet baseline → blocked
  - edit of normal source → silent allow
  - creation of a new config file → silent allow
  - TS file with `console.log()` → stderr warning; clean file → silent
- Registered on a live Hermes install (v0.21.0): `hermes hooks list` shows both with matchers; `hermes hooks test pre_tool_call --for-tool patch` and `post_tool_call --for-tool write_file` fire cleanly.
- `npm test`: 4394/4395 pass. The single failure is a pre-existing ETIMEDOUT flake in `tests/scripts/setup.test.js` ("all interactive choices from an existing install..."); it also reproduces without this PR's files and passes 30/30 when run alone. This PR adds no JS and touches no setup/install code.
- Repo validators on the changed files: `validate-no-personal-paths`, `check-unicode-safety`, `markdownlint` on the new doc — all clean.

## Checklist
- [x] Tested (with Hermes 0.21.0, macOS/arm64)
- [x] No sensitive info (API keys, personal paths — scripts reference `/home/YOU` placeholders in docs)
- [x] Clear descriptions

### DIFF
--- a/docs/HERMES-HOOKS.md
+++ b/docs/HERMES-HOOKS.md
@@ -37,16 +37,22 @@ cp hooks/hermes/config-protection.py ~/.hermes/hooks/ecc-config-protection.py
 cp hooks/hermes/check-console-log.py  ~/.hermes/hooks/ecc-check-console-log.py
 chmod +x ~/.hermes/hooks/ecc-*.py
 
-# 2. Append to ~/.hermes/config.yaml (Hermes wants the mapping keyed by event):
+# 2. Point the commands at the copies you just made. Substitute YOUR absolute
+#    home path — Hermes resolves hook commands verbatim, without ~ expansion.
+HOOK_DIR="$HOME/.hermes/hooks"
+
+# 3. Append to ~/.hermes/config.yaml (Hermes wants the mapping keyed by event):
+cat >> ~/.hermes/config.yaml <<EOF
 hooks:
   pre_tool_call:
     - matcher: "patch|write_file|edit|write|apply_patch|str_replace_editor"
-      command: "python3 /Users/YOU/.hermes/hooks/ecc-config-protection.py"
+      command: "python3 $HOOK_DIR/ecc-config-protection.py"
       timeout: 10
   post_tool_call:
     - matcher: "patch|write_file|edit|write|apply_patch|str_replace_editor"
-      command: "python3 /Users/YOU/.hermes/hooks/ecc-check-console-log.py"
+      command: "python3 $HOOK_DIR/ecc-check-console-log.py"
       timeout: 10
+EOF
 ```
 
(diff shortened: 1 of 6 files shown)


chenhz01:
The exists-check leaves a one-`rm` bypass: blocking requires the target to already exist, so an agent can delete `.eslintrc.json` through a shell turn (never inspected — EDIT_TOOLS covers only file-editing tools) and write the weakened config in the next turn, where the create path is explicitly allowed. Since Hermes payloads also carry shell tools, consider giving delete-shaped calls the same block JSON when a protected basename appears in the command string; otherwise the protection holds exactly until the first deletion. Nice catch keeping lstat for dangling symlinks — that edge is easy to miss. Happy to open a PR with working code if useful.

PR #3039, opened 2026-09-09
feat(commands): add token-card command

iyashjayesh:
## Summary

Adds `/token-card`, a command that renders a token-usage stat card from local agent session logs and writes it into the working tree as a committed file.

It complements `/cost-report` rather than overlapping it — they read different sources and produce different artifacts:

| | `/cost-report` | `/token-card` |
|---|---|---|
| Source | `~/.claude/metrics/costs.jsonl`, written by ECC's `stop:cost-tracker` hook | The agents' own session logs |
| Agents | Claude Code | Claude Code, Codex, OpenCode |
| Output | Terminal summary, optional CSV | An SVG file committed to the repo |

`/cost-report` answers "what did I spend this week". `/token-card` puts a current figure in a README.

The command shells out to `npx -y @tokenchit/cli@latest` (MIT, <https://github.com/iyashjayesh/tokenchit>). Nothing is installed globally. It reads token counts and timestamps only — never prompts, code, or file contents — and the workflow stops before publishing anything, since joining the public leaderboard is a separate explicit opt-in.

## Type

- [ ] Skill
- [ ] Agent
- [ ] Hook
- [x] Command

## Testing

`npm test` — **4396 passed, 0 failed**.

Running it surfaced two things worth calling out:

1. The command count moving 94 → 95 left `docs/tr/AGENTS.md` stale in two places, which `npm run catalog:sync` does not reach but `tests/docs/configure-ecc-install-paths.test.js` asserts on. Both updated. The other `94`s in the tree are historical release prose (`docs/releases/2.1.0/`, README v2.0.0 section, `.pi/README.md` verification note) and are deliberately left alone.
2. `npm install` rewrites `yarn.lock` (~3.6k lines of churn). Reverted, since `package.json` is unchanged and CONTRIBUTING warns that a modified lockfile fails hardened-mode CI on its own.

Generated files come from `npm run command-registry:write` and `npm run catalog:sync`. `agent.yaml` is a single-line insert — no reordering of existing entries. `markdownlint` clean on the new file.

## Checklist

- [x] Tested with Claude Code
- [x] Clear descriptions


chenhz01:
The /cost-report vs /token-card split is clean — different sources, different artifacts, no overlap. One thing to tighten before merge: the command shells out to npx -y @tokenchit/cli@latest, which re-resolves the package on every invocation, so any future publish under that name changes what runs inside users' sessions with no review gate. Pin the exact version in package.json and invoke the local binary through the script entry instead of @latest, or add a first-run check comparing the resolved version against a committed expected value. Also worth documenting in the skill what the CLI is permitted to read from session logs — the token-counts-only claim is this PR's, but the execution surface belongs to the third party. Happy to implement this under a small paid engagement if that's easier than DIY.

PR #3052, opened 2026-09-10
fix(hooks): recognize git remote get-url/-v/show as read-only in gateguard-fact-force

startRED:
## Summary
- `isReadOnlyGitIntrospection()` in `scripts/hooks/gateguard-fact-force.js` allowlists `status`/`diff`/`log`/`show`/`branch`/`rev-parse`, but not `git remote` — so any session whose first Bash command reads the remote URL (e.g. a router/onboarding skill running `git remote get-url origin` to decide which project config to load) gets gated by the routine Fact-Forcing check even though the command is pure read.
- Adds the read-only `remote` forms: bare `git remote`, `-v`/`--verbose`, `get-url [--push|--all] <name>`, `show [-n] <name>`. Mutating subcommands (`add`, `remove`, `rename`, `set-url`, `set-head`, `prune`, `update`, ...) still fall through to the routine Bash gate as before.
- Hit this in practice: a personal Claude Code skill runs `git remote get-url origin` as its very first command every session to route between project configs, so the gate fired on every single session start for a command that never mutates anything.

## Test plan
- [x] `node tests/hooks/gateguard-fact-force.test.js` — extended the existing read-only-introspection allowlist test with the new `git remote` forms, changed the "unsupported git commands still gated" example off of `git remote -v` (now allowed) to `git fetch`, and added a dedicated test asserting mutating `git remote` subcommands stay gated.
- [x] Confirmed no regressions: same 161/162 passing baseline on `upstream/main` before the change (162 after, +1 net new test), same 35 pre-existing unrelated failures (glob-scoping/PowerShell-path tests) before and after on this machine — unaffected by this change.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/scripts/hooks/gateguard-fact-force.js
+++ b/scripts/hooks/gateguard-fact-force.js
@@ -1096,6 +1096,31 @@ function isReadOnlyGitIntrospection(command) {
     return args.length === 2 && args[0] === '--abbrev-ref' && /^head$/i.test(args[1]);
   }
 
+  if (subcommand === 'remote') {
+    // Allows: git remote, git remote -v, git remote get-url [--push|--all] <name>,
+    // git remote show [-n] <name>. Denies add/remove/rename/set-url/set-head/prune/update etc.
+    if (args.length === 0) return true;
+
+    const [first, ...rest] = args;
+    if (first === '-v' || first === '--verbose') {
+      return rest.length === 0;
+    }
+
+    const isRemoteName = arg => /^[a-zA-Z0-9._-]+$/.test(arg);
+
+    if (first === 'get-url') {
+      if (rest.length === 0 || rest.length > 2) return false;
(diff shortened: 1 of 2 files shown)


chenhz01:
Most of this thread is already debating the `remote` allowlist itself, so just two reviewable details: (1) the PR bundles an unrelated `scripts/lib/plan-canvas/server.js` refactor (injectable `clock` for `presenceFor`/`markThinking`/typing timestamps) into a security-sensitive hook change — splitting it out keeps the gateguard diff auditable in isolation. (2) #2755 and #2838 both also edit `gateguard-fact-force.js`, so three open PRs are racing on the same file; whoever merges second hand-resolves around the new `remote` branch inside `isReadOnlyGitIntrospection`. Worth stating a merge order in the thread now to avoid a silent semantic conflict in the matcher/introspection logic.

PR #3057, opened 2026-09-10
fix: serialize Claude tool-use round trips

### LINKED ISSUE #3056: fix: serialize Claude tool-use round trips with Anthropic message blocks
## Claude tool-use round trips need Anthropic-native message blocks

`ReActAgent` appends an assistant message containing `tool_calls` and one or more tool-result messages before the next provider request. `ClaudeProvider.generate()` currently serializes those messages with the generic OpenAI-shaped `Message.to_dict()` payload (`tool_calls` on the assistant and `role: tool` for results). Anthropic Messages API expects `tool_use` blocks on the assistant turn and `tool_result` blocks inside a user turn.

This issue tracks a focused fix to serialize the existing provider-neutral messages into the Anthropic wire format and cover the round trip with unit tests.

auyua9:
## Summary

- serialize assistant tool calls as Anthropic `tool_use` content blocks
- serialize tool results as `tool_result` blocks in a user turn and group consecutive results
- add regression coverage for a Claude tool-use round trip

This fixes the existing `ReActAgent` history path, which currently sends OpenAI-shaped `role: tool` / `tool_calls` messages to the Anthropic Messages API.

Closes #3056

## Testing

- `python3 -m ruff check src/llm tests/test_claude_provider.py`
- `PYTHONPATH=src python3 -m pytest` (96 passed)

### DIFF
--- a/src/llm/core/types.py
+++ b/src/llm/core/types.py
@@ -59,18 +59,18 @@ class Message:
             }
 
         if self.role == Role.ASSISTANT and self.tool_calls:
-            content: list[dict[str, Any]] = []
-            if self.content:
-                content.append({"type": "text", "text": self.content})
-            content.extend(
-                {
-                    "type": "tool_use",
-                    "id": tool_call.id,
-                    "name": tool_call.name,
-                    "input": tool_call.arguments,
-                }
-                for tool_call in self.tool_calls
-            )
+            content: list[dict[str, Any]] = [
+                *([{"type": "text", "text": self.content}] if self.content else []),
+                *(
+                    {
+                        "type": "tool_use",
+                        "id": tool_call.id,
+                        "name": tool_call.name,
+                        "input": tool_call.arguments,
+                    }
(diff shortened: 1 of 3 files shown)


chenhz01:
The review above is solid; two edges I don't see covered: Serialization and the consecutive-result merge look right. Two malformed-history edges still hit the API raw:

1. A TOOL message with no `tool_call_id` serializes to `"tool_use_id": ""`, which Anthropic rejects with a confusing 400. Better to skip the orphan than send an empty id:

```python
if not self.tool_call_id: continue  # log and drop
```

2. A tool result that doesn't immediately follow its assistant `tool_use` turn (e.g. a user text turn interleaved in the transcript) lands in its own later user turn, and the API requires `tool_result` in the turn directly after `tool_use` — same 400 class. The merge loop could detect that orphan case and drop the result with a warning rather than letting the whole request fail. Happy to open a PR with both guards and tests if useful.

PR #3088, opened 2026-09-11
fix: guided dry-run via env, atomic install-state writes, harden hook and provider parsing

VarunGore36:
## Summary

- Guided install now honors the global dry-run state (`ECC_DRY_RUN=1`) in addition to `--dry-run`: `install-guided.js` resolves dry-run from env, and `install-apply.js` forwards it on the `--guided` dispatch path so a prefixed `ecc --dry-run install --guided...` can no longer perform real writes.
- `writeInstallState` is atomic (tmp + fsync + rename) with temp cleanup on write/rename failure and a best-effort parent-directory fsync after rename.
- Hook dispatcher `hasRunExport` check strips comments/strings before matching, so `require()` is only attempted for real `run` exports; spawn fallback preserved.
- Ollama `_parse_tool_arguments` returns a copy instead of aliasing the provider-response dict; OpenAI gains the same malformed/non-dict fallback already used by Atlas.

## Tests

- `tests/scripts/install-guided.test.js`: new case proves `ECC_DRY_RUN=1` without a flag performs no apply and leaves the filesystem unchanged (15/15 pass).
- `tests/lib/install-state.test.js`: rename-failure and write-failure boundary cases preserve the original file with no stale temps (8/8 pass).
- Python: OpenAI malformed/non-dict args, new Ollama provider cases (string/dict/malformed args, connection-vs-auth mapping), async `aexecute` and ReActAgent partial-output/trajectory cases; JS syntax checks and manual harnesses verified.

## Verification

- `ECC_DRY_RUN=1 node scripts/install-apply.js --guided --harness codex --yes --json` emits a `dryRun: true` preview instead of applying.

### DIFF
--- a/scripts/hooks/bash-hook-dispatcher.js
+++ b/scripts/hooks/bash-hook-dispatcher.js
@@ -192,8 +192,23 @@ async function main() {
   if (result.stderr) {
     process.stderr.write(result.stderr);
   }
-  process.stdout.write(result.output);
-  process.exit(result.exitCode);
+  exitWithStdout(result.output, result.exitCode);
+}
+
+function exitWithStdout(text, exitCode) {
+  process.exitCode = exitCode;
+  let pendingWrites = 1;
+  const exitWhenFlushed = () => {
+    pendingWrites -= 1;
+    if (pendingWrites === 0) {
+      process.exit(exitCode);
+    }
+  };
+  if (typeof text === 'string' && text.length > 0) {
+    pendingWrites += 1;
+    process.stdout.write(text, exitWhenFlushed);
+  }
+  process.stderr.write('', exitWhenFlushed);
 }
 
 if (require.main === module) {
--- a/scripts/hooks/pre-bash-dispatcher.js
+++ b/scripts/hooks/pre-bash-dispatcher.js
@@ -19,6 +19,21 @@ process.stdin.on('end', () => {
   if (result.stderr) {
     process.stderr.write(result.stderr);
   }
-  process.stdout.write(result.output);
-  process.exitCode = result.exitCode;
+  exitWithStdout(result.output, result.exitCode);
 });
+
+function exitWithStdout(text, exitCode) {
+  process.exitCode = exitCode;
+  let pendingWrites = 1;
+  const exitWhenFlushed = () => {
+    pendingWrites -= 1;
+    if (pendingWrites === 0) {
+      process.exit(exitCode);
+    }
+  };
+  if (typeof text === 'string' && text.length > 0) {
+    pendingWrites += 1;
+    process.stdout.write(text, exitWhenFlushed);
+  }
+  process.stderr.write('', exitWhenFlushed);
+}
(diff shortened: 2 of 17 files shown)


affaan-m (changes requested):
Reviewed at the exact head c5b3c60d (independent review, all suites run in a clean worktree: install-state 8/8, dry-run 12/12, install-guided 15/15, install-apply 42/42, uninstall, repair, doctor, dispatchers, hooks 253 pass, pytest 20 pass, eslint clean). The dry-run and atomic-write work is good. One blocker before this can merge:

P1, scripts/hooks/run-with-flags.js:112-195 and 292-294: the new comment/string stripper in hasRunExport desyncs on regex literals and template literals, so three shipped hooks that really export run() are now misdetected and silently dispatched via spawnSync instead of require(). Verified by scanning every scripts/hooks/*.js with the old regex, the new regex, and the real typeof require(f).run:
- block-no-verify.js: line 299 `/[\s"']/` opens a double-quoted string that never closes
- pre-bash-commit-quality.js: line 147 `(?!['"])` inside a regex literal
- post-edit-format.js: line 24 a character class containing a backtick opens a template literal
All three keep working only because they have a require.main stdin fallback, but every call now pays the spawn the require() path exists to avoid, and a hook written the same way without that fallback becomes a silent pass-through (resolveLegacySpawnStdout echoes input, exit 0). Secondary: `/module\.exports\s*=\s*\{[^}]*\brun\b/` false-negatives on `module.exports = { helper() { ... }, run }` because [^}]* stops at the first nested brace.

Fix: either add a regex-literal state to the scanner or strip only comments (not strings), replace the object-literal regex with a brace-aware match, and add a test that asserts hasRunExport is true for each scripts/hooks/*.js whose require() exposes run (that test would have caught this).

Non-blocking P2s to consider in the same push or later: writeInstallState's rename-over drops the existing file mode (chmod 600 becomes 644) and a fsync EINVAL/ENOTSUP on some mounts now aborts the write; isDryRun() at module scope in install-apply.js gives a raw stack trace for ECC_DRY_RUN=yes on the guided path; the {"raw": ...} and {"value": ...} sentinel keys in the OpenAI/Ollama argument parsers collide with real tool parameters named raw or value (use a reserved key or return an is_error result); connection errors from Ollama now propagate as raw URLError rather than an LLMError so ReActAgent.run does not catch them; exitWithStdout is duplicated in two dispatchers instead of shared.

#3087 is a strict ancestor of this branch with nothing unique, so it is being closed in favour of this PR.

chenhz01:
Given the review volume here I'll keep this to one point I don't think is covered yet, on the same exit path as the dispatcher change: `exitWithStdout` waits on write callbacks that may never come if the consumer closed stdout (EPIPE), leaving the hook to be killed by its parent — an `'error'` listener plus an `.unref()`'d exit timeout closes that case cheaply. On `writeInstallState`: tmp + fsync + rename is the right shape; one portability caveat is that rename-over-existing can fail with EPERM on Windows while AV or an indexer holds the old state file open. A single bounded retry (two attempts, ~50ms apart) before degrading to the old in-place write keeps the atomic path on POSIX without bricking Windows installs. Happy to open a PR with the retry logic.

PR #3099, opened 2026-09-12
docs(plankton-code-quality): add oxlint complexity rule and ratchet ceiling for JS/TS

gabchess:
Closes the JS/TS gap in the plankton-code-quality model-routing table.

Python complexity already routes to Sonnet via ruff (C901, PLR codes). JS/TS had no
equivalent rule enabled by default: oxlint ships a `complexity` rule (ported from
eslint), but it lives in the `restriction` category, which oxlint does not enable
out of the box. The skill listed oxlint as Optional, so the gap was invisible.

This adds a section covering:

- Enabling `complexity` in `.oxlintrc.json` with the documented default (`max: 20`),
  verified against the oxc.rs rule docs.
- Promoting oxlint from Optional to Required for TypeScript/JS when this rule is in
  play.
- A model-routing row for JS/TS complexity, mirroring the existing Python row.
- A ratchet-ceiling workflow for adopting the rule on a real codebase without a
  big-bang refactor, with a worked example from a public PR (modem-dev/hunk #861)
  where the measured ceiling landed at `max: 80` and ratchets down as hotspots get
  refactored.

Docs-only change: one file, additions only, no behavior or code changes.

### DIFF
--- a/skills/plankton-code-quality/SKILL.md
+++ b/skills/plankton-code-quality/SKILL.md
@@ -235,3 +235,83 @@ Track:
 - average remediation time
 - repeat violations by category
 - merge blocks due to gate failures
+
+---
+
+## Gabe's addition: oxlint complexity + ratchet ceiling (JS/TS)
+
+Closes the JS/TS gap in the model-routing table above. Python complexity (ruff C901,
+PLR) already routes to Sonnet. JS/TS had no equivalent rule turned on by default.
+
+### Turn on oxlint's `complexity` rule, Required not Optional
+
+oxlint's `complexity` rule (source: eslint's `complexity` rule, ported) lives in the
+"restriction" category, which oxlint does not enable by default. It must be turned on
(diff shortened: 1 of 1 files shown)


chenhz01:
Following the review round: with the measure-first setup in place, the ratchet itself is still honor-system. After the initial measured ceiling lands (the `max: 80` example), nothing stops a later PR from loosening `max` in `.oxlintrc.json`, and the ceiling only ratchets down if someone remembers to move it. A three-line CI guard makes it an invariant:
```js
assert(next.max <= prev.max, 'complexity ceiling may only tighten');
```
reading the previous value from the merge base. Combined with the documented workflow, every complexity regression then forces either a refactor or an explicit, reviewable ceiling bump — which is the behavior the ratchet section promises but doesn't yet enforce. Want me to take a crack at it? Happy to open a PR.

PR #3115, opened 2026-09-13
fix: resolve OpenCode source imports and installed skills paths

### LINKED ISSUE #3112: [Problem] Bug: OpenCode app stops responding after installing ECC (ERR_MODULE_NOT_FOUND on custom tools)
### What is the impact?

ECC installs, but nothing loads

### What happened?

# Bug: OpenCode app stops responding after installing ECC (ERR_MODULE_NOT_FOUND on custom tools)

## Summary

After installing ECC (profile `full`, target `opencode-home`), the OpenCode application no longer responds to any prompt. Sessions start but produce no model output, and the logs show `prompt_async failed` with `ERR_MODULE_NOT_FOUND` on the ECC custom tools.

## Environment

- OpenCode **1.18.25** (macOS desktop app + CLI)
- ECC **2.2.1** (commit `8321021c`), installed `2026-09-13T13:17:06Z`
- Install target: `opencode-home` -> root `/Users/<user>/.config/opencode`
- Install state file: `~/.config/opencode/ecc-install-state.json`

## Symptom

1. Every new session fails silently with no model response.
2. OpenCode log (`~/.local/share/opencode/log/opencode.log`):

```
ERROR message="prompt_async failed" sessionID=ses_... cause="Cause([Die(Error [ERR_MODULE_NOT_FOUND]:
Cannot find module '.../.config/opencode/tools/run-tests.js'
imported from .../.config/opencode/tools/index.ts)])"
```

The failure happens while **building the tool registry** (`ToolRegistry.state` -> `SessionTools.resolve`), i.e. **before** any model request - so the app stops responding entirely.

## Root cause

ECC installs the **TypeScript source barrels** into OpenCode's active directories, but these barrels import modules using the **`.js`** extension while only `.ts` files exist next to them (the matching `.js` files only exist after compilation, inside `dist/`).

Installed file: `~/.config/opencode/tools/index.ts` (copied from `ECC/.opencode/tools/index.ts`):

```ts
export { default as runTests } from "./run-tests.js"            // line 8
export { default as checkCoverage } from "./check-coverage.js"  // line 9
export { default as securityAudit } from "./security-audit.js"  // line 10
export { default as formatCode } from "./format-code.js"        // line 11
export { default as lintCheck } from "./lint-check.js"          // line 12
export { default as gitSummary } from "./git-summary.js"        // line 13
export { default as changedFiles } from "./changed-files.js"    // line 14
export { default as dependencyAnalyzer } from "./dependency-analyzer.js" // line 15
```

Installed file: `~/.config/opencode/plugins/index.ts` (copied from `ECC/.opencode/plugins/index.ts`):

```ts
export { ECCHooksPlugin, default } from "./ecc-hooks.js"        // line 9
export * from "./ecc-hooks.js"                                   // line 12
```

In both directories only the `.ts` files exist (`./run-tests.ts`, `./ecc-hooks.ts`, ...). The referenced `.js` files exist only as `dist/tools/*.js` and `dist/plugins/*.js` (the `tsc` output of `scripts/build-opencode.js`). The ESM loader used by OpenCode does **not** do `.js -> .ts` resolution, so the import `./run-tests.js` is not found and the entire tool registry fails.

Verification: the **compiled** `ECC/.opencode/dist/tools/index.js` does reference `./run-tests.js` - and it resolves fine, because `dist/tools/run-tests.js` exists. The bug is strictly in the `.ts` source barrels that get copied into the active directories.

## Reproduction

1. Install ECC with target `opencode-home` (profile `full`).
2. Launch OpenCode (app or CLI).
3. Send a prompt -> no response.
4. Inspect the log: `prompt_async failed ... ERR_MODULE_NOT_FOUND ... tools/run-tests.js`.

## Fix applied locally

In `~/.config/opencode/tools/index.ts` and `~/.config/opencode/plugins/index.ts`, change import extensions from `.js` -> `.ts`:

```ts
// before
export { default as runTests } from "./run-tests.js"
// after
export { default as runTests } from "./run-tests.ts"
```

Side note (also needed on this machine) - in `opencode.json`: the skills path `"../skills"` resolved outside the config folder (missing `/Users/<user>/skills`) -> fix to `"./skills"`.

**Restart required**: the desktop app and `opencode serve` keep the old tool registry in memory -> relaunch them after applying the fix.

## Fix to integrate in the ECC repo

The bug lives in `ECC/.opencode/tools/index.ts` and `ECC/.opencode/plugins/index.ts`. Two options:

- **Option A (simple)**: use `.ts` for imports between adjacent `.ts` modules (the two barrels) - compatible with the `opencode-home` target, which loads the `.ts` files in-place.
- **Option B (structural)**: for the `opencode-home` target, install the **compiled** barrels from `dist/` (`dist/tools/index.js`, `dist/plugins/index.js`) instead of the `.ts` sources.

### Harness

OpenCode

### Install method

ecc or ecc-install CLI

### Operating system

macOS

### ECC and harness versions

ECC 2.1.0

### Optional redacted diagnostics

_No response_

yuefdev:
## What Changed

OpenCode's full home install currently copies TypeScript sources whose `.js` imports point to files that exist only under `dist/`. This causes `ERR_MODULE_NOT_FOUND` before the first model request. Use `.ts` for relative source imports, including the lazy changed-files store, and enable TypeScript's `rewriteRelativeImportExtensions` so published JavaScript still imports `.js` files. Raise the standalone compiler minimum to 5.7 while retaining the locked 5.9.3 version.

Home installation also rewrites the bundled `../skills` path to `./skills`, preserving the repository template and other configured paths. Installation and lifecycle operations share the transformation so repair restores identical content and uninstall recognizes it.

## Why This Change

Fixes #3112.

This is an alternative to #3113: it keeps the existing TypeScript runtime layout and also supports loading the repository sources directly. It does not migrate active runtime files to compiled JavaScript.

## Testing Done

- Reproduced the reported `ERR_MODULE_NOT_FOUND` in a full-profile isolated home before the fix; the same regression test passes afterward.
- Verified all eight tool exports, plugin initialization, and hook-to-tool changed-files tracking in both installed TypeScript and compiled JavaScript. The test loader preserves import specifiers and does not remap missing `.js` files to `.ts`.
- Verified install -> deleted config -> repair -> content digest -> uninstall, including preservation of custom skill paths and the repository config.
- Passed 12 focused suites: build/pack (6), hooks (9), tools (9), changed-files store (8), config (4), installer (21), adapters (51), lifecycle (66), OpenCode migration (9), Antigravity migration (14), planning boundary (3), and publish surface (2).
- Root TypeScript 6.0.3 and isolated standalone TypeScript 5.9.3 builds pass. Native Node 24 TypeScript loading and published declaration type-checking pass.
- Shared transform coverage: 89.65% lines/statements, 84.61% branches, 100% functions.
- Scoped ESLint, install manifest validation, `npm pack`, and `git diff --check` pass.

Validation ran on Windows / Node 24.19.0. The full repository suite, Node 18 execution, and actual macOS/OpenCode desktop startup or model responses were not tested. Existing Windows-incompatible symlink/POSIX subcases were skipped by their suites.

## Type of Change

- [x] `fix:` Bug fix

- [x] Regression and lifecycle edge cases tested

## Dependency Changes

- [x] Ran `corepack yarn install --mode=update-lockfile`; the root lockfile is unchanged. Updated the nested package manifest and lockfile together; no resolved dependency versions changed.


chenhz01:
Right scope — `.ts` relative imports plus `rewriteRelativeImportExtensions` keep one source layout for both Bun's direct `.ts` execution and the published JS build. One upgrade edge worth a fixture: homes installed before this change still have `tools/index.ts` importing `./run-tests.js` (the #3112 state), and `scripts/lib/install/content-transform.js` is now the shared recognition/rewrite path for install, repair, and uninstall. If the recognition patterns only match the new `.ts` form, repair would leave a pre-fix home broken and uninstall would treat those files as foreign content. Worth a fixture with the legacy layout:

```
transform(legacyContentWithJsImports) -> rewrittenContentWithTsImports
```

If this becomes a priority, I contract for exactly this kind of fix — fixed price, working code.

PR #3185, opened 2026-09-20
fix(gateguard): strip C1 controls and Unicode line/paragraph separators from file paths

auyua9:
{
  "title": "fix(gateguard): strip C1 controls and Unicode line/paragraph separators from file paths",
  "body": "## Summary\n\nThe `gateguard-fact-force` hook's `sanitizePath` replaced C0 controls\n(\u2264 U+001F), DEL (U+007F), bidi overrides, and the C1 isolate block\n(U+2066–U+2069) with spaces before splicing a file path into the deny\nreason printed to terminals/logs. Two adjacent control ranges slipped\nthrough:\n\n- C1 controls U+0080–U+009F, including CSI (U+009B), which can\n  rewrite how ANSI-capable terminals interpret the next characters.\n- Unicode line/paragraph separators U+2028 and U+2029, which most\n  terminals render like `\\n`/`\\n\\n` but `sanitizePath` previously\n  left intact.\n\nA user-controlled path containing any of those codepoints therefore\nleaked control structure into the deny reason string, which is then\nmirrored into downstream logs.\n\nThis change extends the existing space-substitution predicate to\ncover both ranges and adds a regression test that constructs a path\nwith CSI + U+2028 + U+2029 around a visible filename and asserts\nnone of the three codepoints appear in `permissionDecisionReason`\nwhile the visible filename does. The existing C0/DEL/bidi-override\nbehaviour is preserved.\n\n## Diff\n\n```\nscripts/hooks/gateguard-fact-force.js    | 12 ++++++++++--\ntests/hooks/gateguard-fact-force.test.js | 33 ++++++++++++++++++++++++++++++++\n2 files changed, 43 insertions(+), 2 deletions(-)\n```\n\n## Scope\n\nSingle-component change scoped to the gateguard hook and its in-repo\ntest. No API change, no other hook touched.\n\n## Validation\n\n- `git diff --check` clean\n- The added test runs in the project's existing zero-dependency Node\n  test runner (`tests/hooks/gateguard-fact-force.test.js` already\n  uses bare `node:test` style counters); CI will exercise it on\n  push."
}


chenhz01:
The C1/CSI and U+2028/2029 gaps are real, and the regression test asserting the visible filename survives is the right check. Two increments while you're in `sanitizePath`: the predicate still passes zero-width and soft-hyphen characters (U+200B–U+200D, U+FEFF, U+00AD). They can't rewrite terminal control flow, but they survive into the deny-reason string, so two visually identical paths produce strings that don't byte-match — which matters if anything downstream logs, dedupes, or diffs decisions. That's a one-line extension of the existing `isHarmless` check. Second, consider collapsing the four boolean categories into a single character-class regex (`/[\u0000-\u001F\u007F-\u009F\u2028\u2029\u200E-\u200F\u202A-\u202E\u2066-\u2069]/u`) — same semantics, and the next range question can be answered by reading one line. Happy to open a PR with both.

